PR #1, opened 2016-05-05
Update externs.m

TheDharc:
adding what lacks

### DIFF
--- a/Source/externs.m
+++ b/Source/externs.m
@@ -583,6 +583,8 @@ NSString *NSMacSimpleTextDocumentType = @"NSMacSimpleText";
 NSString *NSHTMLTextDocumentType = @"NSHTML";
 NSString *NSDocFormatTextDocumentType = @"NSDocFormat";
 NSString *NSWordMLTextDocumentType = @"NSWordML";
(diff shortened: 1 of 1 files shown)


ivucica:
Several things

#### How to contribute

Both #1 and #2 seem to be small enough that

- I can import them without requiring copyright assignment to the FSF
- I would be willing to accept them through Github

However, please note that we are not yet using Git for version control. This is just a mirror of our Subversion repositories. Creating a pull request is not (yet) the best way to contribute. In future, please consider sending the patches to our developer mailing list.

It is also rather good to add some context to what problem your patch is solving.

#### Identity of contributor

While working on GNUstep, I would also prefer to work with a real identity of a person. Could you share a name I could attribute the patches to, if the patches get accepted? 

#### Is this a partial contribution?

Are you just improving source level compatibility? If so, then this is fine.

Are you planning to contribute actual support for OOXML  and ODF? That would be much appreciated! If this is accurate, I'd kindly ask you to reach our to us to give you information on how to assign copyright to FSF.

#### Thanks

Most important of all: thank you for your contributions!

PR #4, opened 2017-07-20
Solve very low-hanging fruits in GUI for compatibility

theiostream:
In this pull request some _very_ easy low-hanging fruits in GUI for WebKit support were solved.

### DIFF
--- a/Headers/AppKit/NSSharingService.h
+++ b/Headers/AppKit/NSSharingService.h
@@ -1,0 +1,45 @@
+/** NSSharingService.h
+
+   <abstract>Interface for sharing things</abstract>
+
+   Copyright <copy>(C) 2017 Free Software Foundation, Inc.</copy>
+
+   Author: Daniel Ferreira <dtf@stanford.edu>
+   Date: 2017
+
+   This file is part of the GNUstep GUI Library.
+
+   This library is free software; you can redistribute it and/or
+   modify it under the terms of the GNU Lesser General Public
+   License as published by the Free Software Foundation; either
+   version 2 of the License, or (at your option) any later version.
+
+   This library is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.	 See the GNU
+   Lesser General Public License for more details.
+
+   You should have received a copy of the GNU Lesser General Public
+   License along with this library; see the file COPYING.LIB.
+   If not, see <http://www.gnu.org/licenses/> or write to the 
+   Free Software Foundation, 51 Franklin Street, Fifth Floor, 
+   Boston, MA 02110-1301, USA.
+*/
+#ifndef __GNUstep_NSSharingService
+#define __GNUstep_NSSharingService
+
+#import <Foundation/NSObject.h>
+
+@interface NSSharingService : NSObject
+@end
+
+@protocol NSSharingServiceDelegate <NSObject>
+@end
+
+@interface NSSharingServicePicker : NSObject 
+@end
+
+@protocol NSSharingServicePickerDelegate <NSObject>
+@end
+
+#endif
--- a/Headers/AppKit/NSSpeechSynthesizer.h
+++ b/Headers/AppKit/NSSpeechSynthesizer.h
@@ -151,8 +151,13 @@ extern NSString *NSSpeechDictionaryEntryPhonemes;
 - (NSString *) phonemesFromText: (NSString *)text;
 @end
 
-// Speech synthesizer delegate informal protocol...
+@protocol NSSpeechSynthesizerDelegate <NSObject>
+#if GS_PROTOCOLS_HAVE_OPTIONAL
+@optional
+#else
+@end
 @interface NSObject (NSSpeechSynthesizerDelegate)
+#endif
 - (void)speechSynthesizer: (NSSpeechSynthesizer *)sender
  didEncounterErrorAtIndex: (NSUInteger)index 
                  ofString: (NSString *)string 
--- a/Headers/AppKit/NSSpeechSynthesizer.h
+++ b/Headers/AppKit/NSSpeechSynthesizer.h
@@ -64,6 +64,12 @@ extern NSString *NSSpeechModePhoneme;
 extern NSString *NSSpeechModeNormal;
 extern NSString *NSSpeechModeLiteral;
 
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_5, GS_API_LATEST)
+extern NSString *NSSpeechResetProperty;
+extern NSString *NSSpeechOutputToFileURLProperty;
+extern NSString *NSSpeechPitchBaseProperty;
+#endif
+
 // values for speech status...
 extern NSString *NSSpeechStatusOutputBusy;
 extern NSString *NSSpeechStatusOutputPaused;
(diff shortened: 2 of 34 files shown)


Headers/AppKit/NSSpeechSynthesizer.h
```diff
@@ -151,8 +151,13 @@ extern NSString *NSSpeechDictionaryEntryPhonemes;
...
 @end
 
-// Speech synthesizer delegate informal protocol...
-@interface NSObject (NSSpeechSynthesizerDelegate)
+@protocol NSSpeechSynthesizerDelegate <NSObject>
+#if GS_PROTOCOLS_HAVE_OPTIONAL
+@optional
+#else
+@end
+@interface NSObject (NSURLConnectionDelegate)
```
tedge: (NSURLConnectionDelegate) -> (NSSpeechSynthesizerDelegate)

theiostream: Thank you. Has been fixed.

(I used `NSURLConnectionDelegate` mistakenly because it was my reference for this change.)

Headers/AppKit/NSUserInterfaceItemIdentification.h
```diff
@@ -0,0 +1,39 @@
...
+*/
+
+#ifndef _GNUstep_H_NSUserInterfaceItemIdentification
+#define _GNUstep_H_NSUserInterfaceItemIdentification
+
+@class NSString;
+
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_7, GS_API_LATEST)
+@protocol NSUserInterfaceItemIdentification
+@property (nullable, copy) NSString *identifier;
```
fredkiefer: Could you please make this normal methods instead of a property? We still try to keep GNUstep gui free of Objc2 features. What you did in NSView.h will do.

Source/externs.m
```diff
@@ -267,22 +267,50 @@
 
 // Pasteboard Type Globals
 NSString *NSStringPboardType = @"NSStringPboardType";
+NSString *NSPasteboardTypeString = @"NSStringPboardType";
+
 NSString *NSColorPboardType = @"NSColorPboardType";
-NSString *NSFileContentsPboardType = @"NSFileContentsPboardType";
-NSString *NSFilenamesPboardType = @"NSFilenamesPboardType";
+NSString *NSPasteboardTypeColor = @"NSColorPboardType";
```
fredkiefer: I think it is better to assign the first definition to the second. Although I hope the compiler will try to do the same. That way we are on the save side if somebody tries to compare the constant strings as pointers. (Yes, this happens in application code)

Source/NSPasteboard.m
```diff
@@ -553,6 +553,15 @@ to a filter implemented using the standard method (of
 #import "GNUstepGUI/GSServicesManager.h"
 #import "GNUstepGUI/GSPasteboardServer.h"
 
+/*
+ * FIXME
+ * We should learn to handle 
+ * NSPasteboardTypePNG
```
ivucica: I'm not sure what this refers to -- in what way do we need to learn to handle them? Do we need to learn how to paste an `NSPasteboardTypePNG` in `NSImageView` etc?

If so, is this the right place for this FIXME?

Source/externs.m
ivucica: +1, please be sure the same value is shared between symbols.

Source/NSView.m
```diff
@@ -5086,6 +5086,30 @@ - (void) bind: (NSString *)binding
...
+- (NSViewLayerContentsPlacement) layerContentsPlacement
+{
+  // FIXME (when views have CALayer support)
+  return NSViewLayerContentsPlacementScaleAxesIndependently;
+}
+
+- (void) setLayerContentsPlacement: (NSViewLayerContentsPlacement)placement
+{
+  // FIXME (when views have CALayer support)
+  return;
```
ivucica: There is no need for this return statement.

Headers/AppKit/NSView.h
```diff
@@ -590,6 +615,19 @@ PACKAGE_SCOPE
 - (NSAttributedString *)pageHeader;
 #endif
 
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_6, GS_API_LATEST)
+#if GS_HAS_DECLARED_PROPERTIES
+@property NSViewLayerContentsPlacement layerContentsPlacement;
```
ivucica: I think you want to tell the compiler that these properties are implemented as custom, non-synthesized methods, i.e. `nonatomic`, right? (I recall OS X's clang printing that as a warning.)

Headers/AppKit/NSSharingService.h
```diff
@@ -0,0 +1,45 @@
+/** NSSharingService.h
+
+   <abstract>Interface for sharing things to Facebook, Twitter etc.</abstract>
```
ivucica: It might be interesting to avoid name-dropping particular services :)

Headers/AppKit/NSSpeechSynthesizer.h
```diff
@@ -151,8 +151,13 @@ extern NSString *NSSpeechDictionaryEntryPhonemes;
 - (NSString *) phonemesFromText: (NSString *)text;
 @end
 
-// Speech synthesizer delegate informal protocol...
```
ivucica: I don't object it, but I'll note that there is no reason to remove this comment.

Source/NSView.m
```diff
@@ -5086,6 +5086,30 @@ - (void) bind: (NSString *)binding
     }
 }
 
+- (NSViewLayerContentsPlacement) layerContentsPlacement
+{
+  // FIXME (when views have CALayer support)
+  return NSViewLayerContentsPlacementScaleAxesIndependently;
```
ivucica: In situations like these, I am also never sure whether to do a one-off `NSLog` to warn the developer that this implementation of `-[NSView setLayerContentsPlacement:]` is a stub. Maybe it makes sense here, given that this getter, if it weren't already, would make the property setter a no-op?

So both for the getter and the setter:

```
static BOOL logged = NO;
if (!logged)
  {
    NSLog("warning: stub no-op implementation of -[NSView setLayerContentsPlacement:]");
    logged = YES;
  }
```

(Is there a wrapper around NSLog for warnings? I don't recall.)

Whether that makes sense -- I'm leaving that to you, I'm not going to insist on it :)

Source/externs.m
theiostream: I'm not sure this is doable. Unless in C++, we're likely to get `initializer element is not a compile-time constant` if we try to assign one value to another.

Headers/AppKit/NSSpeechSynthesizer.h
theiostream: I believe there is; it's lying if it says it's *always* an informal protocol, since if the compiler supports `@optional` it will be an actual protocol.

Source/NSPasteboard.m
theiostream: I do not know where the right place for this FIXME is because I still do not understand in depth how NSPasteboard functions and where this would need to be implemented. I placed it there because I thought it would be visible since NSPasteboard.m is an entrypoint for anyone trying to implement something.

theiostream:
All comments I did not reply to have been fixed. Otherwise, there is a response which says why I did not do it.

I believe the most critical of these is the constant string definition matter.

Let me know if there's anything else I should do or whether this is mergeable (if so, I'll generate a ChangeLog).

Source/externs.m
ivucica: Does something like `const NSString *` or `NSString const*` change anything? I'll merge as is, but consider looking at this in a future PR. Thanks :)

theiostream: Solved with `NSString *const`. Thanks for making me think of that.

theiostream:
ChangeLog has been added.

ivucica:
This has been merged as part of #6. Closing instead.

ivucica:
This has been merged as part of #5. Closing instead.

PR #6, opened 2017-07-22
Implement easy new classes in GUI

theiostream:
**NOTE**: this is based on top of #4.

This pull request contains additions to GUI (mostly for WebKit compatibility) that involved actual implementations and not only stubs.

### DIFF
--- a/Headers/AppKit/NSPasteboard.h
+++ b/Headers/AppKit/NSPasteboard.h
@@ -324,6 +324,51 @@ APPKIT_EXPORT NSString *NSCreateFilenamePboardType(NSString *fileType);
 APPKIT_EXPORT NSString *NSGetFileType(NSString *pboardType);
 APPKIT_EXPORT NSArray *NSGetFileTypes(NSArray *pboardTypes);
 
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_6, GS_API_LATEST)
+typedef NSUInteger NSPasteboardWritingOptions;
+enum {
+  NSPasteboardWritingPromised = 1 << 9
+};
+
+@protocol NSPasteboardWriting <NSObject>
+- (NSArray *)writableTypesForPasteboard: (NSPasteboard *)pasteboard;
+- (id)pasteboardPropertyListForType: (NSString *)type;
+
+#if GS_PROTOCOLS_HAVE_OPTIONAL
+@optional
+#else
+@end
+@interface NSObject (NSPasteboardWriting)
+#endif
+
+- (NSPasteboardWritingOptions)writingOptionsForType: (NSString *)type
+					 pasteboard: (NSPasteboard *)pasteboard;
+@end
+
+typedef NSUInteger NSPasteboardReadingOptions;
+enum {
+  NSPasteboardReadingAsData           = 0,
+  NSPasteboardReadingAsString         = 1 << 0,
+  NSPasteboardReadingAsPropertyList   = 1 << 1,
+  NSPasteboardReadingAsKeyedArchive   = 1 << 2
+};
+
+@protocol NSPasteboardReading <NSObject>
++ (NSArray *)readableTypesForPasteboard:(NSPasteboard *)pasteboard;
+
+#if GS_PROTOCOLS_HAVE_OPTIONAL
+@optional
+#else
+@end
+@interface NSObject (NSPasteboardReading)
+#endif
+
++ (NSPasteboardReadingOptions)readingOptionsForType: (NSString *)type
+				         pasteboard: (NSPasteboard *)pasteboard;
+- (id)initWithPasteboardPropertyList: (id)propertyList ofType: (NSString *)type;
+@end
+#endif
+
 #if defined(__cplusplus)
 }
 #endif
--- a/Headers/AppKit/NSPasteboardItem.h
+++ b/Headers/AppKit/NSPasteboardItem.h
@@ -1,0 +1,81 @@
+/** <title>NSPasteboardItem</title>
+
+   <abstract>class for wrapping pasteboard content</abstract>
+
+   Copyright <copy>(C) 2017 Free Software Foundation, Inc.</copy>
+
+   Author: Daniel Ferreira <dtf@stanford.edu>
+   Date: July 2017
+
+   This file is part of the GNUstep GUI Library.
+
+   This library is free software; you can redistribute it and/or
+   modify it under the terms of the GNU Library General Public
+   License as published by the Free Software Foundation; either
+   version 2 of the License, or (at your option) any later version.
+
+   This library is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+   Library General Public License for more details.
+
+   You should have received a copy of the GNU Library General Public
+   License along with this library; see the file COPYING.LIB.
+   If not, write to the Free Software Foundation,
+   51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+*/
+
+#ifndef _GNUstep_H_NSPasteboardItem
+#define _GNUstep_H_NSPasteboardItem 
+
+#import <AppKit/AppKitDefines.h>
+#import <AppKit/NSPasteboard.h>
+#import <Foundation/NSArray.h>
+#import <Foundation/NSDictionary.h>
+#import <CoreFoundation/CFBase.h>
+
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_6, GS_API_LATEST)
+
+@protocol NSPasteboardItemDataProvider;
+
+@interface NSPasteboardItem : NSObject <NSPasteboardWriting, NSPasteboardReading> {
+  NSMutableDictionary *_providerMap;
+  NSMutableDictionary *_dataMap;
+  NSMutableArray *_types;
+}
+#if GS_HAS_DECLARED_PROPERTIES
+@property (readonly, copy) NSArray *types;
+#else
+- (NSArray *)types;
+#endif
+
+- (NSString *)availableTypeFromArray:(NSArray *)types;
+- (BOOL)setDataProvider:(id<NSPasteboardItemDataProvider>)dataProvider
+               forTypes:(NSArray *)types;
+- (BOOL)setData:(NSData *)data forType:(NSString *)type;
+- (BOOL)setString:(NSString *)string forType:(NSString *)type;
+- (BOOL)setPropertyList:(id)propertyList forType:(NSString *)type;
+
+- (NSData *)dataForType:(NSString *)type;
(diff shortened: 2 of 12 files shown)


Source/NSPasteboardItem.m
```diff
@@ -0,0 +1,177 @@
...
+      _dataMap = [[NSMutableDictionary alloc] init];
+      _types = [[NSMutableArray alloc] init];
+    }
+
+  return self;
+}
+
+- (NSArray *)types
+{
+  return [[NSArray arrayWithArray: _types] copy];
```
fredkiefer: This way you copy the array twice.

Source/NSPasteboardItem.m
```diff
@@ -0,0 +1,177 @@
...
+- (NSString *)availableTypeFromArray:(NSArray *)types
+{
+  // FIXME: this would require a dependency on UTIs
+  return nil;
+}
+
+- (BOOL)setData:(NSData *)data forType:(NSString *)type
+{
+  if (![data isKindOfClass: [NSData class]])
+    return NO;
```
fredkiefer: The GNUstep coding style requires curly braces even for one line if statements to be consistent.

Source/NSPasteboardItem.m
```diff
@@ -0,0 +1,177 @@
...
+  
+  NSData *data = [self dataForType: type];
+  return [[[NSString alloc] initWithData: data
+                                encoding: NSUTF8StringEncoding] autorelease];
+}
+
+- (id)propertyListForType:(NSString *)type
+{
+  id object = [_dataMap objectForKey: type];
+  if (![object isKindOfClass: [NSData class]])
```
fredkiefer: Not sure whether this is correct. In most other places NSString and a few other types would be valid here too.

Source/NSTextAlternatives.m
```diff
@@ -0,0 +1,77 @@
...
+NSString *NSTextAlternativesSelectedAlternativeStringNotification =
+  @"NSTextAlternativesSelectedAlternativeStringNotification";
+
+@implementation NSTextAlternatives
+- (id)initWithPrimaryString:(NSString *)primaryString
+         alternativeStrings:(NSArray *)alternativeStrings
+{
+  if ((self = [super init]))
+    {
+      _primaryString = [primaryString retain];
```
fredkiefer: In GNUstep we prefer the RETAIN() macro. This is done mostly for historical reasons, but with ARC this could also be a speed up.

Source/NSPasteboardItem.m
```diff
@@ -0,0 +1,177 @@
...
+/*
+ * FIXME: this class is useless until NSPasteboard implements
+ * NSPasteboardReading + NSPasteboardWriting.
+ */
+
+#import <AppKit/NSPasteboardItem.h>
+#import <Foundation/Foundation.h>
+
+@implementation NSPasteboardItem
+- (id)init
```
fredkiefer: If you add an -init method you will in most cases also need a -dealloc method.

Headers/AppKit/NSPasteboard.h
```diff
@@ -296,14 +316,54 @@ APPKIT_EXPORT NSString *NSPasteboardCommunicationException;
...
 APPKIT_EXPORT NSArray *NSGetFileTypes(NSArray *pboardTypes);
 
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_6, GS_API_LATEST)
+typedef NSUInteger NSPasteboardWritingOptions;
+enum {
+  NSPasteboardWritingPromised = 1 << 9
+};
+
+@protocol NSPasteboardWriting <NSObject>
+- (NSArray *)writableTypesForPasteboard:(NSPasteboard *)pasteboard;
```
fredkiefer: Shouldn't these methods go into the optional part? Otherwise we end up with a missing @end. The same is true for Reading.

Headers/AppKit/NSPasteboard.h
```diff
@@ -296,14 +316,54 @@ APPKIT_EXPORT NSString *NSPasteboardCommunicationException;
 @end
 
 #endif
-//
-// Return File-related Pasteboard Types
```
ivucica: Similar to the other question: Any reason to remove this comment? I don't require that you keep it, I'm just interested whether we gain something by removing it.

Headers/AppKit/NSPasteboard.h
ivucica: Consider using space after colon please 😃

Headers/AppKit/NSPasteboardItem.h
```diff
@@ -0,0 +1,81 @@
...
+
+@protocol NSPasteboardItemDataProvider;
+
+@interface NSPasteboardItem : NSObject <NSPasteboardWriting, NSPasteboardReading> {
+  NSMutableDictionary *_providerMap;
+  NSMutableDictionary *_dataMap;
+  NSMutableArray *_types;
+}
+#if GS_HAS_DECLARED_PROPERTIES
+@property (readonly, copy) NSArray *types;
```
ivucica: If it's `readonly`, it doesn't matter if it's `copy`, right? That changes just the synthesized _setter_, right?

On the other hand, I believe you'll want to specify that this is `nonatomic`.

Source/NSPasteboardItem.m
```diff
@@ -0,0 +1,177 @@
...
+}
+
+- (id)propertyListForType:(NSString *)type
+{
+  id object = [_dataMap objectForKey: type];
+  if (![object isKindOfClass: [NSData class]])
+    return object;
+
+  return [NSPropertyListSerialization propertyListWithData: object
+                               options: 0
```
ivucica: These colons are not aligned

Source/NSScroller.m
```diff
@@ -130,6 +133,12 @@ + (CGFloat) scrollerWidthForControlSize: (NSControlSize)controlSize
   return scrollerWidth;
 }
 
++ (NSScrollerStyle)preferredScrollerStyle
+{
+  // FIXME: a theme should define this?
```
ivucica: No need to change the comment but: not necessarily the theme.

Either way we do not support any other style of scrollers at this time, so you can optionally remove the comment, too :)

Source/NSPasteboardItem.m
theiostream: It exists in the end of the file.

Source/NSPasteboardItem.m
theiostream: The type is valid! If it's not `NSData` it returns the type itself; if it's a data it will deserialize it.

Headers/AppKit/NSPasteboard.h
theiostream: They shouldn't go to the optional part since per the documentation they are always required. I don't see where we end up with a missing `@end`.

Headers/AppKit/NSPasteboard.h
theiostream: In this case, we don't.

Headers/AppKit/NSPasteboardItem.h
theiostream: This declaration comes from the documentation and I assume that this `copy` is not there by accident. I implemented the getter with copying.

I'm not sure we need `nonatomic` here. Despite us having a custom getter, I do not see it as thread-unsafe.

Source/NSPasteboardItem.m
theiostream: This is because if they were we'd cross the 79-line limit.

Headers/AppKit/NSPasteboard.h
fredkiefer: OK, now I see it on line 337

Headers/AppKit/NSPasteboardItem.h
ivucica: Maybe @fredkiefer can correct me, but I don't think `copy` affects the getter. Read-only+copy is not a mistake as later, in `@implementation`, you can redefine the property just before synthesizing the getter and the setter.

On the other hand, reading quickly (on the phone) about nonatomic, I think you are right.

fredkiefer: Sorry, I am the wrong person to ask. I never use the ObjC 2.0 features myself. I would expect readonly and copy to be a valid mix of attributes for a property. Readonly just means that there is no setter, just a getter and copy means that not the original (potentially mutable) value gets returned. But that is just me trying to make sense out of OBJC 2.0.

ivucica: Alright. I did some reading (I am on the phone, so I cannot try this), and I applied some thinking:

- relevant section in [document about encapsulation](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/EncapsulatingData/EncapsulatingData.html#//apple_ref/doc/uid/TP40011210-CH5-SW5) talks about setters changing behaviour but unless I missed something, not getters; I don't see evidence we getters return a copy (other than what was copied in setter)
- getters should be cheap, and making a new object every time a getter is called seems comparatively expensive

I think "readonly, copy" is identical to "readonly, assign" unless you redeclare the property as "copy" inside implementation. And it makes a difference if you use synthesize.

Either way, let's leave it as "readonly, copy" if the documented api says it should be so.

theiostream:
Haven't added a ChangeLog for this one yet waiting for the others to be merged.

theiostream:
@ivucica has been rebased against master and contains a ChangeLog.

ivucica:
Can you resolve the ChangeLog conflict?

theiostream:
Solved.

Source/NSPasteboardItem.m
```diff
@@ -0,0 +1,179 @@
...
+      _dataMap = [[NSMutableDictionary alloc] init];
+      _types = [[NSMutableArray alloc] init];
+    }
+
+  return self;
+}
+
+- (NSArray *)types
+{
+  return RETAIN([NSArray arrayWithArray: _types]);
```
ivucica: For future reference: if you really want to return a copied-but-immutable array with refcount 1, using `[[NSArray alloc] initWithArray: _types]` would avoid adding the object to the autorelease pool in the first place.

Also, you could use `[_types copy]`.

That said, while I am going to approve this, this seems like a leak. If I were calling `[itemFromPasteboard types]`, [based on the method name](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/MemoryMgmt/Articles/mmRules.html) I would not expect getting an object I have to `-release` myself. 

I will submit a separate commit removing `RETAIN`.

Source/NSTextList.m
```diff
@@ -40,6 +40,7 @@ - (id) initWithMarkerFormat: (NSString *)format
 {
   ASSIGN(_markerFormat, format); 
   _listOptions = mask;
+  _startingItemNumber = 1;
```
ivucica: We are not serializing or deserializing this. I think that's bad.

`-isEqual:` has also not been updated.

We should also add a TODO into `-markerForItemNumber:` to take `_startingItemNumber:` into account.

PR #8, opened 2017-08-02
Add German translation

alexmyczko:


ColorPickers/German.lproj/StandardPicker.strings
```diff
@@ -0,0 +1,46 @@
+/***
+Swedish.lproj/StandardPicker.strings
```
ivucica: This is not Swedish.

ColorPickers/German.lproj/StandardPicker.strings
```diff
@@ -0,0 +1,46 @@
...
+updated by make_strings 2002-05-23 16:24:57 +0200
```
ivucica: Please remove this date, it is clearly inaccurate.

ivucica: This date is incorrect and the comment superfluous, please remove.

ColorPickers/German.lproj/StandardPicker.strings
```diff
@@ -1,6 +1,6 @@
 /***
-Swedish.lproj/StandardPicker.strings
-updated by make_strings 2002-05-23 16:24:57 +0200
+German.lproj/StandardPicker.strings
+updated by make_strings 2017-08-02 16:24:57 +0200
```
ivucica: This has not been updated by make_strings which is why you can eliminate this.

ivucica:
Per discussion in IRC, as you do not use Git on the command line, I will accept these patches, squash them and write a commit description myself.

fredkiefer:
I don't think we need a copyright assignment for these few translations. Please go ahead an merge this nice little patch.

ivucica:
I got a confirmation of copyright assignment.

PR #9, opened 2017-08-04
chore: add Travis CI build to GUI

theiostream:
Add a Travis CI build to gnustep-gui to check if everything is really
okay before accepting patches.

Recent events have shown that bad patches have passed through the review
process, specially when it comes to breaking compatibility with GCC. A
CI routine would distinctly avoid these cases.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,48 @@
+language: cpp
+dist: trusty
+compiler:
+    - clang
+    - gcc
+env:
+    - LIBRARY_COMBO=gnu-gnu-gnu
+    - LIBRARY_COMBO=ng-gnu-gnu
+    - LIBRARY_COMBO=ng-gnu-gnu BASE_ABI=--disable-mixed-abi
+matrix:
+    exclude:
+        - compiler: gcc
+          env: LIBRARY_COMBO=ng-gnu-gnu
+        - compiler: gcc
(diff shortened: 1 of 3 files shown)


theiostream:
@fredkiefer @rmottola

.travis.yml
```diff
@@ -0,0 +1,48 @@
...
+          if [ $CC = 'gcc' ];
+          then
+            sudo apt-get install -y gobjc;
+          fi;
+          sudo apt-get install -y libobjc-4.8-dev;
+        else
+          sudo apt-get install -y libkqueue-dev libpthread-workqueue-dev;
+        fi;
+install: ./travis-deps.sh
+before_script: >
```
ivucica: Happy to merge as-is (rebase please, though. alternatively I'll do it when I get a chance).

Quick check, though, why the distinction between `before_script` and `script`? I tried quickly looking at Travis docs, but I can't find what makes the difference?

ivucica:
@theiostream Maybe we can also use 'reviewers' field and 'assignees' field, I don't know what exactly they do -- but they might be interesting?

rmottola:
As with Fred, I am no expert with Travis, but I like the idea. It should catch the most basic "build"errors .

ivucica:
Merged as 83e505c8a8759baa039318f06250cc6cac272d3b

PR #12, opened 2017-10-28
Named image unarchiving

iamleeg:
2017-10-28  Graham Lee  <graham@iamleeg.com>

	* Source/NSImage.m (initWithCoder:): restore image name when
	loading an archived image with a name that can't be resolved at
	load time. Fixes Gorm bug #46317.

### DIFF
--- a/Source/NSImage.m
+++ b/Source/NSImage.m
@@ -484,8 +484,8 @@ repd_for_rep(NSArray *_reps, NSImageRep *rep)
 	    initByReferencingFile: path];
           if (image != nil)
             {
-              [image setName: realName];
-              image->_flags.archiveByName = YES;
+	      [image setName: realName];
+	      image->_flags.archiveByName = YES;
               AUTORELEASE(image);
             }
         }
@@ -1967,15 +1967,24 @@ static NSSize GSResolutionOfImageRep(NSImageRep *rep)
 - (id) initWithCoder: (NSCoder*)coder
 {
   BOOL flag;
-
+  NSImage *replacementImage;
+  NSString *imageName;
+  
   _reps = [[NSMutableArray alloc] initWithCapacity: 2];
   if ([coder allowsKeyedCoding])
     {
       if ([coder containsValueForKey: @"NSName"])
         {
           RELEASE(self);
-          return RETAIN([NSImage imageNamed:
-            [coder decodeObjectForKey: @"NSName"]]);
+	  imageName = [coder decodeObjectForKey: @"NSName"];
+          replacementImage = [NSImage imageNamed: imageName];
+	  if (replacementImage)
+	    {
+	      return RETAIN(replacementImage);
+	    }
+	  replacementImage = [[NSImage alloc] init];
+	  [replacementImage setName: imageName];
+	  replacementImage->_flags.archiveByName = YES;
         }
       if ([coder containsValueForKey: @"NSColor"])
         {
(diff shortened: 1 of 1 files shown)


ivucica:
Small note: This was lacking a `ChangeLog` update...

tedge:
Hi,

    This patch (which has now been merged to the trunk) looks like it  
will cause a crash if trying to load an unknown named-image from a  
keyed archive:

-[NSImage initWithCoder:] (NSImage.m, lines 1976-1998):
----
       if ([coder containsValueForKey: @"NSName"])
         {
           RELEASE(self);
           imageName = [coder decodeObjectForKey: @"NSName"];
           replacementImage = [NSImage imageNamed: imageName];
           if (replacementImage)
             {
               return RETAIN(replacementImage);
             }
           replacementImage = [[NSImage alloc] init];
           [replacementImage setName: imageName];
           replacementImage->_flags.archiveByName = YES;
         }
----

    self is released in the second line (probably deallocating it),  
but if the named image isn't found ([NSImage imageNamed: imageName]  
returns nil), control passes through the rest of the above scope, to  
the end of the method, which returns (the now-deallocated) self. (The  
new image that's allocated to replace the original instance is stored  
in replacementImage instead of self, and isn't accessed again).

    One fix would be to store the reallocated image in self instead of  
replacementImage (which is how it's done in the non-keyed archive  
logic, later in the same method); Alternatively, I'd suggest delaying  
the release of self until after verifying that the named image was  
found, which avoids having to reallocate if it falls through:

----
       if ([coder containsValueForKey: @"NSName"])
         {
           imageName = [coder decodeObjectForKey: @"NSName"];
           replacementImage = [NSImage imageNamed: imageName];
           if (replacementImage)
             {
               RELEASE(self);
               return RETAIN(replacementImage);
             }
           [self setName: theName];
           self->_flags.archiveByName = YES;
         }
----


    Another thing regarding -[NSImage initWithCoder:] (unrelated to  
the patch): Is it missing a 'self = [self init];' at the beginning?

Cheers,

Josh

iamleeg:
@ivucica thanks, I discovered guidance that says [not to diff the ChangeLog](http://www.gnustep.org/developers/bugs.html). Is that no longer the case?

@tedge that's a great catch, thanks for spotting it! I'd intended to return the `replacementImage` and clearly forgot to add that. As Gorm doesn't use keyed archiving I hadn't spotted the problem when I tested.

ivucica:
>  thanks, I discovered guidance that says not to diff the ChangeLog. Is that no longer the case?

From context, I suspect that refers to the "send patches to mailing list" approach of contributing. I highly disagree with this advice even in this context, as it makes it harder for the maintainer to apply a patch with `/usr/bin/patch`.

In case of Git, it makes even less sense: it means whoever merges your code using automation (such as, but not limited to, GitHub's built in system for merging PRs), they must go out of their way to manually apply your ChangeLog entry, which is too easy to omit.

Source/NSImage.m
```diff
@@ -2055,7 +2064,17 @@ - (id) initWithCoder: (NSCoder*)coder
           NSString *theName = [coder decodeObject];
 
           RELEASE(self);
-          self = RETAIN([NSImage imageNamed: theName]);
+	  replacementImage = [NSImage imageNamed: theName];
```
ivucica: Side remark, this could have been `self`, too.

Source/NSImage.m
```diff
@@ -1967,15 +1967,24 @@ - (void) encodeWithCoder: (NSCoder*)coder
 - (id) initWithCoder: (NSCoder*)coder
 {
   BOOL flag;
-
+  NSImage *replacementImage;
```
ivucica: Side remark, I'm not aware of a compiler in use in 2017 that would mind if `replacementImage` and `imageName` were declared closer to their actual place of use.

ivucica:
Oh, also, some tabs in place of spaces have crept in...

Source/NSImage.m
iamleeg: I admit to not liking the 'locals at the top' convention even while I choose to follow it.

PR #13, opened 2017-10-29
correct decoding of named images from keyed archivers

iamleeg:
As @tedge noted on the original pull request for this work, I didn't return the correct object in the keyed archiver case.


ivucica:
If I understand this correctly, this should work, though @tedge had some more interesting suggestions (such as 'let's not deallocate self in the first place until we are sure we have something to replace the image with').

We can merge this and think about cleaning it up later, though.

ivucica:
(If this was at work, I also would have liked to see a test to cover this, but GNUstep does not have the practice of enforcing tests to be written at every commit -- and gnustep-gui does not have it in particular. :) )

PR #15, opened 2018-03-29
Document methods I have recently used in NSWindowController

iamleeg:


iamleeg:
I can't replicate that build failure locally :(, is anyone able to help out?

PR #16, opened 2018-05-27
Add 'id _coreAnimationData' variable to AppKit/Headers/NSView.h.

whiteShtef:
The variable will point to another class instance that will
contain data necessary data for future implementation of methods
regarding CA layers.

### DIFF
--- a/Headers/AppKit/NSView.h
+++ b/Headers/AppKit/NSView.h
@@ -130,6 +130,7 @@ typedef enum _NSFocusRingType {
   id _boundsMatrix;
   id _matrixToWindow;
   id _matrixFromWindow;
+  id _coreAnimationData;
 
   NSView* _super_view;
 PACKAGE_SCOPE
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+2018-06-04 Stjepan Brkić <stjepanbrkicc@gmail.com>
+
+	* Headers/AppKit/NSView.h: Add 'id _coreAnimationData' variable.
+	The variable will point to another class instance that will
+	contain data necessary data for future implementation of methods
+	regarding CA layers.
+
 2018-03-30 Fred Kiefer <FredKiefer@gmx.de>
 
 	* Source/NSDocument.m (-revertToContentsOfURL:ofType:error:): Add


Headers/AppKit/NSView.h
```diff
@@ -130,6 +130,7 @@ typedef enum _NSFocusRingType {
   id _boundsMatrix;
   id _matrixToWindow;
   id _matrixFromWindow;
+  id _coreAnimationBootstrapData;
```
ivucica: I think just `_coreAnimationData` is enough? What do you think?

ivucica:
@fredkiefer I suggested to @whiteShtef to take this approach; in case we need more than the variables we accounted for (`_wantsLayer` and `_layer`), this will allow for the required space without changing `NSView`'s struct any further, and it will isolate these changes into the helper library.

ivucica:
I reported build failure (the crashes in clang 5) to libobjc2. https://github.com/gnustep/libobjc2/issues/72

Headers/AppKit/NSView.h
fredkiefer: I am not sure about this. The code in NSView will need to decide when the execution should be transferred to the CA code. For this it needs more access to the internal structure. We could either define a proper interface for that or give it access to the data itself.
The current solution with just one id element won't work. I suggest to think about the interaction between gui and CA first and come up with a solution that covers this.

ivucica:
I'll reply as this was done at my suggestion.

The code will be implemented as a class category. Does that not let it
access internal data?

_coreAnimationBootstrapData (note that I don't like the name of the ivar)
would be just a model class, and it would not contain any significant code.
Actual interesting code will be in a class category of NSView. Class
category would be aware of the actual underlying (but opaque) model class
and be able to trivially access the data either ->_asIvars, or [through
methodCalls].

Adding one id element still lets Stjepan define as many variables as he
wants. My concern why I proposed this is to (1) avoid changing structure
layout unnecessarily no matter how many ivars may be needed for CA
integration, (2) avoid polluting NSView with CA related ivars (when
gnustep-gui doesn't actually care much about CA, and rightly so), (3) avoid
synchronizing changes between CA and gnustep-gui.

Once this change is made, nearly all of the integration development can
happen in the libs-quartzcore repo.

Does this address your concerns?

ivucica:
n.b. I am very amused at how this did not get threaded into
https://github.com/gnustep/libs-gui/pull/16#discussion_r191121370 -- I
cannot log into Github at this time, so I tried replying over email.
On Mon, May 28, 2018 at 6:55 PM Ivan Vučica <ivan@vucica.net> wrote:

> I'll reply as this was done at my suggestion.

> The code will be implemented as a class category. Does that not let it
> access internal data?

> _coreAnimationBootstrapData (note that I don't like the name of the ivar)
> would be just a model class, and it would not contain any significant
code.
> Actual interesting code will be in a class category of NSView. Class
> category would be aware of the actual underlying (but opaque) model class
> and be able to trivially access the data either ->_asIvars, or [through
> methodCalls].

> Adding one id element still lets Stjepan define as many variables as he
> wants. My concern why I proposed this is to (1) avoid changing structure
> layout unnecessarily no matter how many ivars may be needed for CA
> integration, (2) avoid polluting NSView with CA related ivars (when
> gnustep-gui doesn't actually care much about CA, and rightly so), (3)
avoid
> synchronizing changes between CA and gnustep-gui.

> Once this change is made, nearly all of the integration development can
> happen in the libs-quartzcore repo.

> Does this address your concerns?

ivucica:
@whiteShtef Did you mean to close this PR? You can update the `whiteShtef:master` branch instead of creating a new PR.

ivucica:
I would ordinarily ask for a rebase, but let's proceed by creating a merge commit.

PR #18, opened 2018-07-20
Fix app icon in dock

gcasa:
Here is the gui portion of the change.

### DIFF
--- a/Source/NSApplication.m
+++ b/Source/NSApplication.m
@@ -369,6 +369,7 @@ struct _NSModalSession {
  
 @interface NSApplication (Private)
 - (void) _appIconInit;
+- (void) _loadAppIconImage;
 - (NSDictionary*) _notificationUserInfo;
 - (void) _openDocument: (NSString*)name;
 - (id) _targetForAction: (SEL)aSelector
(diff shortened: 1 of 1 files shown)


fredkiefer:
Splitting up the loading of the application icon and the display of the icon window is of course fine. So you may go ahead and merge this change.
What I still would like to understand is how this fixes the issue. Could you please add a log statement in line 3855 to see whether this ever gets called. I would expect that this isn't the case.
But then there must be some place where we ask for the application icon between the call to  _loadAppIconImage and the one to _appIconInit.

gcasa:
I will do this in a couple of hours. It is very early morning here. :)

gcasa:
It fixes the issue because, since I can call _loadAppIconImage without initializing a window, it can be done prior to the backend call _createAppIconPixmaps (which is, in turn, called by the GSWindowServer call which occurs early in [NSApplication _init].   When the _create.. method is called the icons are set.  Prior to this change that was not the case.   The reason I was calling _loadIconImage in _appIconInit is that it simply seems appropriate.  I'm aware that in the normal case this won't be called,.   I will comment it out for now.

PR #19, opened 2018-07-21
NSGraphicsContext: When creating w/ graphics port, actually pass it in.

ivucica:
Pass the requested graphics port into the newly created graphics
context using `-GSSetDevice:::.`

---

Hi Fred,

While trying to help @whiteShtef with his project, this seemed like something that prevented Opal backend from using the correct passed `graphicsPort`. Can you take a look and see if this would be the right way to go about this?

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,12 @@
+2018-07-24 Ivan Vucica <ivan@vucica.net>
+
+	* Source/NSGraphicsContext.m (+graphicsContextWithGraphicsPort:flipped:):
+	If the underlying graphics context supports it, pass nil to GSSetDevice.
+
+	In Opal backend, this is intended to trigger creation of backing surface
+	no matter that we don't have a window device. This should make the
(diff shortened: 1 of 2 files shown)


ivucica:
This is wrong.

`-GSSetDevice:::` will eventually invoke `-[OpalContext GSSetDevice:::]` which will invoke `[[OpalSurface alloc] initWithDevice:context:]` using the passed device.

This device is expected to be a `gswindow_device_t*` by the time it reaches `initWithDevice:context:`.

This assumption inside `initWithDevice:context:` does come with a FIXME attached to it, that we should not presume the device to be a window device, but still doesn't mean we will incorrectly pass a CGContext to something expected to be a window device.

Please ignore this PR for now; it should serve as a reminder to me and @whiteShtef  to investigate how to do this properly.

fredkiefer:
I have a version of this change on my local machine that fits better. I just need to find the time to actually commit it.

PR #20, opened 2018-09-10
Update say.m

alexmyczko:
propose some updates to say

### DIFF
--- a/Tools/say/say.m
+++ b/Tools/say/say.m
@@ -2,6 +2,14 @@
 #include <unistd.h>
 #include <getopt.h>
 
+/*
+Would be great to have some control over the voice of say, taken from the Amiga manual:
+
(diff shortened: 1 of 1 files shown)


fredkiefer:
Sorry, I don't understand what you expect from this patch. It does in no way improve the current code, not even by documenting it. If you want to make a change request, I would recommend to do so on savannah (http://savannah.gnu.org/projects/gnustep/)
BTW you should already be able to change the voice with the parameter -v, but I never tested this.

alexmyczko:
How to use the -v option? the manual page nor -h give any information about it. Oh and the PR were meant to be issues, however issues on libs-gui seem to be magically disabled

ivucica:
(Sidenote: I have updated the project description to say 'Our issue tracker is on Savannah; please file bugs there.')

PR #21, opened 2018-09-10
Update set_show_service.m

alexmyczko:


fredkiefer:
Again, GitHub is not intended to be used for change request. Please post these on savannah.
One way to see the available services is to look into the file ".GNUstepServices". But this really might be a valid request.

PR #24, opened 2019-03-27
Lowering down NSFloatingWindowLevel number by one

trunkmaster:
This is a complementary change to https://github.com/gnustep/libs-back/pull/6.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,7 +1,7 @@
 2019-03-27  Sergii Stoian  <stoyan255@gmail.com>
 
 	* Headers/AppKit/NSWindow.h:
-	Lovering down NSFloatingWindowLevel number by one to distinguish
+	Lowering down NSFloatingWindowLevel number by one to distinguish
 	floating panels from menus.
 
 2019-02-26 Fred Kiefer <FredKiefer@gmx.de>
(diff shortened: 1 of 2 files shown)


fredkiefer:
From the GNUstep side I see no problem with this change. What we have to think about is who WindowMaker will behave differently due to this change. I am not using this window manager myself. Somebody with more knowledge will have to decide. If this change causes issues there we could map to different values specifically for WindowMaker. This would allow us to finally switch to use the Cocoa values for that enumerator.

trunkmaster:
Basically, this enum in NSWindow.h is not the same as in WindowMaker.h even without this patch.
WindowMaker uses values from WindowMaker.h to level it's own interface elements (WINGs toolkit).
Layering windows to each other executed by comparing numbers. That's simple.

So we have 2 options:
1. Align NSWindow.h contents to WindowMaker.h to be strictly compatible - I'll update patch
2. Switch to Cocoa values and WindowMaker devs should adopt these changes to WM

The first option is the best to my mind. I'll update patch.

trunkmaster:
WindowMaker doesn't use `NSFloatingWindowLevel` at all. It's simply defined in WindowMaker's GNUstep.h as `WMFloatingWindowLevel`. Window layering processed by comparing level numbers. There will be no harm to WindowMaker users.

trunkmaster:
@fredkiefer, what's next?

fredkiefer:
If WindowMaker doesn't care about the actual numbers we are using for the window levels, this patch is fine and we even could switch to the real values defined in Cocoa.

PR #25, opened 2019-04-02
Handle `-autolaunch YES` application argument (GUI part)

trunkmaster:
Do not activate application on first run of `-activateIgnoringOtherApps:` if commandline argument `-autolaunch YES` was specified.
This is a change complementary to https://github.com/gnustep/libs-back/pull/7.

### DIFF
--- a/Source/NSApplication.m
+++ b/Source/NSApplication.m
@@ -1259,6 +1259,23 @@ static NSSize scaledIconSizeForSize(NSSize imageSize)
  */
 - (void) activateIgnoringOtherApps: (BOOL)flag
 {
+  NSUserDefaults *defaults = [NSUserDefaults standardUserDefaults];
+  NSString       *autolaunch = [defaults objectForKey: @"autolaunch"];
+
+  /* Application was executed with argument '-autolaunch YES'. Set argument 
+   * value to "NO" for future calls to `activateIgnoringOtherApps:`. */
+  if (autolaunch && [autolaunch isEqualToString: @"YES"])
+    {
+      NSMutableDictionary *args;
+      args = [[defaults volatileDomainForName: NSArgumentDomain] mutableCopy];
+      [defaults removeVolatileDomainForName: NSArgumentDomain];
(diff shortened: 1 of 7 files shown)


Source/NSApplication.m
```diff
@@ -1259,6 +1259,24 @@ - (void) dealloc
  */
 - (void) activateIgnoringOtherApps: (BOOL)flag
 {
+  NSUserDefaults *defaults = [NSUserDefaults standardUserDefaults];
+  NSString       *autolaunch = [defaults objectForKey: @"autolaunch"];
+
+  /* Application was executed with argument '-autolaunch YES'. Set argument 
+   * value to "NO" for future calls to `activateIgnoringOtherApps:`. */
+  NSLog(@"autolaunch = %@", autolaunch);
```
fredkiefer: You probably did forget to remove this log statement.

This whole check of the user defaults which gets done on each call of this method looks rather slow to me. Shouldn't we read this value once and keep it in a static variable and only check (and set) that from here?

trunkmaster: I understand your concern but `-activateIgnoringOtherApps:` method called only on application activation. I think such complication is not the case for optimization.

Yes, I forgot to remove temporary debug output. I'll update code right now.

trunkmaster: I forgot to mention that leaving `autolaunch` argument in defaults domain is intentional. If some application need to know if it was autolaunched at startup (for example in -applicationDidFinishLaunching:), one should check if [defaults objectForKey:@"autolaunch"] is not `nil`.

trunkmaster: Hm... I don't know how to update patch in PR. @fredkiefer, could you help me?

trunkmaster: @fredkiefer , PR has added commit. Please, merge if it's ok.

trunkmaster:
@fredkiefer, I've optimized code, but some old commits displayed in this PR and my fork displayed as "unknown" (after deletion and fork again to make changes clear). File changes look correct. Do I need to add new PR?

PR #26, opened 2019-04-05
Added support for WindowMaker's WMFHideApplication action.

trunkmaster:
This is a first part of "Hide Application" functionality (the next in the -back), but I describe changes here.
Problem: Application can be hidden in 2 ways: fist - using application's "Hide" menu item or key equivalent Cmd-h (Alt-h); second - WindowMaker's right-click on miniaturize titlebar button. Any of these actions are independent of each other - there's a weak/absent interconnection between application and window manager (WindowMaker). As a result, if you hide app in a first way window manager knows nothing  besides withdrawing of windows from screen. If you hide app with window manager application is not considered as hidden inside app's code.
Solution:
1. When [NSApp hide] method called application first tries to send client message to WindowMaker. If this message succeds, application marked as hidden (sets ivar, draws dot on appicon) without ordering out of windows (it will be done by WindowMaker). WindowMaker withdraws windows from screen and switch focus to the next application in focus ring. Otherwise (no support from window manager), application hides it's windows by itself.
2. If the source of "Hide" action is WindowMaker - client message will be send to GNUstep Back, NSEvent generated with subtype `GSAppKitAppHide` and application now knows that it was hidden.

The -back part of changes is here https://github.com/gnustep/libs-back/pull/9.

### DIFF
--- a/Source/GSDisplayServer.m
+++ b/Source/GSDisplayServer.m
@@ -601,7 +601,8 @@ GSCurrentServer(void)
   [self subclassResponsibility: _cmd];
 }
 
-/** Hides all application windows */
+/** Ask the window manager to hide all the application windows for us. 
+    Return whether they have been hidden. */
 - (BOOL) hidewindow: (int) win
 {
   return NO;
--- a/Source/GSDisplayServer.m
+++ b/Source/GSDisplayServer.m
@@ -603,7 +603,7 @@ GSCurrentServer(void)
 
 /** Ask the window manager to hide all the application windows for us. 
     Return whether they have been hidden. */
-- (BOOL) hidewindow: (int) win
+- (BOOL) hideApplication: (int) win
 {
   return NO;
 }
--- a/Source/GSDisplayServer.m
+++ b/Source/GSDisplayServer.m
@@ -601,6 +601,12 @@ GSCurrentServer(void)
   [self subclassResponsibility: _cmd];
 }
 
+/** Hides all application windows */
+- (void) hidewindow: (int) win
+{
+  [self subclassResponsibility: _cmd];
+}
+
 /** Returns YES if the application should create the miniwindow counterpart
     to the full size window and own it. Some display systems handle the
     miniwindow themselves. In this case the backend subclass should
--- a/Source/GSDisplayServer.m
+++ b/Source/GSDisplayServer.m
@@ -602,9 +602,9 @@ GSCurrentServer(void)
 }
 
 /** Hides all application windows */
-- (void) hidewindow: (int) win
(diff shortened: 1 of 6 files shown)


Source/GSDisplayServer.m
```diff
@@ -601,6 +601,12 @@ - (void) miniwindow: (int) win
   [self subclassResponsibility: _cmd];
 }
 
+/** Hides all application windows */
+- (void) hidewindow: (int) win
+{
+  [self subclassResponsibility: _cmd];
```
fredkiefer: I think it is better to leave this method empty. You only provided an implementation for the x11 backend on Windows this will result in an exception and hiding no longer will work there.

fredkiefer: Also hidewindow: has a boolean return value. This should be respected and NO returned.

Source/NSApplication.m
```diff
@@ -2496,30 +2496,37 @@ - (void) hide: (id)sender
 	      [_hidden_main resignMainWindow];
 	    }
 	  
-	  windows_list = GSOrderedWindows();
-	  iter = [windows_list reverseObjectEnumerator];
+          // Sends -hidewindow: to the main menu. If window manager supports
+          // _GNUSTEP_WM_HIDE_APP atom - call succeeds - our windows will be
```
fredkiefer: As guy is agnostic about the different backend implementations we should not be talking about specific X atoms here. Rather whether the backend handles the hiding (via the window manager) or we need to do it here.

Source/GSDisplayServer.m
trunkmaster: Fixed.

Source/NSApplication.m
trunkmaster: This comment is just explain logic. Do I need to remove comment?

fredkiefer: Rewrite not remove :-)
This actually is not what the window backend will do and also not what the x11 backend does for most window managers.

Source/GSDisplayServer.m
```diff
@@ -601,6 +601,12 @@ - (void) miniwindow: (int) win
   [self subclassResponsibility: _cmd];
 }
 
+/** Hides all application windows */
```
fredkiefer: The comment should be something like this:
Ask the window manager to hide all the application windows for us. Return whether they have been hidden.

Source/NSApplication.m
trunkmaster: Fixed.

Source/GSDisplayServer.m
trunkmaster: Fixed.

PR #27, opened 2019-04-08
Reduce number of appicon flickering cases

trunkmaster:
Override NSWindow's `becomesKeyOnlyIfNeeded` and NSView `needsPanelToBecomeKey` in appicon implementation to reduce number of order front calls. The source of the calls is NSWindow's sendEvent:.
Steps to reproduce includes: double-click on appicon of inactive application, activate visible windows of inactive application (click on titlebar or window contents). 
Result: appicon flickers (redraws). It can be observed in `cairo` backend. `art` backend doesn't show appicon flickering.

### DIFF
--- a/Source/NSApplication.m
+++ b/Source/NSApplication.m
@@ -547,11 +547,6 @@ static NSSize scaledIconSizeForSize(NSSize imageSize)
   return YES;
 }
 
-- (BOOL) needsPanelToBecomeKey
-{
-  return NO;
-}
-
 - (void) concludeDragOperation: (id<NSDraggingInfo>)sender
 {
 }
--- a/Source/NSApplication.m
+++ b/Source/NSApplication.m
@@ -434,6 +434,11 @@ NSApplication	*NSApp = nil;
   return NO;
 }
 
+- (BOOL) becomesKeyOnlyIfNeeded
+{
+  return YES;
+}
+
 - (BOOL) worksWhenModal
 {
   return YES;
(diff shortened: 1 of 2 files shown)


Source/NSApplication.m
```diff
@@ -542,6 +547,11 @@ - (BOOL) acceptsFirstMouse: (NSEvent*)theEvent
   return YES;
 }
 
+- (BOOL) needsPanelToBecomeKey
```
fredkiefer: I don't see why this change is needed. NSView already should return NO for this method so we are just duplicating code from the super class.
The other part of this change looks good to me.

trunkmaster: OK, let me check NSView code.

trunkmaster: You are right. I've removed `needsPanelToBecomeKey` override.

trunkmaster: By the way, what window manager do you use? I need to know to do additional checks before submitting PR to -gui and -back.

fredkiefer: I normally use kwin, which comes with the KDE desktop environment. Years ago I tried to test GNUstep code with a lot of different window managers but I gave up on that long ago.

trunkmaster: I understand you. I guess you just don't see some appicon flickering (GSSuppressAppIcon?). I have some questions on Cairo backend implementation. What is better the method for it: PM or mail-list?

PR #28, opened 2019-04-09
Attributed key equivalents in menu

trunkmaster:
Fontify menu key equivalents according to the ["OpenStep User Interface Guidelines"](http://www.gnustep.org/resources/documentation/OpenStepUserInterfaceGuidelines.pdf), page 3-25, section "Application-Specific Keyboard Alternatives". Here is the example, how it looks like after applying patch:

AttributedKeyEquivalents

For info: my thoughts about key equivalents design principles for NextSpace is here https://github.com/trunkmaster/nextspace/wiki/Keyboard.

### DIFF
--- a/Source/NSMenuItemCell.m
+++ b/Source/NSMenuItemCell.m
@@ -46,6 +46,7 @@
 #import "AppKit/NSMenuItemCell.h"
 #import "AppKit/NSMenuView.h"
 #import "AppKit/NSParagraphStyle.h"
+#import "AppKit/NSStringDrawing.h"
 #import "GNUstepGUI/GSTheme.h"
 #import "GSGuiPrivate.h"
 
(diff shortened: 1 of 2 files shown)


Source/NSMenuItemCell.m
```diff
@@ -736,7 +784,23 @@ - (void) drawKeyEquivalentWithFrame: (NSRect)cellFrame
...
   else if (![[_menuView menu] _ownedByPopUp] || (_imageToDisplay == nil))
     {    
-      [self _drawText: [self _keyEquivalentString] inFrame: cellFrame];
+      if (_keyEquivalentFont != nil)
+        {
+          NSDictionary       *attrs;
+          NSAttributedString *aString;
+          
+          attrs = [NSDictionary dictionaryWithObject: _keyEquivalentFont
+                                              forKey: NSFontAttributeName];
```
fredkiefer: I am not sure, but shouldn't we also set the colour and respect the enabled/disabled state for this as well?

trunkmaster: You are right. Fixed.

Source/NSMenuItemCell.m
```diff
@@ -736,7 +784,34 @@ - (void) drawKeyEquivalentWithFrame: (NSRect)cellFrame
...
+        {
+          NSDictionary       *attrs;
+          NSArray            *attrObjects, *attrKeys;
+          NSAttributedString *aString;
+          NSColor            *aColor;
+
+          if (_cell.is_disabled)
+            aColor = [NSColor disabledControlTextColor];
+          else
+            aColor = [NSColor controlTextColor];
```
fredkiefer: Here we should use [self textColor]. (See NSCell)

trunkmaster: Thanks. Fixed.

PR #30, opened 2019-04-17
Mouse cursors handling cleanup

trunkmaster:
I've cleaned up some mouse cursors loading/getting images:
- `_setCursor:` simplified, removed code with `common_*` image names, implemented usage of NSCursor's methods;
- new colored images were added with names used in NSCursor's methods - GS*Cursor.tiff (note: color cursor images are displayed weird - fixed by https://github.com/gnustep/libs-back/pull/11);


fredkiefer:
I would prefer not to remove the unused cursor images as some themes may rely on them. Also when removing image, or even if you are just removing the usage of an image, you should also remove it from the mappings file nsmapping.strings.
Overall I like the changes in DnD, as this would allow to remove the cursor recolour method, that isn't fully implemented. What I am wondering is whether this change will lead to a better user feedback? Maybe we should involve Riccardo here?

trunkmaster:
I guess theme - by it's nature - should hold images it relies on. Mapping old (removed files)  names in nsmapping.strings is good - it's good for old names usage, isn't it?
I didn't quite understand what do you mean by "better user feedback". With my changes cursors are colorful, have good visibility. Do you mean something else?

fredkiefer:
Removed images:
We do not know who is using them and if we remove them we may cause these people trouble. For example a theme may have used these files and no, any theme is free to use images from the standard without providing these itself.

The mapping in nsmapping.strings works the other way around. It binds names to images and if these images are gone the mapping is broken. What you should do is not remove the files, just comment out or remove the mappings.

DnD:
With your changes here the user will get different feedback via the cursor. Especially in the case where a drag operation is currently over an invalid target. As I am no user experience expert I would like to have others, for example Riccardo getting involved in the decision.

rmottola:
Fred, thanks for asking. I understand the relevant code has already been merged in or at least part of it? I think the best is to have some testcases and work on them. Also, is there something we need to "solve"?
Currently, I can think of few applications that use proper pointer feedback, let's make a list
1) GWorkspace, of course, with different file operations and it appears to work fine - there is operation and also no-target feedback
2) text drag-drop, e.g. in Ink. I think it "works" but I see an unexpected icon while doing it (an X? I'd expect an insertion point or similar, 
3) Gorm, while dropping items and doing connections. Seems to work for me - a bit confusing is that the link operation has both a link cursor plus another icon which tells you if the result is possible, resulting e.g. with two link icons
4) GNUMail, during message drag-n-drop to folders

rmottola:
about removing images: I think themes should work on named images.
I just checked and Thematic shows cursor images for drag-copy drag-link and drag-forbidden but I see nothing else in system images, like hair cursor, drag hand or fleur.

fredkiefer:
@Riccardo,

no, this hasn't been merged already. This is a pull request that only lives on this specific branch. What you did see in your tests was the standard GNUstep behaviour. You will have to switch to this branch to test  and see the new behaviour.

As for the image removal, any code that is using named images with the official names should still work, what will get broken are any references to the internal file names.

rmottola:
@fredkiefer sorry - git is always confusing, I thoguht part was in a branch, part merged. Also, it is not really a branch, but a fork, right? I checked out trunkmaster:libgs-gui and did some tests.
I redid all the operations I mentioned previously (4 test apps) and they perform the same, I notice no changes.
The only change is that the icons are now coloured, which one may or may not like. I'm fine with the "red" forbid sign, the green copy sign, but the green pointer arrow is a little too much, for my taste. Try dragging some text, for example.

You mention, however different feedback to the user via the cursor which I could not reproduce. Can you detail?

trunkmaster:
@rmottola, green pointer arrow represents "Move" operation. Black arrow doesn't signals to user what will be performed with dragging object after mouse button release. Dragging text is a good example, there's no need for "X" over the green arrow pointer - when user drags some text, green arrow tells to the user that text will be moved to the new location.

@fredkiefer, I also don't quite understand what you observe when dragging over the invalid target - do you see red crossed circle?

rmottola:
@trunkmaster I understand, but I don't like this green for standard dragging operation, I would prefer, like for a text operation, a specific cursor like a crosshair or another. The difference should not be color-coded only. Color should be a useful addition, not the only difference. Also checking what windows, mac or GTK do corroborates this. Where did you get this green-cursor coding inspiration from?

However, here we are speaking about different icons, which is a different matter than the code behaviour itself which interested @fredkiefer . What would be different, in your opinion, Fred, in behaviour?

trunkmaster:
@rmottola, tastes differ... I got this inspiration from NeXTSTEP.

armm77:
OPENSTEP also handles the green pointer.

trunkmaster:
@fredkiefer, what are the next steps? What should I do to make this PR merged?

fredkiefer:
As I already wrote, you should restore the deleted images and remove them from the nsmapping.strings file. After that I would merge the pull request.

@Riccardo
The difference in feedback that I mentioned is that previously we first did choose the cursor image and recoloured it according to whether a drop was possible or not. Now we get either the green cursor or a proper cursor that shows the possible action. As the new behaviour is done for you, I don't have any objections to it.

trunkmaster:
@fredkiefer, Done.

PR #32, opened 2019-09-10
NSFileWrapper move to Foundation

gcasa:
NSFileWrapper was migrated to Foundation from AppKit in macOS 10.7.  This change implements that move for GNUstep.


Headers/AppKit/AppKit.h
```diff
@@ -141,6 +141,7 @@
 #import <AppKit/NSDocumentController.h>
 #import <AppKit/NSDrawer.h>
 #import <AppKit/NSFileWrapper.h>
```
fredkiefer: Why do you still import the old, now empty file AppKit/NSFileWrapper.h?
The same is true in a few places further down in other files.

Headers/AppKit/NSFileWrapperExtensions.h
```diff
@@ -0,0 +1,42 @@
+/* 
+   NSFileWrapper.h
```
fredkiefer: Still the wrong file name.

Source/NSFileWrapper.m
```diff
@@ -44,325 +44,7 @@
 #import "AppKit/NSImage.h"
 #import "AppKit/NSWorkspace.h"
 
-@implementation NSFileWrapper
```
fredkiefer: Looks like this whole file is still named wrong.

Headers/AppKit/AppKit.h
gcasa: Because it's not empty.  It imports the one in foundation.   This is what is done on macOS.

fredkiefer: In the next line you are importing the new extension file, which also includes the Foundation version of NSFileWrapper. So this additional line is not needed.

Source/NSFileWrapper.m
```diff
@@ -44,325 +44,7 @@
...
-  if (_wrapperType == GSFileWrapperSymbolicLinkType)
-    {
-      return YES;
-    }
-  else
-    {
-      return NO;
-    }
-}
+@implementation NSFileWrapper (NSExtensions)
```
fredkiefer: The name of the instance variable has been changed back to _icon, this needs to be changed here. This only works as this file has still the old name and never gets compiled.

gcasa: Just to be clear.... it was never ever changed TO _iconImage. :) Certainly not by me.  I changed it to _icon since you suggested it.

gcasa: This has been changed.  I think you caught me in the middle of a change here.

Source/NSFileWrapper.m
gcasa: I removed it.

PR #33, opened 2019-11-03
gitignore update according to https://www.gitignore.io/api/xcode

IOOI-SqAR:
I want to add a .xcodeproj to libs-gui so I am doing this in preparation


IOOI-SqAR:
Thanks, Fred!

PR #36, opened 2019-12-08
Incorrect string drawing fix for ART backend.

trunkmaster:
It seems that third parameter to `cache_lookup` function (useScreenFonts) is valuable for ART but not for Cairo backend. So ART backend draws strings shortened or misplaced especially in popup buttons. Also some applications menus right border missed by several pixels so it rendered invisible.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+2019-12-09  Sergii Stoian  <stoyan255@gmail.com>
+
+	* Source/NSStringDrawing.m: fixed incorrect string drawing (shortened, misplaced)
+	in ART backend. It seems that third parameter to `cache_lookup` function
+	(BOOL useScreenFonts) is valuable for ART but not for Cairo backend.
+
 2019-12-07 Fred Kiefer <FredKiefer@gmx.de>
 
 	* .travis.yml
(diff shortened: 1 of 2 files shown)


fredkiefer:
Sorry, for causing this issue.
What I still don't understand is how this affects the right border of menus.

trunkmaster:
I've made some tests with old and merged NSStringDrawing code. I've noticed difference in menu item width calculations (I tested Preferences.app):
1. Menu item width is incorrect and varies for different items (varies from 89 to 95; should be 98).
2. Cell width value is not rounded: contains something like 95.5625

As a result black right border painted at half-pixel x origin that leads to the drawing of dark gray line instead of black. Visially it looks like absent right border (it exists but is not clearly visible).

Adding this code to NSMenuView's -rectOfItemAtIndex: 
    theRect.size.width = floorf(_cellSize.width);
fixes menu border drawing.

fredkiefer:
Good point, but in this case we should not just floor one value. Instead we might use the centerScanRect: method to convert the rect into suitable values.

trunkmaster:
I'm not sure it would be correct solution. I guess rounding should be better placed in method where _cellSize is calculated. From the other hand, it will hide future errors like this. What do you think?

PR #37, opened 2019-12-10
Insertion point height fix for NSText* objects

trunkmaster:
Initial intention was to fix insertion point height in Login.app text fields. These text fields use font larger than default 12-point system font. While text field is empty, insertion point drawn as for 12-point font size. After first character was entered insertion point get correct size for large font.

This change fixes insertion point height no matter if text object is empty or not.

### DIFF
--- a/Source/GSHorizontalTypesetter.m
+++ b/Source/GSHorizontalTypesetter.m
@@ -40,6 +40,8 @@
 #import "AppKit/NSTextAttachment.h"
 #import "AppKit/NSTextContainer.h"
 #import "AppKit/NSTextStorage.h"
+#import "AppKit/NSLayoutManager.h"
+#import "AppKit/NSTextView.h"
 #import "GNUstepGUI/GSLayoutManager.h"
 #import "GNUstepGUI/GSHorizontalTypesetter.h"
 
@@ -534,9 +536,20 @@ For bigger values the width gets ignored.
     }
   else
     {
-      // FIXME These should come from the typing attributes
-      curParagraphStyle = [NSParagraphStyle defaultParagraphStyle];
-      curFont = [NSFont userFontOfSize: 0];
+      NSLayoutManager *layoutManager = [[curTextContainer textView] layoutManager];
+
+      if (layoutManager)
+        {
+          NSDictionary *typingAttributes = layoutManager->_typingAttributes;
+          curParagraphStyle = [typingAttributes
+                                  objectForKey: NSParagraphStyleAttributeName];
+          curFont = [typingAttributes objectForKey: NSFontAttributeName];
+        }
+      else
+        {
+          curParagraphStyle = [NSParagraphStyle defaultParagraphStyle];
+          curFont = [NSFont userFontOfSize: 0];
+        }
     }
   
   line_height = [curFont defaultLineHeightForFont];
--- a/Source/GSHorizontalTypesetter.m
+++ b/Source/GSHorizontalTypesetter.m
@@ -40,8 +40,6 @@
 #import "AppKit/NSTextAttachment.h"
 #import "AppKit/NSTextContainer.h"
 #import "AppKit/NSTextStorage.h"
-#import "AppKit/NSLayoutManager.h"
-#import "AppKit/NSTextView.h"
 #import "GNUstepGUI/GSLayoutManager.h"
 #import "GNUstepGUI/GSHorizontalTypesetter.h"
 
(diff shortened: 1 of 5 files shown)


Source/GSHorizontalTypesetter.m
```diff
@@ -534,9 +536,20 @@ - (void) _addExtraLineFragment
...
   else
     {
-      // FIXME These should come from the typing attributes
-      curParagraphStyle = [NSParagraphStyle defaultParagraphStyle];
-      curFont = [NSFont userFontOfSize: 0];
+      NSLayoutManager *layoutManager = [[curTextContainer textView] layoutManager];
+
+      if (layoutManager)
+        {
+          NSDictionary *typingAttributes = layoutManager->_typingAttributes;
```
fredkiefer: I think you should be using curLayoutManager directly here. This also eliminates the "if" and "else" clause and the NSTextView "import". What will be needed is a typecast to NSLayoutManager, as we are using GSLayoutManager here.
Also we should not directly access the instance variables here. Maybe we could use this method: 

-(NSRect) _insertionPointRectForCharacterIndex: (unsigned int)cindex
				 textContainer: (int *)textContainer

trunkmaster: I'll be glad if it could be possible. 
There are some problems to it in way you proposed:
1. GSLayoutManager is a NSObject sublcass. Thus typecast is not correct way.
2. In some situations NSLayoutManager is not available.
3. There's no any character in empty text container to grab typing attributes from.

I don't like direct access to instance variables also. But NSLayoutManager's _typingAttributes marked as @public along with other ivars. I don't know why but assume it was made intentionally.

fredkiefer: 1. typecast actually work pretty well with objects. All objects are just pointers and the compiler is perfectly happy with something like 
NSLayoutManager *layoutManger = (NSLayoutManager *l) curLayoutManager;
3. Yes, I know and according to the code 0 would be a suitable value for cindex here.

2. is the actual hard bit and also the reason I did not use any of the suggested solutions when I ran across this issue. Instead I just put a FIXME here. The layout manager you get back from your calls is really the same object as the one pointed to by curLayoutManager. And Alexander Malmberg, who wrote this code, wanted it to work with just the GSLayoutManager interface, so he could use other subclasses. Nobody besides him ever used this feature. So we could just drop it. Or we introduce a new method, defined on  GSLayoutManager and override it on NSLayoutManager to return the typing attributes. I am fine with both ways, we just need to be aware that your code and my first solution break the usage of GSLayoutManager here.

trunkmaster: Indeed GSLayoutManager in GSHorizontalTypesetter is a instance of NSLayoutManager.

Actually, my fist attemp was to tweak -_insertionPointRectForCharacterIndex:textContainer: method. It's simple and works for me. But I've went deeper into origin of `extra_textcontainer`...

I like your idea adding extra method to *LayoutManager. It's clear and more maintainable than typecating. I dont' quite understand why it "break the usage of GSLayoutManager here". Could you please explain what do you mean by that?

trunkmaster: Form the other hand, direct access to _typingAttributes ivar used in many places of NSTextView. I guess this descision was made due to performance reasons. Why not use it in GSHorizontalTypesetter?

trunkmaster: Fred, I've added new commit. It extends GSLayoutManager with new method `-typingAttributes`, NSLayoutManager overrides it, GSHorizontalTypesetter used new method. Please look at this version.

Source/GSHorizontalTypesetter.m
```diff
@@ -40,6 +40,8 @@
 #import "AppKit/NSTextAttachment.h"
 #import "AppKit/NSTextContainer.h"
 #import "AppKit/NSTextStorage.h"
+#import "AppKit/NSLayoutManager.h"
```
fredkiefer: I think these two lines are no longer needed.

Source/GSLayoutManager.m
```diff
@@ -3280,7 +3281,13 @@ - (void) setIntAttribute: (NSInteger)attributeTag
       else if (attributeTag == NSGlyphAttributeBidiLevel)
         g->bidilevel = anInt;
     }
- }
+}
+
+- (NSDictionary *) typingAttributes
+{
```
fredkiefer: Very nice solution!

Source/GSHorizontalTypesetter.m
trunkmaster: Done.

Source/GSLayoutManager.m
trunkmaster: Thanks!

PR #38, opened 2019-12-11
Miniwindow image and title fixes.

trunkmaster:
I've made some tweaks to miniwindow:
1. Made use of common_MiniWindowTile image instead of appicon tile image.
2. Title font size can now be configured with `NSMiniFontSize` setting in `NSGlobalDomain`.
Also I've increased default font size for NSMiniControlSize if defaults missed NSMiniFontSize setting.

Here are the screenshots to visialize changes. I placed 2 mininwindows side by side: to the left is a GNUstep miniwindow (with UNIX icon), to the right - WindowMaker miniwindow (with emacs icon).
To see the differences look at left and top border of miniwindow title and font size of title text.

Before the change:
Minwindow-0

After the change:
Minwindow-1

### DIFF
--- a/Source/NSFont.m
+++ b/Source/NSFont.m
@@ -166,10 +166,10 @@ keyForFont(NSString *name, const CGFloat *matrix,
     <item>NSFontSize                12 (System Font Size)</item>
     <item>NSLabelFontSize           (none)</item>
     <item>NSMenuFontSize            (none)</item>
-    <item>NSMiniFontSize            6</item>
+    <item>NSMiniFontSize            8</item>
     <item>NSMessageFontSize         (none)</item>
     <item>NSPaletteFontSize         (none)</item>
-    <item>NSSmallFontSize           9</item>
+    <item>NSSmallFontSize           10</item>
     <item>NSTitleBarFontSize        (none)</item>
     <item>NSToolTipsFontSize        (none)</item>
     <item>NSUserFixedPitchFontSize  (none)</item>
@@ -656,7 +656,7 @@ static void setNSFont(NSString *key, NSFont *font)
   
   if (fontSize == 0)
     {
-      fontSize = 9;
+      fontSize = 10;
     }
 
   return fontSize;
--- a/Source/NSFont.m
+++ b/Source/NSFont.m
@@ -684,7 +684,7 @@ static void setNSFont(NSString *key, NSFont *font)
   
           if (fontSize == 0)
             {
-              fontSize = 6;
+              fontSize = 8;
             }
           
           return fontSize;
--- a/Source/NSWindow.m
+++ b/Source/NSWindow.m
@@ -481,7 +481,7 @@ static NSSize scaledIconSizeForSize(NSSize imageSize)
   
   iconSize = GSGetIconSize();
   
-  tileImage = [[GSCurrentServer() iconTileImage] copy];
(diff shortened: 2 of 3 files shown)


Source/NSFont.m
```diff
@@ -684,7 +684,7 @@ + (CGFloat) systemFontSizeForControlSize: (NSControlSize)controlSize
   
           if (fontSize == 0)
             {
-              fontSize = 6;
+              fontSize = 8;
```
fredkiefer: If we change this value here, we also need to adjust the documentation above (and maybe in other places as well).
But I don't think this is a good value, it is very close to the one use used for NSSmallFontSize (9). I need to test on a Mac what the returned sizes are there.

Source/NSWindow.m
```diff
@@ -481,7 +481,7 @@ + (void) initialize
   
   iconSize = GSGetIconSize();
   
-  tileImage = [[GSCurrentServer() iconTileImage] copy];
+  tileImage = [[NSImage imageNamed:@"common_MiniWindowTile"] copy];
```
fredkiefer: Why would we use a different image here to the one used by NSApplication for the app icon?

Source/NSWindow.m
```diff
@@ -600,15 +600,17 @@ - (void) setTitle: (NSString*)aString
...
       [titleCell setSelectable: NO];
       [titleCell setEditable: NO];
       [titleCell setBordered: NO];
       [titleCell setAlignment: NSCenterTextAlignment];
-      [titleCell setDrawsBackground: YES];
-      [titleCell setBackgroundColor: [NSColor blackColor]];
+      [titleCell setDrawsBackground: NO];
       [titleCell setTextColor: [NSColor whiteColor]];
-      [titleCell setFont: [NSFont systemFontOfSize: 8]];
+      fontSize = [NSFont systemFontSizeForControlSize: NSMiniControlSize];
```
fredkiefer: Isn't NSSmallControlSize more suited here?

Source/NSFont.m
trunkmaster: I think 6 is just too small to render some readable text. I can drop this part cause I use 9 for NSMiniFontSize and 10 for NSSmallFontSize. I'll test these values on Mac too.

Source/NSWindow.m
trunkmaster: I attached images to PR. Miniwindow icon has different border around title (black background). With my change title doesn't draw background and shouldn't draw border around title to gain the same result as on screenshots.

Source/NSWindow.m
trunkmaster: Maybe... I use the smallest type of font. I use NSSmallControlSize for labels like Workspace's FileViewer "10GB available on hard disk" label - it should be bigger and differ from the smalest one.

Source/NSFont.m
trunkmaster: On MacOS 10.15 NSControlSizeMini == 9, NSControlSizeSmall == 11. BTW, NSSmallControlSize and NSMiniControlSize is deprecated since 10.12.

fredkiefer:
Sorry for replying so late. My Mac was blocked for almost two days with a Xcode update.
I can confirm your values Apple uses 9, 11, and 13 as the font sizes. As we use 12 as the default font size, 10 and 8 would be suitable values for small and mini. We will just need to adopt the documentation further up in the file as well.

trunkmaster:
Done.

PR #39, opened 2019-12-14
Xib document

fredkiefer:
This is an attempt to integrate the changes byEggplant to implement decoding of XIB documents in newer format.
I tried to clean up and simplify this code. I also extended it to handle NSToolbar and fixed a few issues.
This change also includes an almost unrelated change that brings over the shortening of strings when displayed in a NSCell with a smaller rectangle.
Also I did remove trailing whitespace, which results in a few files showing lots of changes. Please switch off whitespace changes when reviewing these.


Source/GSXib5KeyedUnarchiver.m
```diff
@@ -0,0 +1,3391 @@
...
+#import "AppKit/NSSplitView.h"
+#import "AppKit/NSTableColumn.h"
+#import "AppKit/NSTableHeaderView.h"
+#import "AppKit/NSTableView.h"
+#import "AppKit/NSTabView.h"
+#import "AppKit/NSToolbarItem.h"
+#import "AppKit/NSView.h"
+#import "GSCodingFlags.h"
+
+//#define DEBUG_XIB5
```
ivucica: Hm, I don't really like commented-out code even if it's there for documentation.

Also, `#ifdef` IMHO should often be an `#if` based on the value of the macro, not an `#ifdef`. Which would make it possible to `#define DEBUG_XIB5 0` here (possibly with an `#ifndef DEBUG_XIB5` to allow to pass the `-D` flag to the compiler to change this).

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -0,0 +1,3391 @@
...
+- (NSString *) userLabel
+{
+  return _userLabel;
+}
+
+@end
+
+@interface NSWindowTemplate5 : NSWindowTemplate
+{
+  BOOL _visibleAtLaunch;
```
ivucica: Can't `NSWindowTemplate` contain these two additional values? Based on the comment, I see why `IBActionConnection5` would be necessary, but have the value names changed for `NSWindowTemplate` too? Or are these just new values?

Or are we attempting to avoid to _serialize_ these into NIBs and pre-XIB5 files? I don't see serialization code, so that's hard to judge.

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -0,0 +1,3391 @@
...
+                                           @"isSeparatorItem", @"NSIsSeparator",
+                                           //@"systemMenu", @"NSName",
+                                           @"customClass", @"NSClassName",
+                                           @"catalog", @"NSCatalogName",
+                                           @"name" , @"NSColorName",
+                                           @"selectedItem", @"NSSelectedIndex",
+                                           @"pullsDown", @"NSPullDown",
+                                           @"prototype", @"NSProtoCell",
+                                           @"metaFont", @"IBIsSystemFont",
+                                           //@"headerView", @"NSHeaderClipView",
```
ivucica: Do we know why `headerView`, `minColumnWidth` et al are commented out?

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -0,0 +1,3391 @@
...
+          RETAIN(ClassNamePrefixes);
+
+          XmlReferenceAttributes = [NSArray arrayWithObjects: @"headerView", @"initialItem", nil];
+          RETAIN(XmlReferenceAttributes);
+
+          XmlConnectionRecordTags = [NSArray arrayWithObjects: @"action", @"outlet", @"binding", nil];
+          RETAIN(XmlConnectionRecordTags);
+
+          // These cross-reference from the OLD key to the NEW key that can be referenced and it's value
+          // or object returned verbatum.  IF an OLD XIB key does not exist and contains the 'NS' prefix
```
ivucica: s/verbatum/verbatim

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -0,0 +1,3391 @@
...
+          ClassNamePrefixes = [NSArray arrayWithObjects: @"NS", @"IB", nil];
+          RETAIN(ClassNamePrefixes);
+
+          XmlReferenceAttributes = [NSArray arrayWithObjects: @"headerView", @"initialItem", nil];
+          RETAIN(XmlReferenceAttributes);
+
+          XmlConnectionRecordTags = [NSArray arrayWithObjects: @"action", @"outlet", @"binding", nil];
+          RETAIN(XmlConnectionRecordTags);
+
+          // These cross-reference from the OLD key to the NEW key that can be referenced and it's value
```
ivucica: s/it's/its

Also, do we care about 80ch/line restriction?

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -0,0 +1,3391 @@
...
+          // or object returned verbatum.  IF an OLD XIB key does not exist and contains the 'NS' prefix
+          // the key processing will strip the 'NS' prefix, make the first letter lowercase then check
+          // whether that key exists and use it's presence during 'containsValueForKey:' processing, and
```
ivucica: s/it's/its

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -0,0 +1,3391 @@
...
+          // use it's value for 'decodeXxxForKey:' processing.  So, the keys here should ONLY be those
```
ivucica: s/it's/its

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -0,0 +1,3391 @@
...
+
+  return self;
+}
+
+@end
+
+
+@implementation GSXib5KeyedUnarchiver
+
+static NSDictionary *XmltagToObjectClassCrossReference = nil;
```
ivucica: Wondering about capitalization here. `XmlTagToObjectClassCrossReference`? `xmltagToObjectClassCrossReference`? `xmlTagToObjectClassCrossReference`? 

Reading the line such as 

```
if ([XmltagsToSkip containsObject: elementName] == NO)
```

makes me think `xmlTagsToSkip` and `xmlTagToObjectClassCrossReference` would be better.

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -0,0 +1,3391 @@
...
+                                                       //@"decodeModifierMaskForElement:", @"modifierMask",
+                                                       @"decodeTableViewGridLinesForElement:", @"tableViewGridLines",
+                                                       nil];
+          RETAIN(XmlTagToDecoderSelectorMap);
+
+          // These define XML attribute keys (i.e. '<object key="name" key="name" ...') to an associated decode method...
+          // The associated decode method may process MULTIPLE keyed attributes as in such cases as
+          // decoding the integer flag masks...
+          XmlKeyToDecoderSelectorMap = [NSDictionary dictionaryWithObjectsAndKeys:
+                                                       @"decodeIntercellSpacingHeightForElement:", @"NSIntercellSpacingHeight",
```
ivucica: Most of this file seems to break 80ch/line pretty nastily and confuses Github web UI a lot. (But if we are not sticking to 80ch/line, please ignore this comment).

Does it make sense to have value line + key line + blank line, instead of value+key line?

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -0,0 +1,3391 @@
...
+  if (self == [GSXib5KeyedUnarchiver class])
+    {
+      // Only check one since we're going to load all once...
+      if (XmltagToObjectClassCrossReference == nil)
+        {
+          // These define XML tags (i.e. <objects ...) that should be allocated as the
+          // associated class...
+          XmltagToObjectClassCrossReference =
+            [NSDictionary dictionaryWithObjectsAndKeys:
+                            @"NSMutableArray", @"objects",
```
ivucica: Why does this (and some of the similar code below) indent by 16 spaces? Not sure of GNU coding style off the top of my head, but going by my `$DAYJOB` thinking, four spaces should suffice for a perfectly readable codestyle.

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -0,0 +1,3391 @@
...
+                                    @"NSBorderType", @"NSBoxType", @"NSTitlePosition",
+                                    @"NSTitleCell", @"NSOffsets",
+                                    @"NSMatrixFlags", @"NSNumCols", @"NSNumRows",
+                                    @"NSSharedData", @"NSFlags", @"NSTVFlags", @"NSDefaultParagraphStyle",
+                                    @"NSpiFlags", nil];
+          RETAIN(XmlKeysDefined);
+
+          // These define XML tags (i.e. '<autoresizingMask ...') to an associated decode method...
+          XmlTagToDecoderSelectorMap = [NSDictionary dictionaryWithObjectsAndKeys:
+                                                       @"decodeTableColumnResizingMaskForElement:", @"tableColumnResizingMask",
```
ivucica: Per previous (and below) comment, this really could be indented by four spaces or such compared to the previous line. Aligning to `dictionaryWithObjectsAndKeys:` and then indenting by two additional spaces seems an odd choice.

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -0,0 +1,3391 @@
...
+                                                       //@"decodeAutoresizingMaskForElement:", @"autoresizingMask",
```
ivucica: If unnecessary, let's remove this. If possibly necessary in the future, let's clarify why it's commented out.

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -0,0 +1,3391 @@
...
+                                                    andAttributes: [NSDictionary dictionaryWithObjectsAndKeys:
+                                                                                   @"parent", @"key",
+                                                                                 nil]];
+  GSXibElement *children     = [[GSXibElement alloc] initWithType: @"nil"
+                                                    andAttributes: [NSDictionary dictionaryWithObjectsAndKeys:
+                                                                                   @"children", @"key",
+                                                                                 nil]];
+  GSXibElement *reference    = [self createReference: [element attributeForKey: @"id"]];
+
+  //[element setAttribute: @"connection" forKey: @"key"];
```
ivucica: What does this comment mean?

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -0,0 +1,3391 @@
...
+
+- (id) initForReadingWithData: (NSData*)data
+{
+  // Check data...
+  if (data == nil)
+    {
+      DESTROY(self);
+      return nil;
+    }
+#if     GNUSTEP_BASE_HAVE_LIBXML
```
ivucica: What if we have an `NSXML` implementation that doesn't have `libxml2`?

ivucica: Why four spaces?

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -0,0 +1,3391 @@
...
+- (void) _initCommon
+{
+  [super _initCommon];
+
+  _orderedObjectsDict = RETAIN([NSMutableDictionary dictionary]);
+
+  // Create our object(s)...
+  _orderedObjects      = [[GSXibElement alloc] initWithType: @"array"
+                                              andAttributes: [NSDictionary dictionaryWithObjectsAndKeys:
+                                                                             @"orderedObjects", @"key",
```
ivucica: Ugh, this indentation :(

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -0,0 +1,3391 @@
...
+                                                                           nil]];
+  _objectRecords       = [[GSXibElement alloc] initWithType: @"object"
+                                              andAttributes: [NSDictionary dictionaryWithObjectsAndKeys:
+                                                                             @"IBMutableOrderedSet", @"class",
+                                                                           @"objectRecords", @"key",
```
ivucica: This indentation for arguments is not even consistent with itself

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -0,0 +1,3391 @@
...
+              DESTROY(self);
+            }
+          else
+            {
+              NSXMLParser *theParser  = nil;
+
+              // Initialize...
+              [self _initCommon];
+
+              // Createe the parser and parse the data...
```
ivucica: `s/Createe/Create`

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -0,0 +1,3391 @@
...
+      else
+        {
+          // Test to see if this is an Xcode 5 XIB...
+          NSArray *documentNodes = [document nodesForXPath: @"/document" error: NULL];
+
+          // Need at LEAST ONE document node...we should find something a bit more
+          // specific to check here...
+          if ([documentNodes count] == 0)
+            {
+              NSWarnMLog(@"XIB is NOT a version 5 level");
```
ivucica: So, what is the option that the user has here -- try again with a different deserializer? Isn't this a problem if you can't tell the version from the extension? Would it make sense to transparently downgrade to pre-XIB5 deserializer?

Source/GSXib5KeyedUnarchiver.m
ivucica: Also, maybe `s/CrossReference/Mapping` or `s/CrossReference//` or `s/CrossReference/Map` (the last one would match how the suffix is used in name of `XmlTagToDecoderSelectorMap`).

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -0,0 +1,3391 @@
+/** <title>GSXib5KeyedUnarchiver.m</title>
```
ivucica: I think things become critical when Github refuses to render a diff because it's too large. If this were at `$DAYJOB`, I'd refuse to review even this PR until it was split up, unless justification was given. Otherwise, even the sheer size of this single file is making it difficult to think straight about this file or keep my progress.

Given this file has a few smaller classes in it for decoding various elements in XIB5, how about we split this up into an approximation of one file per class? Plus, split up the absolutely huge `GSXib5KeyedUnarchiver` into... I don't know even what. But this is way too long to make sense of it.

Source/GSXib5KeyedUnarchiver.h
```diff
@@ -0,0 +1,47 @@
...
+
+@interface GSXib5KeyedUnarchiver : GSXibKeyedUnarchiver
+{
+  GSXibElement       *_IBObjectContainer;
+  GSXibElement       *_connectionRecords;
+  GSXibElement       *_objectRecords;
+  GSXibElement       *_orderedObjects;
+  GSXibElement       *_flattenedProperties;
+  GSXibElement       *_runtimeAttributes;
+  NSMutableDictionary *_orderedObjectsDict;
```
ivucica: If we're gonna align by the length of the longest data type, this would be `NSMutableDictionary` and everything else requires one more space.

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -0,0 +1,3391 @@
...
+          ([[element elementForKey: @"keyEquivalent"] attributeForKey: @"base64-UTF8"]))
+      {
+        object = [NSNumber numberWithUnsignedInt: 0];
+      }
+    else
+      {
+        // Seems that Apple decided to omit this attribute IF certain default keys alone
+        // are applied.  If this key is present WITH NO setting then the following is
+        // used for the modifier mask...
+        object = [NSNumber numberWithU
```
ivucica: Urrrrrrr, how about `mask |= 1 << 24` and similar? (I hope I counted this right)

ivucica: Should this be commented out?

ivucica: What does `mized` mean?

Source/GSXibLoader.m
```diff
@@ -272,10 +272,19 @@ - (NSNibConnector*) nibConnector
...
   [result setSource: [self source]];
   [result setLabel: [self label]];
-  
+
   return result;
 }
 
+- (NSString*) description
+{
+  NSMutableString *description = [[super description] mutableCopy];
```
ivucica: Why not `NSString stringWithFormat: @"%@ - label: %@, source: %@, destination: %@", ...`?

Source/GSXibLoader.m
```diff
@@ -317,33 +326,39 @@ - (void) establishConnection
...
-          if (sel && [source respondsToSelector: sel]) 	 
-            { 	 
-              [source performSelector: sel withObject: destination]; 	 
-            } 	 
-          else 	 
-            { 	 
+          SEL sel;
+
+          selName = [NSString stringWithFormat: @"set%@%@:",
+                       [[label substringToIndex: 1] uppercaseString],
```
ivucica: Why two spaces more than the next line?

Source/GSXibLoader.m
```diff
@@ -742,6 +757,17 @@ - (void) dealloc
   [super dealloc];
 }
 
+- (NSString*) description
+{
+  NSMutableString *description = [[super description] mutableCopy];
```
ivucica: Same as before, no need for a `mutableCopy` when a `stringWithFormat` will do just nicely

Source/GSXibLoader.m
```diff
@@ -923,6 +948,12 @@ - (void) awake: (NSArray *)rootObjects
   NSCustomObject *object;
   NSString *className;
 
+  if ([rootObjects count] == 0)
+  {
+    NSWarnMLog(@"no root objects!!!");
```
ivucica: Three exclamation points are maybe too much

Source/NSButtonCell.m
```diff
@@ -1674,7 +1628,9 @@ - (void) encodeWithCoder: (NSCoder*)aCoder
       buttonCellFlags.changeGray = [self cellAttribute: NSChangeGrayCell];
 
       // set these to zero...
-      buttonCellFlags.unused1 = 0; // 32;
+      buttonCellFlags.inset = 0; // 32;
```
ivucica: What does //32 mean in these three lines?

Source/NSToolbar.m
```diff
@@ -583,39 +583,41 @@ - (id) initWithCoder: (NSCoder *)aCoder
...
       [self setVisible: [aCoder decodeBoolForKey: @"NSToolbarPrefersToBeShown"]];
       [self setDelegate: [aCoder decodeObjectForKey: @"NSToolbarDelegate"]];
-    } 
+    }
   else
     {
       // FIXME
     }
 
+  [self _build];
```
ivucica: A huge -- although necessary -- cleanup diff obscures the fact that  `[self _build];` is the only code change.

How does this affect `NSToolbar` behavior? Why did it work before?

Source/NSCell.m
```diff
@@ -1398,10 +1400,10 @@ - (void) setTitleWithMnemonic: (NSString*)aString
   if (r.length > 0)
     {
       NSUInteger location = r.location;
-      
-      
-      [self setTitle: [[aString substringToIndex: location] 
-                        stringByAppendingString: 
+
+
```
ivucica: Since this is being cleaned up, might as well nuke one of the extraneous newlines...

Source/NSCell.m
```diff
@@ -2966,31 +2979,70 @@ - (NSSize) _sizeText: (NSString*)title
...
                   forKey: NSForegroundColorAttributeName];
-      
+
       return AUTORELEASE([[NSAttributedString alloc]
                              initWithString: [attrStr string]
                              attributes: newAttribs]);
     }
 }
 
+
```
ivucica: A huge -- but necessary -- amount of cleanup obscures the fact that this is the new code. Not sure how I feel about this. I can't see if anything else changed in this file.

Source/NSCell.m
```diff
@@ -2966,31 +2979,70 @@ - (NSSize) _sizeText: (NSString*)title
...
+}
+
+- (NSAttributedString*) _resizeAttributedString: (NSAttributedString*)attrstring forRect: (NSRect)titleRect
+{
+  // Redo string based on selected truncation mask...
+  NSMutableAttributedString *mutableString = AUTORELEASE([attrstring mutableCopy]);
+  NSString *ellipsis = @"...";
+  NSLineBreakMode mode = [self lineBreakMode];
+  // This code shortens the string one character at a time.
+  // To speed it up we start off propertional:
```
ivucica: What's `propertional`? If it means `proportional`, I don't think this code uses proportional size, it seems to start with fixed character sizes.

Source/NSCell.m
```diff
@@ -3021,21 +3078,27 @@ - (void) _drawText: (NSString*)aString  inFrame: (NSRect)cellFrame
 
   attributes = [self _nonAutoreleasedTypingAttributes];
   titleSize = [aString sizeWithAttributes: attributes];
+  if ([self _shouldShortenStringForRect: cellFrame size: titleSize length: [aString length]])
```
ivucica: The shortening of a rendered string using an ellipsis seems to be a new behavior unrelated to XIB5 loading. It's welcome, though. Let's pull this in, but I would usually be happier if it was sent to me as a separate change.

Source/NSCell.m
```diff
@@ -443,43 +445,43 @@ - (void) setFloatValue: (float)aFloat
...
  */
 - (void) setStringValue: (NSString*)aString
 {
   /* We warn about nil for compatibiliy with MacOS X, which refuses
      nil.  */
   if (aString == nil)
     {
-      NSDebugMLLog (@"MacOSXCompatibility", 
+      NSDebugMLLog (@"MacOSXCompatibility",
                     @"Attempt to use nil as string value");
```
ivucica: As there is cleanup here anyway, how about we say `macOS would not allow nil as a string value, proceeding anyway.`?

(Would be ideal if the other OPENSTEPish platform was named fewer times in the code, though.)

ivucica:
I just saw

> Please switch off whitespace changes when reviewing these. 

Please disregard my comments about this. I just didn't see this request from you nor a GitHub option to let me hide this.

Headers/Additions/GNUstepGUI/GSNibLoading.h
```diff
@@ -77,9 +77,15 @@ typedef struct _GSWindowTemplateFlags
   unsigned int autoPositionMask:6;
   unsigned int savePosition:1;
   unsigned int style:2;
-  unsigned int _unused:16; // currently not used, contains Cocoa specific info
+  unsigned int _unused2:3;
+  unsigned int isNotShadowed:1;
+  unsigned int autorecalculatesKeyViewLoop:1;
+  unsigned int _unused:11; // currently not used, contains Cocoa specific info
 #else
```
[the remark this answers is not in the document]
fredkiefer: Actually this is not a GNUstep specific data structure. This is how Cocoa stores these flags in an NIB file. Not much use in arguing about it, we just need to follow Apple here.
What was changes is that Eggplant found out the use of to previously unidentified bits and added that.

Headers/AppKit/NSCell.h
```diff
@@ -188,7 +206,7 @@ enum {
     unsigned allows_undo: 1;
     unsigned line_break_mode: 3; // 6 values
 
-    // total 20 bits.  4 bits extension, 8 bits left.
+    // total 23 bits.  4 bits extension, 5 bits left.
```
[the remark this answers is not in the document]
fredkiefer: Yes, this is correct. There are 4 bits (subclass_bool_X) that are for extensions in subclasses. 23 + these 4 = 27

Headers/AppKit/NSWindow.h
```diff
@@ -96,22 +96,37 @@ enum {
...
-  NSClosableWindowMask = 2,
-  NSMiniaturizableWindowMask = 4,
-  NSResizableWindowMask = 8,
+  NSWindowStyleMaskTitled = NSTitledWindowMask,
+  NSClosableWindowMask = 1 << 1,
+  NSWindowStyleMaskClosable = NSClosableWindowMask,
+  NSMiniaturizableWindowMask = 1 << 2,
+  NSWindowStyleMaskMiniaturizable = NSMiniaturizableWindowMask,
+  NSResizableWindowMask = 1 << 3,
+  NSWindowStyleMaskResizable = NSResizableWindowMask,
```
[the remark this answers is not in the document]
fredkiefer: You are right, that would be more correct. It just looks horrible to either have a separate block with the new values or the #if scattered throughout the definition.

Source/GSCodingFlags.h
```diff
@@ -0,0 +1,385 @@
+/*
+   GSCodingFlags.h
+
+   Define flags used in keyed coding. Extracted from existing files.
```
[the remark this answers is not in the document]
fredkiefer: These flags structures where extracted from various files where Eggplant had added them. A next step would be to actually use these flag structures and the unions in more places where we currently still copy data from a integer into a structure.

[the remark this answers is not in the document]
fredkiefer: As for GSWindowTemplateFlags I wanted to move it here, but non of the other flags would be used in the same file as that. But you are right, it still would be more consistent.

Source/GSCodingFlags.h
```diff
@@ -0,0 +1,385 @@
...
+   Free Software Foundation, 51 Franklin Street, Fifth Floor,
+   Boston, MA 02110-1301, USA.
+*/
+#ifndef _GNUstep_H_GSCodingFlags
+#define _GNUstep_H_GSCodingFlags
+
+#import "GNUstepGUI/GSNibLoading.h"
+
+typedef struct _GSCellFlags {
+#if GS_WORDS_BIGENDIAN == 1
```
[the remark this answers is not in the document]
fredkiefer: OK, I will do so.

Source/GSXibLoader.m
fredkiefer: I am terribly sorry for that. As you can see the indentation is also wrong. I must have missed these lines when adding this code.

Source/NSCell.m
fredkiefer: Yes, this should be "proportional". The idea is to cut off as many characters as the width of the string exceeds the width of the rectangle. In German this is called Dreisatz, which translates into "rule of proportion".

Source/NSCell.m
fredkiefer: Yes, this is unrelated, but comes originally from the Eggplant branch as well. I just rewrote the whole code to understand it better.
Here it would be great if you could help me cherry pick these changes into a separate pull request. I know you can do that :-)

Source/NSCell.m
fredkiefer: The whitespace changes are the result of one careless use of delete-trailing-whitespace. There is an option on Github to ignore such changes.

ivucica: Yes, I didn't look at PR description as I was asked in a chat to take a look at this. Sorry about that!

Either way, it took me quite a bit of time to sift through it, and I'm not sure my comments are useful, but eh.

Headers/AppKit/NSCell.h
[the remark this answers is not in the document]
ivucica: Let's maybe rephrase:

```
// 23 bits for NSCell use, 4 bits for subclass use.
// 5 bits remain unused.
```

Headers/Additions/GNUstepGUI/GSNibLoading.h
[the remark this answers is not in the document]
ivucica: Per below, if we document this, that's fine.

Headers/AppKit/NSWindow.h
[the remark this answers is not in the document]
ivucica: Separate block is better :)

Source/GSNibLoading.m
```diff
@@ -1004,6 +1021,11 @@ - (id) nibInstantiate
 	{
```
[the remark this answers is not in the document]
fredkiefer: Most likely this is just a difference between tab and spaces for indentation.

fredkiefer:
> Many of these styles have been available since the beginning (macOS 10.0) and others in the versions they were originally in in this list. Why move them to 10.12?

Could you please name one example? My impression is that you are confused here.

Source/NSToolbar.m
fredkiefer: I am again sorry, this is another result of delete-training-whitespace. My only explanation why this worked before is that most likely we never had a toolbar encoded.

Source/NSCell.m
fredkiefer: As this is completely unrelated to my changes I will postpone any changes here.

Source/NSButtonCell.m
fredkiefer: I removed these, as I also don't understand what this is supposed to explain.

Source/GSXib5KeyedUnarchiver.m
fredkiefer: I see you point and I tried to remove some of the other classes where similar changes where made but in this case I am unsure what is the bast way. I just didn't want to break old code.

Source/GSXib5KeyedUnarchiver.m
fredkiefer: I could just remove two of the commented out values that where already handle differently, but there are still a few remaining. Just like you I don't like commented out code. Here we would need an example of a XIB file with that element to write code that correctly handles this. Until then this is a reminder that something is missing.

Source/GSXib5KeyedUnarchiver.m
fredkiefer: I choose XmlTagToObjectClassCrossReference as this is more consistent with the other static variable names in this file.

Source/GSXib5KeyedUnarchiver.m
fredkiefer: That is what happens within GSXibLoader.

Source/GSXib5KeyedUnarchiver.m
fredkiefer: I do not understand you point here. What do you want to change and why?

fredkiefer: Hmm, looks like Github displays this and the last few comments with the wrong lines attached to. Sorry, could you please add the lines (plus the code) you are referring to?

fredkiefer: In this case it is line 2794 and most likely this should read "mixed", but I have to try frist with a menu item in that state.

Source/GSXibLoader.m
fredkiefer: That is the way Eggplant writes description methods. I have changed it to the way we normally do this.

Source/GSXib5KeyedUnarchiver.m
fredkiefer: Done

Source/GSXib5KeyedUnarchiver.m
fredkiefer: The indentation is the Emacs default indentation for Objective-C. I agree it looks wiered

ivucica:
> Many of these styles have been available since the beginning (macOS 10.0) and others in the versions they were originally in in this list. Why move them to 10.12?

@gcasa, if you refer to the new `NSWindowStyleMask*`-style names for existing constants, I was the one that suggested marking them as 10.12+ as these names are 10.12+ according to Cocoa docs. I may be wrong as I only documented the online publicly available docs. Did I make a mistake?

Source/GSCodingFlags.h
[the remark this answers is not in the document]
ivucica: I'm probably missing something, but I don't see this is documented in the `.h` file yet?

This:

> Define flags used in Cocoa for keyed coding.

doesn't to me, as a casual reader, suggest "we are really trying to binary compatible with flags used in Cocoa for some values during keyed coding"? :/

Source/GSCodingFlags.h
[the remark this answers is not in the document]
ivucica: > As for GSWindowTemplateFlags I wanted to move it here, but non of the other flags would be used in the same file as that

I don't fully understand this; it still hurts the code readability by being inconsistently thrown into another file, no? :/

Source/GSNibLoading.m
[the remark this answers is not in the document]
ivucica: Indeed, but tabs are not according to the style guide anyway, I think? Anyway, doesn't matter for this PR.

Source/GSXib5KeyedUnarchiver.m
ivucica: As long as the decision is explicitly made, this is fine.

It feels to me like we don't need an additional class here, but eh.

Source/GSXib5KeyedUnarchiver.m
ivucica: Not a fan of it, but local consistency is even more important than project consistency.

Source/GSXib5KeyedUnarchiver.m
ivucica: While 'whatever emacs does' would be a good idea, I don't think Emacs has done a stellar job here or that it's an official policy? :/

Source/GSXib5KeyedUnarchiver.m
ivucica: I don't understand why Github is obnoxiously showing this comment on the wrong line.

It's referring to this code section:

```
- (id) decodeBrowserFlagsForElement: (GSXibElement*)element
{
  NSUInteger    mask                         = 0;
  NSDictionary *attributes                   = [element attributes];
  id            takesTitleFromPreviousColumn = [attributes objectForKey: @"takesTitleFromPreviousColumn"];
  id            allowsEmptySelection         = [attributes objectForKey: @"allowsEmptySelection"];
  id            acceptsArrowKeys             = [attributes objectForKey: @"acceptsArrowKeys"];

  // Set the flags...
  if ([[attributes objectForKey: @"hasHorizontalScroller"] boolValue])
    mask |= 0x10000;
  if ((allowsEmptySelection == nil) || ([allowsEmptySelection boolValue] == NO))
    mask |= 0x20000;
  if ([[attributes objectForKey: @"sendsActionOnArrowKeys"] boolValue])
    mask |= 0x40000;
  if ((acceptsArrowKeys == nil) || [acceptsArrowKeys boolValue])
    mask |= 0x100000;
  if ([[attributes objectForKey: @"separatesColumns"] boolValue])
    mask |= 0x4000000;
  if ((takesTitleFromPreviousColumn == nil) || [takesTitleFromPreviousColumn boolValue])
    mask |= 0x8000000; // Cocoa default is YES if Omitted...
  if ([[attributes objectForKey: @"titled"] boolValue])
    mask |= 0x10000000;
  if ([[attributes objectForKey: @"reusesColumns"] boolValue])
    mask |= 0x20000000;
  if ([[attributes objectForKey: @"allowsBranchSelection"] boolValue])
    mask |= 0x40000000;
  if ([[attributes objectForKey: @"allowsMultipleSelection"] boolValue])
    mask |= 0x80000000;
  if ([[attributes objectForKey: @"prefersAllColumnUserResizing"] boolValue])
    mask |= 0; // FIXME: do we handle this yet???

  return [NSNumber numberWithUnsignedInt: mask];
}
```

image

As you can see, this is different than how the PR UI is showing this:

image

I don't know what Github did here.

ivucica: Similar to above, there's some sort of a Github bug. This refers to lines:

```
#if 0
      // If a font is encoded then change the title to an attributed
      // string and set the font on it...
```

image

ivucica: Unlike lines above, I can't find exactly now, but believe it was this:

image

Source/GSXibLoader.m
ivucica: > I am terribly sorry for that.

It's not like I haven't written or let through code like this, maybe even submitted it to GNUstep :D

Source/NSToolbar.m
ivucica: >  this is another result of delete-training-whitespace. 

Sorry again for not noticing your direct note that I should ignore the whitespace, my mistake!

Source/NSCell.m
ivucica: Yes, my mistake was not seeing that `[mutableString size].width` will return proportional size. Oops!

Source/NSCell.m
ivucica: > Here it would be great if you could help me cherry pick these changes into a separate pull request. I know you can do that :-)

Yeah, it would likely be doable with `git rebase -i $BASE_COMMIT` or similar (which is interactive even if not a GUI), but that would likely break Github pull request review GUI even worse than what it did up there :(

So either we do it as the very last step, or if you plan on squishing all the commits when 'merging' onto `master`, it's likely easiest just to create a new commit that removes these lines and creates new ones.

I can probably sit down tomorrow evening after work and pull this PR into a local working copy, test it, split this off and send it to you as a new PR onto your branch + a new PR for the new changes. 

Not sure if it's fully worth it, which is why I said (emphasis added):

> **Let's pull this in**, but I would **_usually_ be happier** if it was sent to me as a separate change.

Source/GSCodingFlags.h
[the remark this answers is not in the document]
fredkiefer: It does for me :-)
Give me some text to put there and I will do so.

Source/NSCell.m
fredkiefer: OK, so we just leave it in this pull request the way it is.

Source/GSCodingFlags.h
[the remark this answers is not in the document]
fredkiefer: Yes it really does, but I don't know how to resolve this. This structure gets used within NSWindowTemplate in GSNibLoading.h and this makes it hard to move the definition into a private file like GSCodingFlags.h.

Source/GSXib5KeyedUnarchiver.m
fredkiefer: OK, I will try to remove more of these in the future but for now we leave them in.

Source/GSXib5KeyedUnarchiver.m
fredkiefer: Done

Source/GSXib5KeyedUnarchiver.m
fredkiefer: I had to cheat here to get the line better visible. Emacs tried to align the colons within the strings. This really doesn't make sense.

Source/GSXib5KeyedUnarchiver.m
fredkiefer: The smaller classes are just the ones you rightfully want to remove :-)
I think that the chance to remove these are higher if we leave the code here so it is visible.

Source/GSXib5KeyedUnarchiver.m
fredkiefer: Ah now I see. Yes, this is bad code but just making the numbers more pleasant doesn't seem to be the right solution. We should define another structure to go into the GSCodingFlags.h file. But not tonight, I hope to find time for that sometimes next week.

fredkiefer: There are a few more places with "#if 0". I am really not sure what to do with these.

gcasa:
Font decoding is an issue with the latest xibs.... there appears to be a metafont setting named "message" which is being used.... 
2020-01-02 06:19:51.536 SimpleApp[5867:5867] File GSXib5KeyedUnarchiver.m: 1624. In -[GSXib5KeyedUnarchiver decodeFontSizeForElement:] unknown meta font value: message
2020-01-02 06:19:51.542 SimpleApp[5867:5867] File GSXib5KeyedUnarchiver.m: 1624. In -[GSXib5KeyedUnarchiver decodeFontSizeForElement:] unknown meta font value: message
2020-01-02 06:19:51.545 SimpleApp[5867:5867] File GSXib5KeyedUnarchiver.m: 1624. In -[GSXib5KeyedUnarchiver decodeFontSizeForElement:] unknown meta font value: message

This is from a small test program I created just for the xib file.  I transplanted it into a new ProjectCenter project and made the appropriate changes.  When it tries to load the xib it gets these errors.

It should be simple to correct.

gcasa:
Working well so far.  I'm looking forward to this being merged.   I also need to look at changes for loading in Gorm.
Untitled

fredkiefer:
Greg, could you please write your comments on the corresponding source code line? I see now way how to mark your comment as resolved here.

Source/GSXib5KeyedUnarchiver.m
fredkiefer: I removed that line.

Source/GSXib5KeyedUnarchiver.m
fredkiefer: The strange thing is that we actually have an XML parser that works without libxml2, so that part would work, but NSXMLDocument, which only gets used to decide whether we have the correct sort of XML, will only work correctly when that library is present. maybe I should make this check more local? But what would be the default if we don't have libxml2? Old XIB or new XIB?

fredkiefer: I don't know. Please ask the person who added these in GSXibLoader.m the first time. Form there it was just copied over again and again. Looks like Richard did that four years ago right after our Dublin meeting. So all of us could have been involved...

gcasa:
> > Many of these styles have been available since the beginning (macOS 10.0) and others in the versions they were originally in in this list. Why move them to 10.12?
> 
> @gcasa, if you refer to the new `NSWindowStyleMask*`-style names for existing constants, I was the one that suggested marking them as 10.12+ as these names are 10.12+ according to Cocoa docs. I may be wrong as I only documented the online publicly available docs. Did I make a mistake?

No, i withdraw my comment, I think we are good here.    I was confused.

Source/GSCodingFlags.h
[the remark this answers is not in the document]
gcasa: Perhaps something like:

"The following code is architecture specific.  On different platforms (big endian vs little endian) the byte ordering is different.   This is done here so that bitwise operations to determine the state of certain flags works correctly."

Not sure if that is concise enough, but it does describe why it's done.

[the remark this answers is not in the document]
fredkiefer: This would explain what the code does but not why it does that. And the code itself already states that. To me this would violate one of the fundamental rules on code comments: Not to duplicate the code.

Source/GSXib5KeyedUnarchiver.h
fredkiefer: Done

Source/GSXib5KeyedUnarchiver.m
fredkiefer: I moved that code over to GSXibLoader.m and only protect the NSDocument part.

fredkiefer: I removed these spaces in the new place.

fredkiefer:
This pull request is now mostly complete. I am still trying to get my test application running but this could as well be done on the master branch.

gcasa:
> This pull request is now mostly complete. I am still trying to get my test application running but this could as well be done on the master branch.

Do you feel like this is ready for merge yet?

fredkiefer:
> > This pull request is now mostly complete. I am still trying to get my test application running but this could as well be done on the master branch.
> 
> Do you feel like this is ready for merge yet?

Yes, this is what I was trying to say with my last comment. But we should still await Ivan's final review. This pull request is marked as needing rework and this is something I have to respect. I already pinged Ivan in an offline mail and he promised a review over the weekend.

Ivan was the only one who really looked through this huge amount of code so he deserves to have the final say.

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -3005,6 +2962,79 @@ - (id) decodeObjectForXib: (GSXibElement*)element
...
+            [object setImagePosition: NSImageBelow];
+          else
+            [object setImagePosition: NSImageAbove];
+        }
+    }
+  else
+    {
+      [object setImagePosition: NSNoImage];
+    }
+#if 0
```
ivucica: What's this?

```
#if 0
  [object setBordered: NO];
...
...
```

(I see this was in old code too, just wondering about it now)

Source/NSPredicateEditor.m
```diff
@@ -47,4 +47,38 @@ - (void) setRowTemplates: (NSArray *)templates
...
+- (void) encodeWithCoder: (NSCoder*)aCoder
+{
+  [super encodeWithCoder: aCoder];
+
+  if ([aCoder allowsKeyedCoding])
+    {
+      [aCoder encodeObject: _rowTemplates forKey: @"NSRowTemplates"];
+    }
+  else
+    {
```
ivucica: Else empty?

(Maybe put a comment that this is intentional, if that's worth documenting; as in 'nothing needs to be encoded unless using a keyed coder'. Same below.)

Source/NSTableView.m
```diff
@@ -2866,7 +2867,7 @@ - (void) selectRowIndexes: (NSIndexSet *)indexes byExtendingSelection: (BOOL)ext
       if ([indexes lastIndex] >= _numberOfRows)
         {
 	  [NSException raise: NSInvalidArgumentException
-		       format: @"Row index out of table in selectRow"];
+                      format: @"Row index out of table in selectRow"];
```
ivucica: This seems to have changed since last review, but is still not aligned on `:`.

Source/GSXibLoading.m
```diff
@@ -158,3 +194,831 @@ - (id) initWithCoder: (NSCoder *)coder
   return self;
 }
 @end
+
+@implementation FirstResponder
```
ivucica: Class name without prefix? Is this correct? It seems like it was in the old code, too, so it's fine, but I'm just wondering.

ChangeLog
```diff
@@ -1,3 +1,73 @@
...
+
+2020-01-10 Fred Kiefer <FredKiefer@gmx.de>
+
+	* Headers/AppKit/NSKeyValueBinding.h,
+        * Source/externs.m: Add some more bindings.
+	* Source/NSPopUpButton.m,
+	* Source/NSPopUpButtonCell.m,
+	* Source/NSTableView.m: Use the new bindings.
+
+2020-01-01 Fred Kiefer <FredKiefer@gmx.de>
```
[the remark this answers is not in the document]
fredkiefer: This is one of the issues with branches and pull requests. In the ChangeLog we fake a linear history for something that happened in parallel. I see no good solution for this.

Source/GSBindingHelpers.h
```diff
@@ -78,4 +78,10 @@
 @interface GSKeyValueAndBinding : GSKeyValueBinding 
 @end
 
+@interface GSObservableArray : NSArray
+{
+  NSArray *_array;
```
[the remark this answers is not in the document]
fredkiefer: This class gets used for KVO on NSArray, which normally is not possible. In this case we do not want to observe the array but the content elements.
But apart from this having an array within an array is fine. We only need to implement the basic methods like count and objectAtIndex: and everything else will be provided by NSArray.
Mutability does not come into play here at all.

ChangeLog
```diff
@@ -1,3 +1,73 @@
...
+	* Source/GSXib5KeyedUnarchiver.m: Add generic mechanism to decode
+	keys that include the class name.
+	Also add header clip view to subviews of scroller.
+
+2020-01-11 Fred Kiefer <FredKiefer@gmx.de>
+
+        * Headers/AppKit/NSRuleEditor.h
+        * Source/NSRuleEditor.m: Add action and target.
+        * Source/NSPredicateEditor.m: Add encoding and decoding.
+	* Source/GSXib5KeyedUnarchiver.m: Decode row templates.
```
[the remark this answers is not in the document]
fredkiefer: I'll look into this.

Source/GSBindingHelpers.h
```diff
@@ -78,4 +78,10 @@
 @interface GSKeyValueAndBinding : GSKeyValueBinding 
 @end
 
+@interface GSObservableArray : NSArray
```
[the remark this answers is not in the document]
fredkiefer: No, as I wrote above this is about KVO with key paths that include an array. At the moment mostly something like "arrangedObjects.parentTask.name" where "arrangedObjects" is a property of NSArrayController. As this is an NSArray it would normally not be observable. So we need a way around which gets provided by this class.
This is a rather clever concept but marginal for this pull request and also not fully working in GNUstep at the moment.

Source/NSPredicateEditor.m
fredkiefer: This is rather common in GNUstep. For many classes we only have one implementation for (de-)coding, either keyed or the old one. Of course we will both at some point so we leave one if branch empty. I would expect that this is the case for almost one quarter of our classes.

Source/GSXibLoading.m
fredkiefer: This gets explained in the corresponding header, which wasn't changed in the commit so you do not see it.
At the time when I wrote this code I was still trying to be clever, which I hopefully stopped doing :-) There is an object in XIB files that has the name FirstResponder. With this class in place that object gets correctly decoded as nil. Sorry, I promise never to do it again.

Source/GSXib5KeyedUnarchiver.m
fredkiefer: Yes, one of the cases where I keep the commented out code because it may be useful in the future. I really don't like that myself. Not sure what to do.

Source/NSTableView.m
fredkiefer: I will try to do better :-)

fredkiefer: Done

ChangeLog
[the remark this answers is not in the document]
fredkiefer: Done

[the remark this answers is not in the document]
fredkiefer: Done

PR #40, opened 2019-12-17
Scroller border drawing fix in NSScrollView

trunkmaster:
Fixes scroller border positioning and length - do not overlap border if any. Plus, less magic numbers - more maintainable code.

### DIFF
--- a/Source/GSThemeDrawing.m
+++ b/Source/GSThemeDrawing.m
@@ -2597,6 +2597,7 @@ typedef enum {
   NSColor	*color;
   NSString	*name;
   NSBorderType   borderType = [scrollView borderType];
+  CGFloat        borderWidth = 0;
   NSRect         bounds = [view bounds];
   BOOL hasInnerBorder = ![[NSUserDefaults standardUserDefaults]
 			   boolForKey: @"GSScrollViewNoInnerBorder"];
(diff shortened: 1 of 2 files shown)


Source/GSThemeDrawing.m
```diff
@@ -2624,14 +2625,17 @@ - (void) drawScrollViewRect: (NSRect)rect
 	case NSLineBorder:
 	  [color set];
 	  NSFrameRect(bounds);
+          borderWidth = 1;
```
fredkiefer: We need to use [self sizeForBorderType: borderType] here and later use the correct component (width or height) as the theme may override the actual border drawing.

Source/GSThemeDrawing.m
```diff
@@ -2665,8 +2669,8 @@ - (void) drawScrollViewRect: (NSRect)rect
 	    {
               xpos = [vertScroller frame].origin.x + scrollerWidth;
 	    }
-          NSRectFill(NSMakeRect(xpos, [vertScroller frame].origin.y - 1.0, 
-                                1.0, scrollerHeight + 1.0));
+          NSRectFill(NSMakeRect(xpos, [vertScroller frame].origin.y, 
+                                1.0, scrollerHeight - (borderWidth * 2)));
```
fredkiefer: I had to look through the code of [NSScrollerView tile] for quite some time, but now I am sure what you are doing here is correct.
Maybe it could be simplified by reusing [horizScroller frame].size.width but I am unsure. There are so many different cases that get handled in the tile code.

Source/GSThemeDrawing.m
trunkmaster: I've updated the patch.

Source/GSThemeDrawing.m
trunkmaster: There's still one place where scrollers are drawn incorrectly - NSTableView. I guess magic numbers in original code appear for the sake of NSTableView drawing. I'll create a separate PR.

PR #41, opened 2019-12-17
More scroller drawing improvement

trunkmaster:
Current implementation fixes drawing scrollers in NSTableView. 
Moreover I removed all magic numbers from the method. At current state its implementation completely relies on scroller rect calculated in [NSScrollView tile].

### DIFF
--- a/Source/GSThemeDrawing.m
+++ b/Source/GSThemeDrawing.m
@@ -1446,8 +1446,6 @@ static NSImage *spinningImages[MaxCount];
 
   if (tiles == nil)
     { 
-      [[NSColor blackColor] set];
-      NSRectFill(divide);
       rect = [self drawDarkButton: rect withClip: aRect];
       [[NSColor controlShadowColor] set];
       NSRectFill(rect);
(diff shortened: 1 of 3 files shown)


Source/GSThemeDrawing.m
```diff
@@ -2646,51 +2644,47 @@ - (void) drawScrollViewRect: (NSRect)rect
...
-	  CGFloat ypos;
-	  CGFloat scrollerY = [horizScroller frame].origin.y;
-	  CGFloat scrollerLength = bounds.size.width;
+	  CGFloat ypos = [horizScroller frame].origin.y;
 
-	  if ([scrollView hasVerticalScroller])
-	    {
-	      scrollerLength -= [NSScroller scrollerWidth];
-	    }
+          scrollerHeight = [horizScroller frame].size.width;
```
fredkiefer: This value definitely is not a "height". We need to be very careful with variable names. I would also suggest to keep variable definitions as local as possible. Move scrollerHeight into the other "if" branch and have a new local variable here (hScrollerWidth?). Perhaps we should also get the frame of the scroller only once in each branch? Then we would not need a local variable at all.

Source/NSScrollView.m
```diff
@@ -1213,10 +1213,15 @@ - (void) tile
 
       /** Vertically expand the scroller by 1pt on each end */
       if (overlapBorders)
-	{
+        {
 	  vertScrollerRect.origin.y -= 1;
 	  vertScrollerRect.size.height += 2;
 	}
+      else if (_hasHeaderView || _hasCornerView)
```
fredkiefer: What if we have overlapBorders and a header view? Not sure what should happen, but for now this is OK.

Source/GSThemeDrawing.m
trunkmaster: As opposite to [NSScroller scrollerWidth] it would be logical to name it `scrollerHeight`. And it is definitely not "width" because it's confusing to method name. Width of horizontal scroller is visual height, while width of vertical scroller is visual width.
Anyway, I've made changes.

Source/NSScrollView.m
trunkmaster: I don't know either.

PR #42, opened 2019-12-18
NSBrowser column positionning and size calculation fixes

trunkmaster:
There is difference in column positionning with and without horizontal scroller in separated columns mode.


ChangeLog
```diff
@@ -1,3 +1,10 @@
+2019-12-18  Sergii Stoian  <stoyan255@gmail.com>
+
+	* Source/NSBrowser.m (frameOfColumn:): shift up column only for browser
+	with separated sparated columns layout.
```
fredkiefer: "sparated" seems to be a typo.

PR #43, opened 2019-12-19
NSPopUpButton's popup menu in pulldown mode displaying fix

trunkmaster:
I've changed the way how poup menu is displayed when popup button operates in pulldown mode. Before this change popup menu doesn't display first item and was placed to the bottom edge of popup button. Although popup menu can't draw dark gray border around poup button.

With this change pulldown menu draws exatrly the same way as in normal mode except first item is not highlighted. So now popup menu looks visually consistent no matter what mode it opeartes in. 
Plus, NSMenuView contains less popup button specific code.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,18 @@
+2019-12-19  Sergii Stoian  <stoyan255@gmail.com>
+
+	* Source/NSMenuView.m (setHighlightedItemIndex:): do not highlight first
+	item for pulldown popup button's menu.
+	(sizeToFit): removed commented useless code; fixed formatting.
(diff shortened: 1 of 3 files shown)


fredkiefer:
Some of the changes of this pull request are obviously correct. What I doubt is the overall change. Why would we want to display the title item (that is the first item in a pull down) in the menu?
This is the main difference between pull down and pop up and it is well documented. I just tired this on Cocoa and the difference is still there.
Of course changing this would simplify the code here a lot and I would love to see that, but this changes the whole appearance of pull down items and this needs to be decided on a completely different level.

PR #47, opened 2020-01-20
NSTouchBar branch

gcasa:
After much consideration... I really think we might be able to do the touchbar.  I am opening this request early to discuss that.

I believe what can be done is that, for applications which use the touchbar, we could have a view on the bottom or a floating view which represents the set of touchbar items.   This would allow us to provide that.  The reason why I haven't implemented anything with respect to the touchbar up until now is because Apple, specifically, mentions in it's guidelines to NOT design your app to be dependent on the touchbar as not all macs have them.

It's a lot of extra work, it will be "cute" but not necessarily needed.  Any thoughts?   For now I am going to provide the interface, but not a full implementation.  If people feel strongly that we should have a touchbar implementation I will implement it.


fredkiefer:
I really see no point in implementing these classes, as they never will improve the look and feel of an application. But if you have a specific project that uses these classes it is fine to implement them.

gcasa:
> I really see no point in implementing these classes, as they never will improve the look and feel of an application. But if you have a specific project that uses these classes it is fine to implement them.

I am only implementing some of these classes now because I want things that use them in source to compile.  At this point I will leave them as non-functional skeletons.  In the future I may put a warning in them so that when they are used to build, it tells the developer that the touchbar classes are non-functional.

PR #48, opened 2020-01-21
Rtf file encoding fix

sensetalkdoug:
This fixes the problem with reading the LanguageBrowserContents.rtf file.

### DIFF
--- a/TextConverters/RTF/RTFConsumer.m
+++ b/TextConverters/RTF/RTFConsumer.m
@@ -571,7 +571,7 @@ static BOOL classInheritsFromNSMutableAttributedString (Class c)
   CREATE_AUTORELEASE_POOL(pool);
   RTFscannerCtxt scanner;
   StringContext stringCtxt;
-  char buffer[5];
+  char buffer[15];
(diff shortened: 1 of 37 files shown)


TextConverters/RTF/RTFConsumer.m
```diff
@@ -595,6 +595,14 @@ - (NSAttributedString*) parseRTF: (NSData *)rtfData
       // FIXME: Code page 850 kCFStringEncodingDOSLatin1
       encoding = NSISOLatin1StringEncoding;
     }
+  else if (strncmp(buffer, "ansi", 4) == 0)
```
fredkiefer: Please have a look at the change to RTF parsing that I just committed on GNUstep master.

sensetalkdoug:
Thanks, Fred! Your fix looks much more comprehensive and correct (not that I could totally follow how the bison grammar works). We intend to work at merging GNUstep master into our branch soon — unfortunately, it’s been a while since we did that so we have to be careful about what might break for us when we get there.

Doug

PR #50, opened 2020-01-26
Update window position on screen parameters change

trunkmaster:
This is a complementary to GNUstep Back [PR #18](https://github.com/gnustep/libs-back/pull/18). 

Some explanation of logic in this and Back's changes:
- backend releases NSScreen list. It destroys list, but NSScreen instance still exists because every window of application stiil retains it;
- backend posts NSApplicationDidChangeScreenParametersNotification on XRandR event;
- NSWindow handles notification with -applicationDidChangeScreenParameters:
   - saves old screen frame;
   - releases old NSScreen instance and get new with actual parameters (frame in this case);
   - calculates diffrerences between screens' dimensions and position;
   - sets frame to calculated one and saves frame if autosaveFrame was set.

As a result, NSScreen list (array) will be deallocated after last window proceed with notification handling.


Source/NSWindow.m
```diff
@@ -2701,6 +2705,50 @@ - (NSScreen *) screen
...
+  NSRect newFrame;
+
+  // NSLog(@"[NSWindow - %@] Old/New Screen RC: %lu/%lu",
+  //       [self className], [_screen retainCount],
+  //       [[[NSScreen screens] objectAtIndex: 0] retainCount]);
+  
+  // We need to get new screen from renewed screen list because
+  // [NSScreen mainScreen] returns NSScreen object of key window and that object
+  // will never be released.
+  for (NSScreen *scr in [NSScreen screens])
```
fredkiefer: Not sure whether this is supported by all compilers we want to work with.

Source/NSWindow.m
```diff
@@ -2701,6 +2705,50 @@ - (NSScreen *) screen
...
 }
 
+- (void) applicationDidChangeScreenParameters: (NSNotification *)aNotif
+{
+  NSRect oldScreenFrame = [_screen frame];
+  int    screenNumber = [_screen screenNumber];
+  NSRect newScreenFrame;
+  NSRect newFrame;
+
+  // NSLog(@"[NSWindow - %@] Old/New Screen RC: %lu/%lu",
```
fredkiefer: Could you please remove these commented out lines or turn them into NSDebugLog statements?

Source/NSWindow.m
```diff
@@ -2701,6 +2705,50 @@ - (NSScreen *) screen
...
+
+  newScreenFrame = [_screen frame];
+
+  // NSLog(@"[NSWindow - %@] Old/New Screen (%lu): %.0f/%.0f",
+  //       [self className], [_screen retainCount],
+  //       oldScreenFrame.size.width, newScreenFrame.size.width);
+
+  newFrame = _frame;
+  // Screen Y origin change.
+  newFrame.origin.y += newScreenFrame.origin.y - oldScreenFrame.origin.y;
```
fredkiefer: These calculation seem correct for the case where the origin of a screen changes, what about resolution changes where we should also adjust the size of the window? We have similar code in the restore of a window. Please have a look at that.

trunkmaster: We don't need to change size of a window if resolution changes. It's not a DPI - it's a pixel dimensions of a screen. So we need to update origin of a window to actual position of a window.

Source/NSWindow.m
trunkmaster: Changed to NSEnumerator.

Source/NSWindow.m
trunkmaster: Done.

trunkmaster:
> 
> 
> Looks OK to me. Maybe it would be worthwhile to consolidate the Change log as some of the entries are now obsolete.

Done, thanks.

PR #51, opened 2020-01-30
NSFontCollection branch

gcasa:
The NSFontCollection classes in Cocoa will require the ability to physically move or copy fonts on the filesystem.  The GNUstep backend doesn't have that ability at the moment, so this is a placeholder implementation at the moment.  I implemented it as I did so that it can return some data which is relevant.


Source/NSFontCollection.m
```diff
@@ -0,0 +1,341 @@
...
+
+static NSMutableDictionary *__sharedFontCollections;
+static NSMutableDictionary *__sharedFontCollectionsVisibility;
+static NSMutableSet *__sharedFontCollectionsHidden;
+
++ (void) initialize
+{
+  if (self == [NSFontCollection class])
+    {
+      __sharedFontCollections = [[NSMutableDictionary alloc] initWithCapacity: 100];
```
ivucica: Double underscore?

gcasa: https://developer.apple.com/documentation/appkit/nsfontcollection/1497512-showfontcollection?language=objc

There's the proof it manipulates the filesystem.  It returns an NSError from NSFileManager.   Creating GSFontCollections.plist might not be a bad idea.

Double underscore because it's global.

gcasa: Actually interesting thing...

heron@glados FontCollections % pwd                                                                                                                                                                                  [127]
/Users/heron/Library/FontCollections
heron@glados FontCollections %                                                                                                                                                                                        [0]
heron@glados FontCollections % ls                                                                                                                                                                                     [0]
Fixed Width.collection		Modern.collection		Traditional.collection		com.apple.Recents.collection
Fun.collection			PDF.collection			Web.collection

Apparently they are stored as a set of attributes under the Library which means they should have an analogous home in the GNUstep/Library directory.

Source/NSFontCollection.m
```diff
@@ -0,0 +1,341 @@
...
+@end
+
+@implementation NSCTFontCollection
+@end
+
+@implementation NSFontCollection 
+
+static NSMutableDictionary *__sharedFontCollections;
+static NSMutableDictionary *__sharedFontCollectionsVisibility;
+static NSMutableSet *__sharedFontCollectionsHidden;
```
fredkiefer: This variable should be deductible from __sharedFontCollectionsVisibility

Source/NSFontCollection.m
fredkiefer: No, this is user not system library.
Looking at the files in my FontCollections directory these are simple archives that store a FontCollection each.

Source/NSFontCollection.m
```diff
@@ -0,0 +1,341 @@
...
+{
+  RELEASE(_fonts);
+  RELEASE(_queryDescriptors);
+  RELEASE(_exclusionDescriptors);
+  RELEASE(_queryAttributes);
+  [super dealloc];
+}
+
+// This method will get the actual list of fonts
+- (void) _runQueryWithDescriptors: (NSArray *)queryDescriptors
```
fredkiefer: Looks like we do not need this method. But we need something similar that turns logical font descriptors into physical ones. (Maybe based on NSFontDescriptor matchingFontDescriptorsWithMandatoryKeys:?) This should also respect the _exclusionDescriptors of course.

Source/NSFontCollection.m
```diff
@@ -0,0 +1,341 @@
...
+
++ (NSFontCollection *) fontCollectionWithAllAvailableDescriptors
+{
+  return [self fontCollectionWithDescriptors:
+                 [[GSFontEnumerator sharedEnumerator] availableFontDescriptors]];
+}
+
++ (NSFontCollection *) fontCollectionWithLocale: (NSLocale *)locale
+{
+  return [self fontCollectionWithAllAvailableDescriptors];
```
fredkiefer: Wrong, but I really don't understand what should happen here. Did you try this method on Cocoa to see what it does?

Source/NSFontCollection.m
```diff
@@ -0,0 +1,341 @@
...
+                                        forKey: name];
+  return YES;
+}
+
++ (BOOL) hideFontCollectionWithName: (NSFontCollectionName)name
+                         visibility: (NSFontCollectionVisibility)visibility
+                              error: (NSError **)error
+{
+  
+  [__sharedFontCollectionsHidden addObject: name];
```
fredkiefer: This clearly should work on __sharedFontCollectionsVisibility

Source/NSFontCollection.m
```diff
@@ -0,0 +1,341 @@
...
+}
+
++ (BOOL) renameFontCollectionWithName: (NSFontCollectionName)aname
+                           visibility: (NSFontCollectionVisibility)visibility
+                               toName: (NSFontCollectionName)name
+                                error: (NSError **)error
+{
+  NSFontCollection *fc = [__sharedFontCollections objectForKey: aname];
+  [__sharedFontCollections setObject: fc forKey: name];
+  return YES;
```
fredkiefer: This method does not set the visibility of the font collection. Nor does it remove the old name from any of the collections.

Source/NSFontCollection.m
```diff
@@ -0,0 +1,341 @@
...
+
++ (NSFontCollection *) fontCollectionWithName: (NSFontCollectionName)name
+{
+  return [__sharedFontCollections objectForKey: name];
+}
+
++ (NSFontCollection *) fontCollectionWithName: (NSFontCollectionName)name
+                                   visibility: (NSFontCollectionVisibility)visibility
+{
+  return [__sharedFontCollections objectForKey: name];
```
fredkiefer: This looks wrong. It seems we don't quite understand the visibility concept here. This seems to be independent of show/hide.

Source/NSFontCollection.m
```diff
@@ -0,0 +1,341 @@
...
+}
+
+// Descriptors
+- (NSArray *) queryDescriptors  // copy
```
fredkiefer: Please remove trailing comment

Source/NSFontCollection.m
```diff
@@ -0,0 +1,341 @@
...
+
+- (NSArray *) matchingDescriptors
+{
+  return [self matchingDescriptorsWithOptions: nil];
+}
+
+- (NSArray *) matchingDescriptorsWithOptions: (NSDictionary *)options
+{
+  GSFontEnumerator *fen = [GSFontEnumerator sharedEnumerator];
+  return [fen matchingFontDescriptorsFor: options];
```
fredkiefer: Here we need to add _queryDescriptors and _exclusionDescriptors to get the actual font descriptors. The options provided here cannot the handled by GSFontEnumerator at the moment. Maybe we should handle them here? As matchingDescriptorsForFamily:options: seems to be the most general method, all others should call that and we need a better implementation there.

Source/NSFontCollection.m
```diff
@@ -0,0 +1,341 @@
...
+}
+
+- (instancetype) initWithCoder: (NSCoder *)coder
+{
+  self = [super init];
+  if (self != nil)
+    {
+      if ([coder allowsKeyedCoding])
+        {
+        }
```
fredkiefer: Code missing

Source/NSFontCollection.m
```diff
@@ -0,0 +1,341 @@
...
+    }
+  return self;
+}
+
+- (void) encodeWithCoder: (NSCoder *)coder
+{
+}
```
fredkiefer: Completely missing

Source/NSFontCollection.m
```diff
@@ -0,0 +1,341 @@
...
+  
+@end
+ 
+
+@implementation NSMutableFontCollection 
+
++ (NSMutableFontCollection *) fontCollectionWithDescriptors: (NSArray *)queryDescriptors
+{
+  NSMutableFontCollection *fc = [[NSMutableFontCollection alloc] init];
+  ASSIGNCOPY(fc->_queryDescriptors, queryDescriptors);
```
fredkiefer: Shouldn't you assign a mutable array here?

Source/NSFontCollection.m
```diff
@@ -0,0 +1,341 @@
...
+}
+
++ (NSMutableFontCollection *) fontCollectionWithLocale: (NSLocale *)locale
+{
+  return [self fontCollectionWithAllAvailableDescriptors];
+}
+
++ (NSMutableFontCollection *) fontCollectionWithName: (NSFontCollectionName)name
+{
+  return [__sharedFontCollections objectForKey: name];
```
fredkiefer: This will not suddenly return a mutable instance. You should call the super implementation and make a mutable copy. The same is true for more of these methods.

Source/NSFontCollection.m
```diff
@@ -0,0 +1,341 @@
...
+}
+
+- (void) setExclusionDescriptors: (NSArray *)exclusionDescriptors
+{
+  ASSIGN(_exclusionDescriptors, [exclusionDescriptors mutableCopy]);
+}
+
+- (void)addQueryForDescriptors: (NSArray *)descriptors
+{
+  [_queryDescriptors addObjectsFromArray: descriptors];
```
fredkiefer: The documentation is a bit unclear here, but I would expect that this method also removes the descriptors from _exclusionDescriptors. While the next method does the opposite. This should be fairly easy to test on Cocoa.

Source/NSFontCollection.m
```diff
@@ -0,0 +1,341 @@
...
+{
+  return [self fontCollectionWithAllAvailableDescriptors];
+}
+
++ (BOOL) showFontCollection: (NSFontCollection *)collection
+                   withName: (NSFontCollectionName)name
+                 visibility: (NSFontCollectionVisibility)visibility
+                      error: (NSError **)error
+{
+  NSNumber *v = [NSNumber numberWithInt: visibility];
```
fredkiefer: I think all these show/hide methods also need to send the notifications. Otherwise those don't make any sense.

Source/NSFontCollection.m
```diff
@@ -0,0 +1,341 @@
...
+{
+  NSFontCollection *fc = [[NSFontCollection alloc] init];
+  ASSIGNCOPY(fc->_queryDescriptors, queryDescriptors);
+  return fc;
+}
+
++ (NSFontCollection *) fontCollectionWithAllAvailableDescriptors
+{
+  return [self fontCollectionWithDescriptors:
+                 [[GSFontEnumerator sharedEnumerator] availableFontDescriptors]];
```
fredkiefer: I am rather sure that we should not use the physical font descriptors here, instead a logical font descriptor that results in all fonts should be created.

fredkiefer:
What is the status of this branch? Are you still working on it or do you expect me to take over?

Source/NSFontCollection.m
gcasa: Correct, it is the user library.  That's why I mentioned GNUstep/Library.   I simply didn't want to type it each and every time.   If I had meant the system library I would have said /usr/GNUstep/System/Library. :)

Source/NSFontCollection.m
```diff
@@ -0,0 +1,533 @@
...
+{
+  [_fontCollectionLock lock];
+  if (_availableFontCollections != nil)
+    {
+      // Nothing to do ... already loaded
+      [_fontCollectionLock unlock];
+    }
+  else
+    {
+      NSString			*dir;
```
fredkiefer: Indentation is wrong here.

Source/NSFontCollection.m
```diff
@@ -0,0 +1,533 @@
...
+      NSString			*file;
+      NSEnumerator		*e;
+      NSFileManager		*fm = [NSFileManager defaultManager];
+      NSDirectoryEnumerator	*de;
+      NSFontCollection		*newCollection;
+
+      if (_availableFontCollections == nil)
```
fredkiefer: You just checked that in line 59.

Source/NSFontCollection.m
```diff
@@ -0,0 +1,533 @@
...
+	  // Create the global array of font collections...
+	  _availableFontCollections = [[NSMutableArray alloc] init];
+	}
+      else
+	{
+	  [_availableFontCollections removeAllObjects];
+	}
+      
+      /*
+       * Load color lists found in standard paths into the array
```
fredkiefer: "color lists"?

Source/NSFontCollection.m
```diff
@@ -0,0 +1,533 @@
...
+
+	  de = [fm enumeratorAtPath: dir];
+	  while ((file = [de nextObject])) 
+	    {
+	      if ([[file pathExtension] isEqualToString: @"collection"])
+		{
+		  NSString	*name;
+		  name = [file stringByDeletingPathExtension];
+		  newCollection = [NSKeyedUnarchiver unarchiveObjectWithFile: [dir stringByAppendingPathComponent: file]];
+                  if (newCollection != nil)
```
fredkiefer: Indentation is wrong

Source/NSFontCollection.m
```diff
@@ -0,0 +1,533 @@
...
+                    {
+                      [_availableFontCollections addObject: newCollection];
+                      RELEASE(newCollection);
```
fredkiefer: I am a bit unsure here but I would expect that objects returned from the keyed archiver are autoreleased. If this is true you are creating a zombie here.

Source/NSFontCollection.m
```diff
@@ -0,0 +1,533 @@
...
+                    }
+                }
+	    }
+	}  
+      /*
```
fredkiefer: Just remove this commented out code.

Source/NSFontCollection.m
```diff
@@ -0,0 +1,533 @@
...
+    }
+
+  if (isDir)
+    {
+      ASSIGN (_fullFileName, [[path stringByAppendingPathComponent: _name] 
+        stringByAppendingPathExtension: @"collection"]);
+    }
+  else // it is a file
+    {
+      if ([[path pathExtension] isEqual: @"collection"] == YES)
```
fredkiefer: The processing here and in the next if block are only needed when path has been explicitly specified, maybe it would be better to move that code up into that branch?

Source/NSFontCollection.m
```diff
@@ -0,0 +1,533 @@
...
+	  path_is_standard = NO;
+	}
+    }
+
+  /*
+   * If path is standard and it does not exist, try to create it.
+   * System standard paths should always be assumed to exist; 
+   * this will normally then only try to create user paths.
+   */
+  if (path_is_standard && ([fm fileExistsAtPath: path] == NO))
```
fredkiefer: Shouldn't we also create a non existing non standard path? Otherwise writing the archive will fail.

Source/NSFontCollection.m
```diff
@@ -0,0 +1,533 @@
...
+    }
+
+  success = [NSKeyedArchiver archiveRootObject: self 
+                                        toFile: _fullFileName];
+
+  if (success && path_is_standard)
+    {
+      [_fontCollectionLock lock];
+      if ([_availableFontCollections containsObject: self] == NO)
+	[_availableFontCollections addObject: self];
```
fredkiefer: Please add curly braces here and adjust the indentation.

Source/NSFontCollection.m
```diff
@@ -0,0 +1,533 @@
...
+
+- (void) _removeFile
+{
+  if (_fullFileName) //  && _is_editable)
+    {
+      // Remove the file
+      [[NSFileManager defaultManager] removeFileAtPath: _fullFileName
+					       handler: nil];
+      
+      // Remove the color list from the global list of colors
```
fredkiefer: Copying code is OK, but please adjust the comments to fit this use case.

Source/NSFontCollection.m
```diff
@@ -0,0 +1,533 @@
...
+    }
+}
+
+// Initializers...
+- (instancetype) init
+{
+  self = [super init];
+  if (self != nil)
+    {
+      //      _fonts = [[NSMutableArray alloc] initWithCapacity: 50];
```
fredkiefer: Please remove all _fonts related code.

Source/NSFontCollection.m
```diff
@@ -0,0 +1,533 @@
...
++ (NSMutableFontCollection *) fontCollectionWithDescriptors: (NSArray *)queryDescriptors
+{
+  NSMutableFontCollection *fc = [[NSMutableFontCollection alloc] init];
+  ASSIGNCOPY(fc->_queryDescriptors, queryDescriptors);
+  return fc;
+}
+
++ (NSMutableFontCollection *) fontCollectionWithAllAvailableDescriptors
+{
+  return [self fontCollectionWithDescriptors:
```
fredkiefer: Somehow you managed to remove my comment here. It said that you should call the super implementation and call mutableCopy on the result.

Source/NSFontCollection.m
```diff
@@ -1,4 +1,4 @@
-/* Implementation of class NSFontCollection
+man/* Implementation of class NSFontCollection
```
fredkiefer: Here are a few leftover characters.

Source/NSFontCollection.m
```diff
@@ -127,14 +129,15 @@ + (void) _loadAvailableFontCollections
...
-- (BOOL) _writeToFile: (NSString *)path
+- (BOOL) _writeToFile
 {
   NSFileManager *fm = [NSFileManager defaultManager];
   NSString      *tmpPath;
   BOOL          isDir;
   BOOL          success;
   BOOL          path_is_standard = YES;
-
+  NSString     *path = nil;
```
fredkiefer: Now path will always be nil further down and you may remove most of the non standard logic there.

fredkiefer:
@gcasa would it be OK if I start working on this branch? It would be great to get these changes included in our upcoming release.

gcasa:
I can finish it.

fredkiefer:
I know, but this does not answer my question.

gcasa:
> I know, but this does not answer my question.

Sure.... I don't see why help would be a problem.   One question, do you think we should make the encoding compatible since it is unlikely that font collection data will be shared between macOS and GNUstep systems?

fredkiefer:
> Sure.... I don't see why help would be a problem. One question, do you think we should make the encoding compatible since it is unlikely that font collection data will be shared between macOS and GNUstep systems?

I think we should keep these compatible. I really don't see a reason that should stop somebody copying over a specific font collection from MacOS. The point here is that most font collection won't include any specific fonts just criteria for fonts.

gcasa:
archive weirdness...

One thing I have noticed is that the keyed archive for most normal archives has a "root" entry.   The one for these doesn't.  It has "NSFontCollectionDictionary".   I'm wondering why this is different and if we can make this compatible at all.   I still want this class in the library, I am just not sure it can be made compatible.

gcasa:
If you can take a look at the keyed coding methods I would appreciate it.

fredkiefer:
I have a closer look, maybe even contribute to that class. But the basic idea of having just a dictionary in the class seems correct. This is what I get when I save a keyed archived version of such an object on Cocoa. There still is a root object of class NSCTFontCollection in the archive.

Source/NSFontCollection.m
```diff
@@ -0,0 +1,645 @@
...
+}
+
+- (instancetype) initWithCoder: (NSCoder *)coder
+{
+  self = [super init];
+  if (self != nil)
+    {
+      if ([coder allowsKeyedCoding])
+        {
+          [coder encodeObject: _fontCollectionDictionary
```
fredkiefer: You seem to have mixed the code between initWithCoder and encodeWithCoder. Just switch the implementation

Source/NSFontCollection.m
```diff
@@ -0,0 +1,622 @@
...
+  if (fc == nil)
+    {
+      BOOL rv = NO; 
+
+      fc = [[NSFontCollection alloc] init];
+      [fc _setName: name];
+      rv = [fc _writeToFile];
+      if (rv == YES)
+        {
+          [NSFontCollection _loadAvailableFontCollections];
```
fredkiefer: Again this should not be needed.

Source/NSFontCollection.m
```diff
@@ -0,0 +1,622 @@
...
+  return [self fontCollectionWithDescriptors: [NSArray arrayWithObject: fd]];
+}
+
++ (BOOL) showFontCollection: (NSFontCollection *)collection
+                   withName: (NSFontCollectionName)name
+                 visibility: (NSFontCollectionVisibility)visibility
+                      error: (NSError **)error
+{
+  BOOL rv = [collection _writeToFile];
+  [NSFontCollection _loadAvailableFontCollections];
```
fredkiefer: Same here

Source/NSFontCollection.m
```diff
@@ -0,0 +1,622 @@
...
+{
+  NSFontCollection *collection = [_availableFontCollections objectForKey: aname];
+  BOOL rv = [collection _removeFile];
+
+  if (rv == YES)
+    {
+      [collection _setName: name];
+      [collection _writeToFile];
+    }
+  [NSFontCollection _loadAvailableFontCollections];
```
fredkiefer: Another call that has no effect.

Source/NSFontCollection.m
```diff
@@ -361,27 +379,34 @@ - (void) dealloc
...
+        }
     }
-  
   return fc;
 }
 
 + (NSFontCollection *) fontCollectionWithLocale: (NSLocale *)locale
 {
-  return [self fontCollectionWithAllAvailableDescriptors];
+  NSDictionary *fa = [NSDictionary dictionaryWithObject: [locale languageCode]
```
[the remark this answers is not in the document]
fredkiefer: Just update base, I committed that change before changing it here. I could have used [locale objectForKey: XXX] instead but this looks clumsy and these methods are required for base anyway. There are a lot more of them if you want to contribute :-)

Source/NSFontCollection.m
```diff
@@ -0,0 +1,627 @@
...
+
++ (NSFontCollection *) _readFileAtPath: (NSString *)path
+{
+  NSFontCollection *fc = [[NSFontCollection alloc] init];
+  NSData *d = [NSData dataWithContentsOfFile: path];
+  NSKeyedUnarchiver *u = [[NSKeyedUnarchiver alloc] initForReadingWithData: d];
+
+  [fc _setFontCollectionDictionary: [u decodeObjectForKey: @"NSFontCollectionDictionary"]];
+  RELEASE(u);
+  RELEASE(d);
```
fredkiefer: This release looks wrong, the NSData object should already be autoreleased.

Source/NSFontCollection.m
```diff
@@ -0,0 +1,637 @@
...
+}
+
+- (NSArray *) exclusionDescriptors
+{
+  return [_fontCollectionDictionary objectForKey: @"NSFontExclusionDescriptorAttributes"];
+}
+
+- (void) setExclusionDescriptors: (NSArray *)exclusionDescriptors
+{
+  NSMutableArray *qd = [[self queryDescriptors] mutableCopy];
```
fredkiefer: I just run a few tests on Cocoa and it looks like this method just sets the exclusion descriptors without looking at the query descriptors. You could drop the first three lines here.

gcasa:
I think this is ready to merge, please let me know if you approve.

fredkiefer:
I am currently away from home and have no access to github that's why I reply per mail.

The main functionality of this class is still missing. I put a FIXME in the method where this would belong. In the current state the class is pretty useless. I hope to find time to implement this over the next weekend. I had more time at hand a few weeks ago when I offered my help.
And the last time I looked there also still was unused code in another method. There should be a comment on that.

Cheers,
Fred


Am 15.03.2020 um 21:19 schrieb Gregory Casamento <notifications@github.com>:

> I think this is ready to merge, please let me know if you approve.

PR #52, opened 2020-01-31
Keep menu inside visible screen frame.

trunkmaster:
Limit menu movement to screen frame for top/left/right edges. 
Menu can be moved to the bottom of the screen while title is completely visible.

Prerequisite: backend should be compiled with XRandR support for per-monitor limits.

Side effects:
- user can easily place main application menu to default top-left position of the screen;
- no usability confusion after implementation of "main menu follows key window" on multi-monitor setups - menu won't cross monitors boundaries.

### DIFF
--- a/Source/GSTitleView.m
+++ b/Source/GSTitleView.m
@@ -249,7 +249,7 @@
   // Define move constrains for menu
   if (_ownedByMenu)
     {
-      NSRect screenFrame = [[NSScreen mainScreen] frame];
+      NSRect screenFrame = [[_window screen] frame];
       leftLimit = screenFrame.origin.x;
       topLimit = NSMaxY(screenFrame) - [_window frame].size.height;
       rightLimit = NSMaxX(screenFrame) - [_window frame].size.width;
--- a/Source/GSTitleView.m
+++ b/Source/GSTitleView.m
@@ -2,7 +2,7 @@
 
    Copyright (C) 2003 Free Software Foundation, Inc.
 
-   Author: Serg Stoyan <stoyan@on.com.ua>
+   Author: Sergii Stoian <stoyan255@gmail.com>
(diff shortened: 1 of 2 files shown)


Source/GSTitleView.m
```diff
@@ -238,9 +239,24 @@ - (void) mouseDown: (NSEvent*)theEvent
...
+  CGFloat  topLimit;
+  CGFloat  rightLimit;
+  CGFloat  bottomLimit;
 
   NSDebugLLog (@"NSMenu", @"Mouse down in title!");
 
+  // Define move constrains for menu
+  if (_ownedByMenu)
+    {
+      NSRect screenFrame = [[NSScreen mainScreen] frame];
```
fredkiefer: As I already wrote in a private mail, we should be using [_window screen] here. There is even a method on NSWindow to handle this code constrainFrameRect: toScreen:, why not use that?

trunkmaster: > As I already wrote in a private mail, we should be using [_window screen] here. 

Fixed.

> There is even a method on NSWindow to handle this code constrainFrameRect: toScreen:, why not use that?

Apple documentation states: "If the window is resizable and the window’s height is greater than the screen height, the rectangle’s height is adjusted to fit within the screen as well. The rectangle’s width and horizontal location are unaffected." This is not I want to achieve. I want constrains at left and right also. Plus, I want user to be able move menu off the screen while menu title should remain visible when moving menu to the bottom.

fredkiefer: As always we should protect the code against _window and [_window screen] being nil. That would result in a crash on some machines.
And maybe it would be worthwhile to save [_window frame] in a local variable.

trunkmaster: Thanks, done.

PR #53, opened 2020-02-04
NSSpeechRecognizer branch

gcasa:
Changes to implement NSSpeechRecognizer.


Source/NSSpeechRecognizer.m
```diff
@@ -23,8 +23,163 @@
...
+                }
+              [NSThread sleepForTimeInterval: 0.1];
+            }
+          
+          // Set a flag so we don't bother waiting for the speech recognition server to
+          // launch the next time if it didn't work this time.
+          _serverLaunchTested = YES;
+        }
+      // If there is no server, this will return nil
+      return [_speechRecognitionServer newRecognizer];
```
fredkiefer: I must admit I don't understand the logic here. If this method gets called on the NSSpeechRecognizer class we return an NSProxy to a server object. Doesn't this make all our local methods obsolete?

Source/NSSpeechSynthesizer.m
```diff
@@ -102,58 +102,60 @@ @interface NSObject (GSSpeechServer)
...
-	                                                            host: nil] retain];
-	if (nil == server)
-	{
-		NSWorkspace *ws = [NSWorkspace sharedWorkspace];
-		[ws launchApplication: @"GSSpeechServer"
-		             showIcon: NO
-		           autolaunch: NO];
-	}
+  NSSpeechSynthesizerClass = [NSSpeechSynthesizer class];
+  server = [[NSConnection rootProxyForConnectionWithRegisteredName: @"GSSpeechServer"
```
fredkiefer: Why not put @"GSSpeechServer" in a constant?

Source/NSSpeechSynthesizer.m
```diff
@@ -102,58 +102,60 @@ @interface NSObject (GSSpeechServer)
...
-			}
-			// Set a flag so we don't bother waiting for the speech server to
-			// launch the next time if it didn't work this time.
-			serverLaunchTested = YES;
-		}
-		// If there is no server, this will return nil
-		return [server newSynthesizer];
-	}
-	return [super allocWithZone: aZone];
+  if (self == NSSpeechSynthesizerClass)
```
fredkiefer: For this class most of the comments for NSSpeechRecognizer apply as well.

Tools/speech_recognizer/GNUmakefile
```diff
@@ -0,0 +1,36 @@
...
+	GSSpeechRecognizer.m \
+	main.m
+
+
+ifeq ($(GSSPEECHRECOGNITIONENGINE), Pocketsphinx)
+POCKETSPHINX_BASE_LIBS = -lpocketsphinx -lsphinxbase -lsphinxad
+
+# Add includes and link dirs.... need to replace triplet.
+GSSpeechRecognitionServer_OBJC_FILES += PocketsphinxSpeechRecognitionEngine.m
+GSSpeechRecognitionServer_INCLUDE_DIRS += -I/usr/local/include/pocketsphinx \
```
fredkiefer: Isn't there some sort of autodetection for Pocketsphinx? This way we have quite a lot of hard coded paths that won't work on most machines.

Tools/speech_recognizer/GSSpeechRecognitionEngine.h
```diff
@@ -0,0 +1,25 @@
+#import <Foundation/Foundation.h>
+#import <AppKit/NSSpeechRecognizer.h>
+
+/**
+ * GSSpeechRecognitionEngine is an abstract speech server.  One concrete subclass should
+ * be implemented for each speech engine.  Currently, only one may be compiled
+ * in to the speech server at any given time.  This limitation may be removed
```
fredkiefer: "into"

Tools/speech_recognizer/GSSpeechRecognitionServer.m
```diff
@@ -0,0 +1,69 @@
...
+}
+
++ (id)sharedServer
+{
+  return sharedInstance;
+}
+
+- (id)init
+{
+  if (nil == (self = [super init])) { return nil; }
```
fredkiefer: Normally we do not use this one line variant.

Tools/speech_recognizer/GSSpeechRecognizer.h
```diff
@@ -0,0 +1,8 @@
+#import "GSSpeechRecognitionServer.h"
+#import <AppKit/NSSpeechRecognizer.h>
+
+
+@interface GSSpeechRecognizer : NSSpeechRecognizer {
+}
+- (id)init;
```
fredkiefer: You do not need to declare this method.

Tools/speech_recognizer/GSSpeechRecognizer.m
```diff
@@ -0,0 +1,73 @@
...
+  self = [super init];
+  if (self != nil)
+    {
+    }
+  return self;
+}
+
+- (void)dealloc
+{
+  clients--;
```
fredkiefer: I don't see where clients ever gets increased.

Source/NSSpeechRecognizer.m
```diff
@@ -23,8 +23,163 @@
...
+  _speechRecognitionClass = [NSSpeechRecognizer class];
+  _speechRecognitionServer = [[NSConnection rootProxyForConnectionWithRegisteredName: SPEECH_RECOGNITION_SERVER
+                                                                                host: nil] retain];
+  if (nil == _speechRecognitionServer)
+    {
+      NSWorkspace *ws = [NSWorkspace sharedWorkspace];
+      [ws launchApplication: SPEECH_RECOGNITION_SERVER
+                   showIcon: NO
+                 autolaunch: NO];
+    }
```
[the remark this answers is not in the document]
gcasa: I believe in +initialize it tries to connect to an existing server.   It's in -allocWithZone: that it tries to start it and connect to it.   I followed the same logic that was used in NSSpeechSynthesizer.

Source/NSSpeechRecognizer.m
```diff
@@ -23,8 +23,163 @@
...
+           selector: @selector(processNotification:)
+               name: GSSpeechRecognizerDidRecognizeWordNotification
+             object: nil];
+    }
+  return self;
+}
+
++ (id) allocWithZone: (NSZone *)aZone
+{
+  if (self == _speechRecognitionClass)
```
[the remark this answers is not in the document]
gcasa: Done

Source/NSSpeechRecognizer.m
```diff
@@ -23,8 +23,163 @@
...
+                 autolaunch: NO];
+    }
+}
+
+- (void) processNotification: (NSNotification *)note
+{
+  NSString *word = (NSString *)[note object];
+  NSEnumerator *en = [_commands objectEnumerator];
+  id obj = nil;
+  while((obj = [en nextObject]) != nil)
```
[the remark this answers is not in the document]
gcasa: Does fast enumeration work properly over all platforms?  If so I'll go ahead and change it.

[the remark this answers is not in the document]
gcasa: Spaces changed.

Tools/speech_recognizer/GNUmakefile
gcasa: Agreed I will find a better way to do this.

Source/NSSpeechRecognizer.m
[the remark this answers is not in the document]
fredkiefer: We have special Macros for it which could be used. But even a modern for loop is supported by recent gcc. (Still do not recommend that)

Tools/speech_recognizer/GSSpeechRecognizer.h
gcasa: Removed.

Tools/speech_recognizer/GSSpeechRecognizer.m
gcasa: I added something to increase the count.

Tools/speech_recognizer/GSSpeechRecognitionServer.m
gcasa: Corrected

Tools/speech_recognizer/GNUmakefile
fredkiefer: Why not use pkg-config as described here https://cmusphinx.github.io/wiki/tutorialpocketsphinx/

Tools/speech_recognizer/GSSpeechRecognitionServer.m
fredkiefer: I don't see the difference.

Tools/speech_recognizer/GSSpeechRecognizer.m
```diff
@@ -0,0 +1,101 @@
...
+    }
+}
+
+- (id)init
+{
+  self = [super init];
+  if (self != nil)
+    {
+      NSLog(@"self = %@",self);
+      RETAIN(server);
```
fredkiefer: This line doesn't make any sense. You should rather increase your clients variable here.

Tools/speech_recognizer/GSSpeechRecognizer.m
```diff
@@ -0,0 +1,101 @@
...
+
+@implementation GSSpeechRecognizer
++ (void)initialize
+{
+  server = [GSSpeechRecognitionServer sharedServer];
+
+  RETAIN(server);
+  if (server != nil)
+    {
+      clients++;
```
fredkiefer: That would allow for exactly one client.

Tools/speech_recognizer/PocketsphinxSpeechRecognitionEngine.m
```diff
@@ -0,0 +1,197 @@
...
+  BOOL utt_started, in_speech;
+  int32 k;
+  char const *hyp;
+
+  NSLog(@"************************* In recognize");
+  if ((ad = ad_open_dev(cmd_ln_str_r(config, "-adcdev"),
+                        (int) cmd_ln_float32_r(config,
+                                               "-samprate"))) == NULL)
+    {
+      NSLog(@"Failed to open audio device");
```
fredkiefer: Shouldn't all these NSLog cases break out of this method?

Source/NSSpeechRecognizer.m
```diff
@@ -23,8 +23,174 @@
...
 
+@interface NSObject (GSSpeechRecognitionServer)
+- (NSSpeechRecognizer *) newRecognizer;
 @end
 
+@implementation NSSpeechRecognizer
+
++ (void) initialize
+{
+  _speechRecognitionServer = [NSConnection
```
fredkiefer: This part will get executed in the subclass GSSpeechRecognizer as well. Is this intentional?

Tools/speech_recognizer/GSSpeechRecognizer.m
gcasa: I apologize... I am trying to diagnose a crash that has been occurring.

Source/NSSpeechRecognizer.m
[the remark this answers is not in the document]
gcasa: DOne.

Tools/speech_recognizer/GSSpeechRecognitionServer.m
gcasa: Apologies, I changed it in the speech synthesizer.  Should be corrected now.

Tools/speech_recognizer/GSSpeechRecognizer.m
gcasa: Added to count here.

Tools/speech_recognizer/GSSpeechRecognizer.m
gcasa: Fixed

Source/NSSpeechRecognizer.m
gcasa: No, I don't think it is...

gcasa: Fixed.

Tools/speech_recognizer/GNUmakefile
gcasa: > I am very unsure whether the way the work gets split up between the NSSpeechRecognizer class and its subclass GSSpeechRecognizer that lives in a separate process is efficient. I would prefer to have Richard look at that code as well. He has a lot more experience with server processes.

I understand.   I modeled this very much after the speech synthesis component.  If we can improve how it's structured it might also be worth reviewing that side as well.

Tools/speech_recognizer/PocketsphinxSpeechRecognitionEngine.m
gcasa: Corrected.

gcasa:
> I am very unsure whether the way the work gets split up between the NSSpeechRecognizer class and its subclass GSSpeechRecognizer that lives in a separate process is efficient. I would prefer to have Richard look at that code as well. He has a lot more experience with server processes.

I understand. I modeled this very much after the speech synthesis component. If we can improve how it's structured it might also be worth reviewing that side as well.

gcasa:
There is a test driver which works on macOS in my personal repo called NSSpeechRecognizer_test.  It recognizes "hello" and "test" as commands and will display them if recognized.   It currently hangs or crashes on GNUstep.   This might be of some help to debug the issues with my implementation.

Source/NSSpeechRecognizer.m
```diff
@@ -23,8 +23,176 @@
...
++ (id) allocWithZone: (NSZone *)aZone
+{
+  if (self == [NSSpeechRecognizer class])
+    {
+      if (nil == _speechRecognitionServer && !_serverLaunchTested)
+        {
+          unsigned int i = 0;
+
+          // Wait for up to five seconds  for the server to launch, then give up.
+          for (i=0 ; i < 50 ; i++)
```
fredkiefer: Still no spaces around "="

Source/NSSpeechRecognizer.m
```diff
@@ -23,8 +23,176 @@
...
+  END_FOR_IN(_commands);
+}
+
+// Initialize
+- (instancetype) init
+{
+  self = [super init];
+  if (self != nil)
+    {
+      [[NSDistributedNotificationCenter defaultCenter]
```
fredkiefer: As I stated further down, I am really confused where which part of the code of this class will get executed. If this code here happens on the server, than we wouldn't need a distributed notification. Could you please clear up your view of the general working of this code?

gcasa: Sure...  one sec....

1)  When the application is loaded the first thing that happens is that +[NSSpeechRecognizer initialize] is called.  This method attempts to connect with the server (this covers the case where the server is already running from a previous invocation). 

2) if the server isnt running, it is started using workspace as the server is registered as an application. (like the synthesizer)

3) when -[NSSpeechRecognizer allocWithZone] is called it will try to connect to the server if it doesn't currently exist and keep doing so over 5 seconds until it gets the connection.  Once it has the connection it returns the concrete implementation provided by the GSSpeecRecognitionServer which is GSSpeechRecognizer.  

4) +[GSSpeechRecognizer initialize] then registers the "connectionDied" method to respond to NSConnectionDidDIeNotification in case the NSConnection dies.   It also initializes the shared server.   The shared server is a subclass of GSSpeechRecognitionServer.

5) In GSSpeechRecognitionServer init, it instantiates the GSSpeechRecognitionEngine which returns the concrete instance of PocketsphinxSpeechRecognitionEngine.

6)  At this point, when the method  startListening or stopListening are invoked it is done so on the GSSpeechRecognizer instance which was return and invokes _engine startListening or stopListening.   This method calls the code needed to start and stop the thread which contains the code which is listening to the mic input.

As discussed above.  This code is based on the NSSpeechSynthesizer "architecture".   I have no issues simplifying it.   As it is it is crashing when calling startListening and I am not sure why.

rfm:
Coming at this from the point of view of someone building 24*76 high availability server processes ...

DO is great for rapid development, but comes with reliability issues:  every DO request has additional failure modes (exceptions and/or hanging when there is a problem with the remote process or networking) and possible performance issues.
There's no simple answer to the performance issue (where they arise they need to be addressed on a case by case basis), but developers can be shielded from the failures, and for fault tolerant systems they need to be.

So, if the idea is for the thing to be robust, we should not be exposing a proxy to the developer.
When +allocWithZone: returns a proxy, it makes all the local instance methods irrelevant/confusing boilerplate and it means that all the user code which is expecting an NSSpeechRecognizer instance, has a theoretical requirement to wrap everything in exception handlers to deal with loss of the server or network issues.  It should work fine as long as there are no problems, but it's what I would consider rapid prototype code (something DO is extremely good for).

If we want it to be robust, a better approach would be:
1. +allocWithZone: returns a local instance of NSSpeechRecognizer
2. NSSpeechRecognizer has an ivar to hold the proxy
3. instance methods are responsible for sending messages to the proxy, insulating developers from the extra DO error handling

To make this easier, we can override -performSelector: and -performSelector:withObject:  to send the messages to the server, so we only need the exception handling in those methods (the other instance methods can the call those two).

So, for instance, the local instance methods could store (in ivars) the configuration state for the speech recogniser. before calling -performSelector:withObject: to send that config info to the server.
Then, if the server crashed, the error handling code in  -performSelector:withObject: could re-launch it, get a new proxy to the new server, configure it, and carry on without the code that's using the speech recogniser ever knowing that anything went wrong.

gcasa:
> Coming at this from the point of view of someone building 24*76 high availability server processes ... DO is great for rapid development, but comes with reliability issues: every DO request has additional failure modes (exceptions and/or hanging when there is a problem with the remote process or networking) and possible performance issues. There's no simple answer to the performance issue (where they arise they need to be addressed on a case by case basis), but developers can be shielded from the failures, and for fault tolerant systems they need to be. So, if the idea is for the thing to be robust, we should not be exposing a proxy to the developer. When +allocWithZone: returns a proxy, it makes all the local instance methods irrelevant/confusing boilerplate and it means that all the user code which is expecting an NSSpeechRecognizer instance, has a theoretical requirement to wrap everything in exception handlers to deal with loss of the server or network issues. It should work fine as long as there are no problems, but it's what I would consider rapid prototype code (something DO is extremely good for). If we want it to be robust, a better approach would be: 1. +allocWithZone: returns a local instance of NSSpeechRecognizer 2. NSSpeechRecognizer has an ivar to hold the proxy 3. instance methods are responsible for sending messages to the proxy, insulating developers from the extra DO error handling To make this easier, we can override -performSelector: and -performSelector:withObject: to send the messages to the server, so we only need the exception handling in those methods (the other instance methods can the call those two). So, for instance, the local instance methods could store (in ivars) the configuration state for the speech recogniser. before calling -performSelector:withObject: to send that config info to the server. Then, if the server crashed, the error handling code in -performSelector:withObject: could re-launch it, get a new proxy to the new server, configure it, and carry on without the code that's using the speech recogniser ever knowing that anything went wrong.

I just merged changes basically along these lines suggested by Richard, thank you. :)   The server is working now and matches the functionality on macOS.   I am making some further enhancements to increase robustness and also making the changes Fred suggested so that there are no hardcoded paths in the makefile.  I will have those merged.

One thing which was suggested a long time ago (when the speech server was written) was the idea of pluggable speech synth and recognition services.   That might be something to explore at a later date.

Source/NSSpeechRecognizer.m
gcasa: This should now be resolved.  I have simplified the architecture.

Source/NSSpeechRecognizer.m
gcasa: Fixed

Tools/speech_recognizer/GNUmakefile
gcasa: Done.

Source/NSSpeechRecognizer.m
```diff
@@ -23,8 +23,163 @@
...
+             object: nil];
+    }
+  return self;
+}
+
++ (id) allocWithZone: (NSZone *)aZone
+{
+  if (self == _speechRecognitionClass)
+    {
+      if (nil == _speechRecognitionServer && !_serverLaunchTested)
```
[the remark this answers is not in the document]
fredkiefer: This code here belongs in the special communication method Richard talked about. This will include error handling as well as (re-)starting a missing server.

configure
```diff
@@ -5736,10 +5739,15 @@ done
 
 if test $have_pocketsphinx = yes -a $have_speech_recognizer = yes -a $enable_speech_recognizer = yes; then
    BUILD_SPEECH_RECOGNIZER="speech_recognizer"
-   POCKETSPHINX_BASE_LIBS="-lpocketsphinx"
+   RECOGNIZER_BASE_LIBS=`pkg-config --libs pocketsphinx sphinxbase`
+   RECOGNIZER_BASE_CFLAGS=`pkg-config --cflags pocketsphinx sphinxbase`
+   RECOGNIZER_ENGINE_CLASS=PocketsphinxSpeechRecognitionEngine.m
```
fredkiefer: If you would use just the class name here you could use this variable to create an instance in GSSpeechRecognitionEngine with NSClassFromString().

Tools/speech_recognizer/GSSpeechRecognitionServer.m
```diff
@@ -0,0 +1,146 @@
...
+}
+
+- (void) removeFromBlockingRecognizers: (NSString *)s
+{
+  [_blocking removeObject: s];
+}
+
+- (BOOL) isBlocking: (NSString *)s
+{
+  return [_blocking containsObject: s];
```
fredkiefer: As in Highlander there can only be one blocking all others. So the logic here should be that of a stack. The object at the top blocks all other recognizers. You should ask whether a recognizer is blocked, that would be the case if the stack is not empty and the recogniser is not at the top.

Source/NSSpeechRecognizer.m
```diff
@@ -23,8 +23,241 @@
...
+  
+  if (_listensInForegroundOnly)
+    {
+      if (_appInForeground == NO)
+        {
+          return;
+        }
+    }
+
+  if (_blocksOtherRecognizers)
```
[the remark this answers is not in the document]
gcasa: Changed.

Source/NSSpeechRecognizer.m
```diff
@@ -23,8 +23,241 @@
...
+    }
+}
+
+// Initialize
+- (instancetype) init
+{
+  self = [super init];
+  if (self != nil)
+    {
+      [[NSDistributedNotificationCenter defaultCenter]
```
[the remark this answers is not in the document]
gcasa: Changed.

Source/NSSpeechRecognizer.m
```diff
@@ -23,8 +23,241 @@
...
+- (BOOL) blocksOtherRecognizers
+{
+  return _blocksOtherRecognizers;
+}
+
+- (void) setBlocksOtherRecognizers: (BOOL)blocksOtherRecognizers
+{
+  if (blocksOtherRecognizers == YES)
+    {
+      [_speechRecognitionServer addToBlockingRecognizers: [_uuid UUIDString]];
```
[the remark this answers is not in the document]
gcasa: Instead of overriding performSelector: and performSelector:withObject: I have simply wrapped everywhere where the server is called with exception/error handling logic.   There are only two places in which this is done in the recognizer so I didn't see the point in centralizing the error handling.

Source/NSSpeechRecognizer.m
```diff
@@ -23,8 +23,241 @@
...
+    {
+      if (_appInForeground == NO)
+        {
+          return;
+        }
+    }
+
+  if (_blocksOtherRecognizers)
+    {
+      if ([_speechRecognitionServer isBlocking: [_uuid UUIDString]] == NO)
```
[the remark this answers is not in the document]
gcasa: Same comment as below.   I wrapped the places where error handling is needed individually since it only happens in two places.

Tools/speech_recognizer/GSSpeechRecognitionServer.m
gcasa: I was always partial to Ramirez.... he had style... plus he was played by Sean Connery. ;) Anyway, I have changed this.

Source/NSSpeechRecognizer.m
```diff
@@ -23,8 +23,279 @@
...
+        {
+          NSWorkspace *ws = [NSWorkspace sharedWorkspace];
+          [ws launchApplication: SPEECH_RECOGNITION_SERVER
+                       showIcon: NO
+                     autolaunch: NO];
+        }
+    }
+}
+
+- (void) _restartServer
```
fredkiefer: This only reconnects but does not restart the server.

Source/NSSpeechRecognizer.m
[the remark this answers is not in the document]
fredkiefer: No, it still checks only for blocking recognizers.

Source/NSSpeechRecognizer.m
```diff
@@ -23,8 +23,279 @@
...
+          if ([_speechRecognitionServer isBlocking: [_uuid UUIDString]] == NO)
+            {
+              // If we are not a blocking recognizer, then we are blocked...
+              return;
+            }
+        }
+      NS_HANDLER
+        {
+          NSLog(@"%@", localException);
+          [self _restartServer];
```
fredkiefer: This is some error handling, but not sufficient. If the connection is broken you should reconnect and retry the same call. That is why Richard suggested the performXXX methods. That way a retry is fairly simple.

configure
gcasa: One of the things which was thought of, but not implemented in the speech synthesis system is to use bundles instead of linking the class directly.  What I am going to do is the following:

I am going to create a bundles directory in speech_synthesizer.   I am going to move the engine into that and separate out everything into a bundle for pocketsphinx.  Right now this will be the only one, but it will make it easy to add them and, ultimately, select between them.   We could even do speech recognizer bundles in programs separately or have them in their own directory in the system library.

Just some thoughts.   Right now it's proving to be a pain to add this so that the .m file can see it.   So why add something now that I am going to feel compelled to remove later.

Tools/speech_recognizer/GSSpeechRecognitionServer.m
```diff
@@ -0,0 +1,146 @@
...
+}
+
+- (void) removeFromBlockingRecognizers: (NSString *)s
+{
+  [_blocking removeObject: s];
+}
+
+- (BOOL) isBlocking: (NSString *)s
+{
+  return [[_blocking firstObject] isEqualToString: s];
```
fredkiefer: You add to the end of the array, this means your stack is the other way around.You need to check the last element. Also you should return false if the array is empty.

rfm:
> On 10 Feb 2020, at 08:22, Fred Kiefer <notifications@github.com> wrote:
> 
> @fredkiefer commented on this pull request.
> 
> In Tools/speech_recognizer/GSSpeechRecognitionServer.m:
> 
> > +  [super dealloc];
> +}
> +
> +- (void) addToBlockingRecognizers: (NSString *)s
> +{
> +  [_blocking addObject: s];
> +}
> +
> +- (void) removeFromBlockingRecognizers: (NSString *)s
> +{
> +  [_blocking removeObject: s];
> +}
> +
> +- (BOOL) isBlocking: (NSString *)s
> +{
> +  return [[_blocking firstObject] isEqualToString: s];
> 
> You add to the end of the array, this means your stack is the other way around.You need to check the last element. Also you should return false if the array is empty.

It would be sufficient to do:
return [[_blocking lastObject] isEqualToString: s];
If the array is empty, this would send an -isEqualToString: message to nil, which would return zero, so the method would return NO.

You could add an explicit check for an empty array to make the code clearer perhaps, but a single line implementation of the method would be correct (and efficient).

configure
gcasa: I'm going to defer this one until later.   The bundle approach should work it's just not loading fast enough.

Source/NSSpeechRecognizer.m
gcasa: It will now restart the server if it is not started.

Source/NSSpeechRecognizer.m
gcasa: At this point the server works.   Overbuilding it for reliability doesn't seem super necessary since it's not a mission critical process.   I will take this as a note for the next release along with the bundles for both the recognizer and the synthesizer.

gcasa:
Hey guys.  Please let me know if it is okay to merge.  All changes requested have been made except those I have deferred.

gcasa:
Will be merging by noon local time (US-EST) if there is no feedback by then.

rfm:
> On 15 Feb 2020, at 15:10, Gregory Casamento <notifications@github.com> wrote:
> 
> Will be merging by noon local time (US-EST) if there is no feedback by then.
> 

As far as I'm concerned, go ahead.  I have not been able to review properly, but I don't see anything *obviously* breaking compilation, and I'm a believer in getting *new* code out on the main branch asap (or even developing there).
The utility (for me) of separate branches is really to ensure safety when modifying existing working code. not for completely new stuff that's really separate from the existing code.

gcasa:
> On 15 Feb 2020, at 15:10, Gregory Casamento ***@***.***> wrote: Will be merging by noon local time (US-EST) if there is no feedback by then.
> As far as I'm concerned, go ahead. I have not been able to review properly, but I don't see anything *obviously* breaking compilation, and I'm a believer in getting *new* code out on the main branch asap (or even developing there). The utility (for me) of separate branches is really to ensure safety when modifying existing working code. not for completely new stuff that's really separate from the existing code.

Thanks.  I will go ahead.  Also it is tested and working.   I don't like things sitting out too long either.

PR #54, opened 2020-02-12
New feature: Main menu follows key window

trunkmaster:
This is initial implementation of feature for the following scenarios with multi-monitor setup:
1. Key window was moved to other screen (monitor).
2. Several windows of one application are placed on different monitors and key window was changed.

In both cases all menus (main, detached submenus) moved into screen where key window resides.


Source/NSMenu.m
```diff
@@ -712,6 +712,15 @@ - (id) initWithTitle: (NSString*)aTitle
       name: NSEnqueuedMenuMoveName
       object: self];
   
+  [nc addObserver: self
```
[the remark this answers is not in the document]
trunkmaster: No, detached submenus should move with main menu.

Source/NSMenu.m
```diff
@@ -1860,6 +1869,36 @@ - (void) _showOnActivateApp: (NSNotification*)notification
...
+  NSRect   frame;
+  NSRect   oldScreenFrame;
+  NSRect   newScreenFrame;
+  NSPoint  offset;
+
+  if (window == _aWindow
+      || [window isKeyWindow] == NO
+      || [_aWindow screen] == [window screen]
+      || [_aWindow isVisible] == NO)
+    return;
```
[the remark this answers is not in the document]
trunkmaster: Done.

Source/NSMenu.m
```diff
@@ -1860,6 +1869,36 @@ - (void) _showOnActivateApp: (NSNotification*)notification
...
+  // Keep left offset fixed
+  offset.x = frame.origin.x - oldScreenFrame.origin.x;
+  frame.origin.x = newScreenFrame.origin.x + offset.x;
+
+  // Keep top offset fixed
+  offset.y = NSMaxY(oldScreenFrame) - NSMaxY(frame);
+  frame.origin.y = NSMaxY(newScreenFrame) - offset.y - frame.size.height;
+  
+  // setFrame: changes _screen value.
+  [_aWindow setFrame: frame display: NO];
```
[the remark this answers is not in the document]
trunkmaster: What do you mean by "very different"? I use 2 monitors with resolutions 1024x768 and 640x480. Works good.

Source/NSWindow.m
```diff
@@ -4180,6 +4180,17 @@ window that has been ordered out (and thus is logically invisible). We
                 }
               [nc postNotificationName: NSWindowDidMoveNotification
                   object: self];
+              // Check if window moved to an other screen
+              {
+                NSScreen *newScreen = [self _screenForFrame: _frame];
```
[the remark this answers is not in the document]
trunkmaster: > 
> 
> It would be better to call [self screen] here as this is already optimised for the common case that the screen stays the same. But you will have to store _screen in a local variable _oldScreen as it will change. Or even better move the notification over to that method. 

Done.

>When we have proper multi monitor support 

What do you mean by "proper"? I think we already have proper multi-monitor support.

> windows may change the screen without being moved, just by rearranging the monitors. This will need to be handled as well.

Agree. I think it should be a separate task/PR.

Source/NSMenu.m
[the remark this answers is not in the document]
trunkmaster: I have 2 monitors setup: 1024x768 and 640x480. Quite different sizes, isn't it? Even if I place main menu off the bottom of the screen and then move window - menu stays visible on new screen.

Source/NSMenu.m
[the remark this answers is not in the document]
fredkiefer: I think the main window will take care of that.

Source/NSWindow.m
```diff
@@ -4170,16 +4170,27 @@ window that has been ordered out (and thus is logically invisible). We
...
-              if (_autosaveName != nil)
-                {
-                  [self saveFrameUsingName: _autosaveName];
-                }
-              [nc postNotificationName: NSWindowDidMoveNotification
-                  object: self];
+              {
+                NSScreen *oldScreen;
+                NSScreen *newScreen;
+                oldScreen = [self screen];
```
fredkiefer: You could use the instance variable here. The frame may already have changed if the move was initiated by code.

Source/NSWindow.m
```diff
@@ -2702,6 +2702,8 @@ - (NSScreen *) screen
         }
     }
   ASSIGN(_screen, [self _screenForFrame: _frame]);
+  [nc postNotificationName: NSWindowDidChangeScreenNotification
```
fredkiefer: This did not work as you did not check whether the screen did change here.

trunkmaster: This code is outdated.

Source/NSMenu.m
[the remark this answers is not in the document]
trunkmaster: My tests show that it's not true.

Source/NSWindow.m
trunkmaster: Indeed this is a problem. If position was changed programmatically (with setFrame:display:)  instance variable already changed. Probably I need to move this code into setFrame...

trunkmaster: Done. I've removed reassign of `_screen` in `setFrame:display:` and `setFrameFromString:` methods. Everything works correctly.
The results: 1. Less places in code where `_screen` is changing. 2. No need to use extra instance variable.

The idea behind this change is the following: change to window frame can be applied with these methods. Both methods will call to [GSDisplayServer placewindow::] and placewindow:: will generate GSAppKitWindowMoved event from -back to -gui. What do you think?

trunkmaster: Further thinking about it I've came to a conclusion that some logic should be quite different. For the case when window gets its frame programmatically window should be opened in screen where main menu resides (mainScreen). In that case origin of new window should adjusted to the frame of main screen.

It sounds good from usability perspective - user expects new window to appear on screen where main menu resides. 

I'm going to implement it and see how it works.

Source/NSMenu.m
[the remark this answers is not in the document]
fredkiefer: If this is not the case then something else is going wrong. Moving a menu should mostly be done via the method nestedSetFrameOrigin: that takes care of sub menus as well. Could you please find out where a call to that is missing?

Source/NSMenu.m
```diff
@@ -1860,6 +1869,38 @@ - (void) _showOnActivateApp: (NSNotification*)notification
...
+      return;
+    }
+
+  oldScreenFrame = [[_aWindow screen] frame];
+  newScreenFrame = [[window screen] frame];
+  frame = [_aWindow frame];
+  
+  // Keep left offset fixed
+  offset.x = frame.origin.x - oldScreenFrame.origin.x;
+  frame.origin.x = newScreenFrame.origin.x + offset.x;
```
fredkiefer: I think it would be more clear to write this as

frame.origin.x += newScreenFrame.origin.x - oldScreenFrame.origin.x;

The y direction is not as simple.

Source/NSMenu.m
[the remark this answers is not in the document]
fredkiefer: OK, this means some other code in NSMenu takes care of that.

Source/NSMenu.m
trunkmaster: Done.

Source/NSWindow.m
fredkiefer: There are different cases where the frame gets set programatically. The method setFrameFromString: is just one of them, center would be another.
And please also remember that there are other backends (MS Windows and soon Wayland), where the interaction between gui and back might be different.

Source/NSWindow.m
```diff
@@ -5022,9 +5024,6 @@ - (void) setFrameFromString: (NSString*)string
         }
     }
 
-  // Make sure we are using the new screen we are applying to...
```
fredkiefer: You probably removed this, because the later interaction will also try to set the screen. But now we have a window that does not have a screen assigned for some time. Is this a problem?

Source/NSMenu.m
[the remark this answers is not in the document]
trunkmaster: Found (windowDidChangeScreen) and fixed. Also I've rearranged notification observer setup to make receiving of notification by main menu an torn off menus. Plus do not reposition menus if panel has changed screen.

Source/NSWindow.m
trunkmaster: This is not a problem because window has screen assigned in initializer methods (initWithContentRect...).

Source/NSWindow.m
```diff
@@ -2249,12 +2249,13 @@ - (void) setFrame: (NSRect)frameRect display: (BOOL)flag
...
     }
-      
+
   /* Windows need to be constrained when displayed or resized - but only
      titled windows are constrained */
   if (_styleMask & NSTitledWindowMask)
     {
-      frameRect = [self constrainFrameRect: frameRect toScreen: [self screen]];
+      frameRect = [self constrainFrameRect: frameRect
+                                  toScreen: [self _screenForFrame: frameRect]];
```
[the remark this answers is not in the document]
trunkmaster: I can agree that this could be slow at some degree (how often window changes its frame?). But why is it wrong?

Source/NSWindow.m
trunkmaster: Yes, but center calls setFrame:display: finally. The call sequence is following `center -> setFrameOrigin: -> setFrame:display:`. That's why I think `setFrame:display:` is a good single place to tweak `_screen` value.

As for other backends: MS Windows should use in-window menus so this is not a case. Wayland - as new backend - should adopt current approach or propose new, better.

Source/NSWindow.m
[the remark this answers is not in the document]
fredkiefer: Wrong in the sense that it makes the code more complicated. I see _screenForFrame: as an internal method that should only be used in very few places.

Source/NSMenu.m
```diff
@@ -1860,6 +1889,41 @@ - (void) _showOnActivateApp: (NSNotification*)notification
...
 
+- (void) windowDidChangeScreen: (NSNotification*)notification
+{
+  NSWindow *window = [notification object];
+  NSRect   frame;
+  NSRect   oldScreenFrame;
+  NSRect   newScreenFrame;
+  CGFloat  yOffset;
+
+  if ([window isKindOfClass: [NSMenuPanel class]]
```
fredkiefer: Could we merge these two if statements into one?

trunkmaster: Sure, done.

Source/NSWindow.m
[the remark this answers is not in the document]
trunkmaster: I understand. Although there is a problem with `[self screen]` usage here. At this point new values for `_frame` are not set yet. `-screen` method makes its calculations using `_frame` ivar values. So basically current code is wrong: it determines wrong screen. I consider using `_screen` instead of `[self screen]` at this call. Otherwise we need to set `_frame` to `newFrame` before `-constrain...` call and revert after. What do you think?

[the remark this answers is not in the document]
fredkiefer: You are right, sorry I did overlook this case.

Source/NSWindow.m
fredkiefer: OK, so back to the start, why don't you use the instance variable to get the oldScreen?

Source/NSWindow.m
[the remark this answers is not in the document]
trunkmaster: I have a better solution for this. I'll make separate PR (this involves PR into libs-back also).

Source/NSWindow.m
trunkmaster: No problem, done.

PR #56, opened 2020-03-05
Window placement fix on screen change

trunkmaster:
Call backend's `placewindow::` directly to change window placement even if `_frame` is not changed. Corresponding PR in back is [#24](https://github.com/gnustep/libs-back/pull/24) - allows window reposition in Xlib coordinate system for this case.
Made cleanup in `GSAppKitWindowMoved` event handling - removed extra variables.

### DIFF
--- a/Source/NSWindow.m
+++ b/Source/NSWindow.m
@@ -4175,12 +4175,10 @@ checkCursorRectanglesExited(NSView *theView,  NSEvent *theEvent, NSPoint lastPoi
             {
             case GSAppKitWindowMoved:
               {
-                NSScreen *oldScreen;
-                NSScreen *newScreen;
-                oldScreen = _screen;
+                NSScreen *oldScreen = _screen;
+                
                 _frame.origin.x = (CGFloat)[theEvent data1];
                 _frame.origin.y = (CGFloat)[theEvent data2];
-                newScreen = [self screen];
                 NSDebugLLog(@"Moving", @"Move event: %d %@",
                             (int)_windowNum, NSStringFromPoint(_frame.origin));
                 if (_autosaveName != nil)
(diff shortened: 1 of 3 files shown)


Source/NSWindow.m
```diff
@@ -2737,10 +2737,24 @@ - (void) applicationDidChangeScreenParameters: (NSNotification *)aNotif
...
-  if (_autosaveName != nil)
+
+  if (_windowNum)
     {
-      [self saveFrameUsingName: _autosaveName];
+      /* Call backend's `placewindow::` directly because our origin in OpenStep 
+         coordinates might left unchanged and `setFrame:display:` has check 
+         for it. */
+      [GSServerForWindow(self) placewindow: newFrame : _windowNum];
+      if (_autosaveName != nil)
```
fredkiefer: I think this part belongs after the outer if. Even when the window was not instantiated the frame could be saved.

As for the rest of the code I really would prefer not to duplicate this code. Maybe we could use a new helper method from both places?

Source/NSWindow.m
```diff
@@ -2737,10 +2737,24 @@ - (void) applicationDidChangeScreenParameters: (NSNotification *)aNotif
...
   // Screen X origin change. Screen width change shouldn't affect our frame.
   newFrame.origin.x += newScreenFrame.origin.x - oldScreenFrame.origin.x;
-  [self setFrame: newFrame display: NO];
-  if (_autosaveName != nil)
+
+  if (_windowNum)
     {
-      [self saveFrameUsingName: _autosaveName];
+      /* Call backend's `placewindow::` directly because our origin in OpenStep 
+         coordinates might left unchanged and `setFrame:display:` has check
```
fredkiefer: "might be unchanged"

trunkmaster: Of course, thanks! :-)

Source/NSWindow.m
trunkmaster: Done. I hope it looks much better.

PR #60, opened 2020-04-20
NSSwitch branch

gcasa:
This branch adds the NSSwitch widget the GNUstep's GUI library.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,23 @@
+2020-04-26 Gregory John Casamento <greg.casamento@gmail.com>
+
+	* Headers/Additions/GNUstepGUI/GSTheme.h: Add declarations
+        for new drawing methods.
+	* Headers/AppKit/AppKit.h: Add new headers
+	* Headers/AppKit/NSAccessibilityProtocols.h
+	* Headers/AppKit/NSCell.h: Add NSControlStateValue enum
+	* Headers/AppKit/NSSwitch.h: Header for NSSwitch
+	* MISSING: Remove NSSwitch
+	* Source/GNUmakefile: Add .h and .m files
+	* Source/GSThemeDrawing.m: Add drawing code for knob and
+        switch basin.
+	* Source/GSXib5KeyedUnarchiver.m: Add decoding for
+        NSControlContents.  Make sure enabled is YES when not
+        present.
+	* Source/NSColor.m: Minor changes
+	* Source/NSMatrix.m: Remove uneeded code
+	* Source/NSSwitch.m: Implementation of NSSwitch
+	* Source/NSToolbarItem.m: Remove uneeded code
+
 2020-04-13 Gregory John Casamento <greg.casamento@gmail.com>
 
 	* Headers/AppKit/AppKitErrors.h:
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,31 @@
+2020-04-13  Ivan Vucica <ivan@vucica.net>
+
+	* Documentation/announce.texi:
+	* ANNOUNCE:
+	Normalize the accompanying text for the release announcement across
+	core packages: standardize chapter name and GPG information.
+
+2020-04-05  Ivan Vucica  <ivan@vucica.net>
+
+	* ANNOUNCE:
+	* Documentation/ReleaseNotes.gsdoc:
+	* Documentation/news.texi:
+	* NEWS:
+	Updating documentation for the 0.28.0 release.
+
+2020-04-01  Wolfgang Lux  <wolfgang.lux@gmail.com>
+
+	* Source/NSAlert.m (NSBeginAlertSheet): Restore call to delegate
+	method which was lost a while ago.
+
+	* Source/NSAlert.m (NSBeginCriticalAlertSheet,
+	NSBeginInformationalAlertSheet): Remove redundant calls to close
+	method.
+
+	* Source/NSAlert.m (NSBeginAlertSheet, NSBeginCriticalAlertSheet,
+	NSBeginInformationalAlertSheet): Change arguments names to reduce
+	confusion about the intended use of the corresponding selectors.
+
 2020-03-26 Fred Kiefer <FredKiefer@gmx.de>
 
 	* Documentation/news.texi: Started to fill in some details for
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,30 @@
+2020-04-13 Gregory John Casamento <greg.casamento@gmail.com>
+
+	* Headers/AppKit/AppKitErrors.h:
+	* Headers/AppKit/AppKit.h:
+	* Headers/AppKit/NSFontAssetRequest.h:
+	* Headers/AppKit/NSMediaLibraryBrowserController.h:
+	* Headers/AppKit/NSScrubber.h:
+	* Headers/AppKit/NSScrubberItemView.h:
+	* Headers/AppKit/NSScrubberLayout.h:
+	* Headers/AppKit/NSSharingServicePickerToolbarItem.h:
+	* Headers/AppKit/NSUserInterfaceCompression.h:
+	* Headers/AppKit/NSUserInterfaceItemSearching.h:
+	* Headers/AppKit/NSWindowRestoration.h:
+	* MISSING: Remove mention of these classes.
+	* Source/GNUmakefile
+	* Source/NSFontAssetRequest.m:
+	* Source/NSMediaLibraryBrowserController.m:
+	* Source/NSScrubberItemView.m:
+	* Source/NSScrubberLayout.m:
+	* Source/NSScrubber.m:
+	* Source/NSSharingServicePickerToolbarItem.m: Add code to allow
+        compilation using these classes.  These are Mac specific so they
+        can't be implemented on any other platform due to HW or
+        service limitations.  At some point these might be able to be
+        implemented, but, for now, they are here to allow apps which
+        use them to compile.
+
 2020-04-13  Ivan Vucica <ivan@vucica.net>
 
 	* Documentation/announce.texi:
--- a/Headers/AppKit/AppKit.h
+++ b/Headers/AppKit/AppKit.h
@@ -47,6 +47,7 @@
 
 #import <AppKit/NSAccessibility.h>
 #import <AppKit/NSAccessibilityConstants.h>
+#import <AppKit/NSAccessibilityProtocols.h>
 #import <AppKit/NSActionCell.h>
 #import <AppKit/NSAnimationContext.h>
 #import <AppKit/NSAppearance.h>
--- a/Headers/AppKit/AppKit.h
+++ b/Headers/AppKit/AppKit.h
@@ -42,6 +42,7 @@
 //
 // GNUstep GUI Library functions
 //
+#import <AppKit/AppKitErrors.h>
 #import <AppKit/NSGraphics.h>
 
 #import <AppKit/NSAccessibility.h>
@@ -117,6 +118,9 @@
 #import <AppKit/NSScreen.h>
 #import <AppKit/NSScroller.h>
 #import <AppKit/NSScrollView.h>
+#import <AppKit/NSScrubber.h>
+#import <AppKit/NSScrubberItemView.h>
+#import <AppKit/NSScrubberLayout.h>
 #import <AppKit/NSSelection.h>
 #import <AppKit/NSSharingService.h>
 #import <AppKit/NSSlider.h>
@@ -157,6 +161,7 @@
 #import <AppKit/NSDocumentController.h>
 #import <AppKit/NSDrawer.h>
 #import <AppKit/NSFileWrapperExtensions.h>
+#import <AppKit/NSFontAssetRequest.h>
 #import <AppKit/NSFontCollection.h>
 #import <AppKit/NSFontDescriptor.h>
 #import <AppKit/NSGestureRecognizer.h>
@@ -173,6 +178,7 @@
 #import <AppKit/NSLevelIndicator.h>
 #import <AppKit/NSLevelIndicatorCell.h>
 #import <AppKit/NSMagnificationGestureRecognizer.h>
+#import <AppKit/NSMediaLibraryBrowserController.h>
 #import <AppKit/NSMovie.h>
 #import <AppKit/NSMovieView.h>
 #import <AppKit/NSPanGestureRecognizer.h>
@@ -201,6 +207,7 @@
 #import <AppKit/NSSegmentedCell.h>
 #import <AppKit/NSSegmentedControl.h>
 #import <AppKit/NSShadow.h>
+#import <AppKit/NSSharingServicePickerToolbarItem.h>
 #import <AppKit/NSSharingServicePickerTouchBarItem.h>
 #import <AppKit/NSSliderTouchBarItem.h>
 #import <AppKit/NSSound.h>
@@ -235,6 +242,8 @@
 #import <AppKit/NSTreeNode.h>
 #import <AppKit/NSUserDefaultsController.h>
 #import <AppKit/NSUserInterfaceItemIdentification.h>
+#import <AppKit/NSUserInterfaceCompression.h>
+#import <AppKit/NSUserInterfaceItemSearching.h>
 #import <AppKit/NSUserInterfaceLayout.h>
 #import <AppKit/NSUserInterfaceValidation.h>
 #import <AppKit/NSViewController.h>
--- a/Headers/AppKit/AppKit.h
+++ b/Headers/AppKit/AppKit.h
@@ -216,6 +216,7 @@
 #import <AppKit/NSStepperTouchBarItem.h>
 #import <AppKit/NSStepper.h>
 #import <AppKit/NSStepperCell.h>
+#import <AppKit/NSSwitch.h>
 #import <AppKit/NSTableColumn.h>
 #import <AppKit/NSTableHeaderCell.h>
 #import <AppKit/NSTableHeaderView.h>
--- a/Headers/AppKit/NSCell.h
+++ b/Headers/AppKit/NSCell.h
@@ -123,6 +123,16 @@ enum {
 };
 typedef NSUInteger NSCellStateValue;
 
+/*
+ * Control state values as of 10.13
+ */ 
+enum { 
+  NSControlStateValueMixed	= -1,
+  NSControlStateValueOff	=  0,
+  NSControlStateValueOn		=  1
+};
+typedef NSUInteger NSControlStateValue;
+
 /**
  *  <p>Enumeration of the ways that you can display an image in an
  *  NSImageCell.  The available ones are:</p>
--- a/Headers/AppKit/NSCell.h
+++ b/Headers/AppKit/NSCell.h
@@ -123,15 +123,17 @@ enum {
 };
 typedef NSUInteger NSCellStateValue;
 
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_13, GS_API_LATEST)
 /*
  * Control state values as of 10.13
  */ 
 enum { 
-  NSControlStateValueMixed	= -1,
-  NSControlStateValueOff	=  0,
-  NSControlStateValueOn		=  1
+  NSControlStateValueMixed = -1,
+  NSControlStateValueOff   =  0,
+  NSControlStateValueOn    =  1
 };
 typedef NSUInteger NSControlStateValue;
+#endif
 
 /**
  *  <p>Enumeration of the ways that you can display an image in an
--- a/Headers/AppKit/NSSwitch.h
+++ b/Headers/AppKit/NSSwitch.h
@@ -1,0 +1,47 @@
+/* Definition of class NSSwitch
+   Copyright (C) 2020 Free Software Foundation, Inc.
+   
+   By: Gregory John Casamento
+   Date: Wed Apr  8 22:01:02 EDT 2020
+
+   This file is part of the GNUstep Library.
+   
+   This library is free software; you can redistribute it and/or
+   modify it under the terms of the GNU Lesser General Public
+   License as published by the Free Software Foundation; either
+   version 2.1 of the License, or (at your option) any later version.
+   
+   This library is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+   Lesser General Public License for more details.
+   
+   You should have received a copy of the GNU Lesser General Public
+   License along with this library; if not, write to the Free
+   Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+   Boston, MA 02110 USA.
+*/
+
+#ifndef _NSSwitch_h_GNUSTEP_GUI_INCLUDE
+#define _NSSwitch_h_GNUSTEP_GUI_INCLUDE
+
+#import <AppKit/NSControl.h>
+
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_0, GS_API_LATEST)
+
+#if	defined(__cplusplus)
+extern "C" {
+#endif
+
+@interface NSSwitch : NSControl
+
+@end
+
+#if	defined(__cplusplus)
+}
+#endif
+
+#endif	/* GS_API_MACOSX */
+
+#endif	/* _NSSwitch_h_GNUSTEP_GUI_INCLUDE */
+
--- a/Headers/AppKit/NSSwitch.h
+++ b/Headers/AppKit/NSSwitch.h
@@ -35,7 +35,7 @@
 extern "C" {
 #endif
 
-@interface NSSwitch : NSControl
+@interface NSSwitch : NSControl <NSAccessibilitySwitch>
 {
   NSControlStateValue _state;
   id _target;
--- a/Headers/AppKit/NSSwitch.h
+++ b/Headers/AppKit/NSSwitch.h
@@ -40,6 +40,7 @@ extern "C" {
   NSControlStateValue _state;
   id _target;
   SEL _action;
+  BOOL _enabled;
 }
 
 - (void) setState: (NSControlStateValue)s;
--- a/Headers/AppKit/NSSwitch.h
+++ b/Headers/AppKit/NSSwitch.h
@@ -26,15 +26,22 @@
 #define _NSSwitch_h_GNUSTEP_GUI_INCLUDE
 
 #import <AppKit/NSControl.h>
+#import <AppKit/NSCell.h>
 
-#if OS_API_VERSION(MAC_OS_X_VERSION_10_0, GS_API_LATEST)
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_15, GS_API_LATEST)
 
 #if	defined(__cplusplus)
 extern "C" {
 #endif
 
 @interface NSSwitch : NSControl
+{
+  NSControlStateValue _state;
+}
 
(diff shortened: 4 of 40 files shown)


gcasa:
One thing that is happening is that I can't make a connection.  The connections are there in the test code I wrote NSSwitch_test... but even when unarchiving it is not showing the connection.   I added this pull request a bit early so you might be able to give me an idea as to why that is not working.

Source/NSSwitch.m
```diff
@@ -0,0 +1,135 @@
...
+{
+  return _target;
+}
+
+- (void) drawRect: (NSRect)rect
+{
+  [[GSTheme theme] drawSwitchInRect: rect
+                           forState: _state
+                            enabled: [self isEnabled]
+                       bezelOnColor: [NSColor selectedControlColor]
```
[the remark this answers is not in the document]
gcasa: After you create a macOS project

Click the + on the upper right of the xcode window.  That should show you the library of controls.  You can simply search for NSSwitch.

I hope you don't want to make the NSSwitch square.  Right now I think I have a good analog of what it looks like on the mac.   I thought about reusing the button logic, but decided against it.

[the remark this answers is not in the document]
fredkiefer: Looks like the reason I cannot use an NSSwitch is that I am still using MacOS 10.14. Could you please provide me with a screenshot showing an NSSwitch plus an NSButton in switch mode of the same size on MacOS 10.15.
You could also do the same screenshot on GNUstep allowing for a comparison.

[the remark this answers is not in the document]
gcasa: I am selecting switch state, but the button stays the same.. :/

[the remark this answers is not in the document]
gcasa: button_switch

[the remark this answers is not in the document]
gcasa: It looks the same when running as well and is non-responsive. :/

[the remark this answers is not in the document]
gcasa: There is no switch mode in Gorm for the button... I will check the code.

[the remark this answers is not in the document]
fredkiefer: Thank you for the screen shot! Now I understand what a switch should look like.
On NSButton the type switch only makes sense when combined with the style check,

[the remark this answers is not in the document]
gcasa: here are the images used for the switch mode.  I am surprised I never added this to Gorm, but I will in the next release.   I apologize for not having the time to show this at runtime, but I am busy with day job stuff and should be able to do a more thorough discussion later.
switch_on_off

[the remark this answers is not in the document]
gcasa: I very much disagree with making NSSwitch (the class) look like NSButton in switch mode.   It seems just... backwards.

Source/GSThemeTools.m
```diff
@@ -481,7 +481,6 @@ - (void) drawCircularBezel: (NSRect)cellFrame
   [[NSColor shadowColor] set];
   [oval stroke];
 }
-
```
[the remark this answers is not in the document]
gcasa: I believe it was just a mistake.

gcasa:
Seems like there is some logic missing in GSXibUnarchiver....   According to these files NSControlContent corresponds to the state="on" portion of the XIB file for the switches, yet it is not found when loading the file...

[switch1.txt](https://github.com/gnustep/libs-gui/files/4515090/switch1.txt)
[switch2.txt](https://github.com/gnustep/libs-gui/files/4515091/switch2.txt)
[switch3.txt](https://github.com/gnustep/libs-gui/files/4515092/switch3.txt)

Headers/AppKit/NSSwitch.h
```diff
@@ -0,0 +1,57 @@
...
+   License along with this library; if not, write to the Free
+   Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+   Boston, MA 02110 USA.
+*/
+
+#ifndef _NSSwitch_h_GNUSTEP_GUI_INCLUDE
+#define _NSSwitch_h_GNUSTEP_GUI_INCLUDE
+
+#import <AppKit/NSControl.h>
+#import <AppKit/NSCell.h>
```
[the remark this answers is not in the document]
gcasa: The NSCell.h file is included since that is where NSControlStateValueOn/Off is defined.   I am adding the methods for the protocol now.

Source/NSSwitch.m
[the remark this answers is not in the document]
gcasa: I'm going to resolve this, if you feel strongly about the look/feel then reopen it.   I also made the adjustments you suggested in your email regarding changing the size of the knob by 2 pixels.  I did one pixel and moved it down a little, so I think it looks pretty perfect, but you decide.

gcasa:
> Most of the NSControl behaviour seems to be missing.

I believe I have implemented all of them.

gcasa:
Awaiting further comment or approval.

gcasa:
> Awaiting further comment or approval.

Finishing up implementation of methods (per NSScroller)

Headers/AppKit/NSCell.h
```diff
@@ -123,6 +123,16 @@ enum {
 };
 typedef NSUInteger NSCellStateValue;
 
+/*
```
fredkiefer, line 126: Should we add a version guard around this?

Headers/AppKit/NSCell.h
```diff
@@ -123,6 +123,16 @@ enum {
...
+ * Control state values as of 10.13
+ */ 
+enum { 
+  NSControlStateValueMixed	= -1,
+  NSControlStateValueOff	=  0,
+  NSControlStateValueOn		=  1
```
fredkiefer, line 132: Indentation of the "=" looks wrong.

Source/GSThemeDrawing.m
```diff
@@ -816,6 +816,129 @@ - (void) drawStepperCell: (NSCell*)cell
...
+{
+  NSBezierPath *p;
+  NSPoint point;
+  CGFloat radius;
+  NSColor *backgroundColor = [NSColor windowBackgroundColor]; // offColor;
+
+  if (NSControlStateValueOn == v)
+    {
+      backgroundColor = [NSColor selectedControlColor]; // onColor;
+    }
```
fredkiefer, line 832: I would prefer to set eh off colour in an else branch.

Source/GSThemeDrawing.m
```diff
@@ -816,6 +816,129 @@ - (void) drawStepperCell: (NSCell*)cell
...
+  
+  // make smaller than enclosing frame
+  frame = NSInsetRect(frame, 4, 4);
+  radius = frame.size.height / 2.0;
```
fredkiefer, line 836: What happens if width is smaller than height?

Source/GSThemeTools.m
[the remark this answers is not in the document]
fredkiefer: Why don't you just undo this?

Headers/AppKit/NSCell.h
gcasa: Weird, I don't know what is wrong with my emacs settings... Ugh.

Headers/AppKit/NSCell.h
gcasa: I believe so.  I will track down when this change was made and do that.

Source/GSThemeDrawing.m
gcasa: Indeed.  Will do that.

Source/GSThemeDrawing.m
gcasa: I appears as though macOS will never generate anything other than the sizes I have in the test, so it shouldn't be an issue, but this is a good point.   I will find a better way to calculate this.

Source/NSSwitch.m
```diff
@@ -0,0 +1,259 @@
...
+{
+  return (NSInteger)[self doubleValue];
+}
+
+- (void) setStringValue: (NSString *)val
+{
+  [self setDoubleValue: [val doubleValue]];
+}
+
+- (NSString *) stringValue
```
fredkiefer, line 121: What about objectValue?
And does an NSSwitch have a tag and what other NSControl methods need to be implemented to make this class usable?
Are there bindings?

Headers/AppKit/NSCell.h
gcasa: Removed the extra spaces.

Source/GSThemeDrawing.m
gcasa: Done

Headers/AppKit/NSCell.h
gcasa: Done

Source/GSThemeDrawing.m
```diff
@@ -816,6 +816,139 @@ - (void) drawStepperCell: (NSCell*)cell
...
+  // Draw initial path to enclose the button...
+  // left half-circle
+  p = [NSBezierPath bezierPath];
+  [p appendBezierPathWithArcWithCenter: point
+				radius: radius
+			    startAngle: 90.0
+			      endAngle: 270.0];
+
+  // line to first point and right halfcircle
+  point.x += frame.size.width - frame.size.height;
```
fredkiefer, line 856: Mixing width and height to set an x coordinate looks wrong. Most likely you want to do something like 

`point.x += frame.size.width - 2 * radius;`

But even there I am not totally sure.

Source/GSThemeDrawing.m
```diff
@@ -816,6 +816,139 @@ - (void) drawStepperCell: (NSCell*)cell
...
+			      endAngle: 315.0];
+  [[NSColor controlLightHighlightColor] set];
+  [p stroke];
+}
+
+- (void) drawSwitchKnob: (NSRect)frame
+               forState: (NSControlStateValue)value
+                enabled: (BOOL)enabled
+{
+
```
fredkiefer, line 896: Here is a left over empty line and the following comment belongs three lines below.

Source/GSThemeDrawing.m
```diff
@@ -816,6 +816,139 @@ - (void) drawStepperCell: (NSCell*)cell
...
+  // make smaller so that it does not touch frame
+  NSColor *backgroundColor = enabled ? [NSColor windowBackgroundColor] : [NSColor disabledControlTextColor];
+  NSBezierPath *oval;
+  NSRect rect = NSZeroRect;
+  CGFloat w = (frame.size.width / 2) - 2;
+  CGFloat h = frame.size.height - 5;
+  CGFloat y = frame.origin.y + 2;
+
+  [backgroundColor set];
+  if(value == NSControlStateValueOff)
```
fredkiefer, line 906: Missing space after the `if`

Source/GSThemeDrawing.m
```diff
@@ -816,6 +816,139 @@ - (void) drawStepperCell: (NSCell*)cell
...
+  
+  // fill oval with background color
+  [backgroundColor set];
+  [oval fill];
+  
+  // and stroke rounded button
+  [[NSColor shadowColor] set];
+  [oval stroke];
+  
+  // NSRectFill(frame);
```
fredkiefer, line 932: Remove commented out code.

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -2789,6 +2790,32 @@ - (id) decodeObjectForXib: (GSXibElement*)element
         [object setHeaderToolTip: [element attributeForKey: @"toolTip"]];
     }
 
+  // Handle state for NSSwitch
```
fredkiefer, line 2793: As the code for NSSwitch has gotten a lot bigger now it would be better to move it into a separate method and add a special decoder for the tag "switch".
But it could also just be that it would be sufficient to add "enabled" to the array "XmlBoolDefaultYes". I really don't understand the special handling here.

Source/GSXibLoading.m
```diff
@@ -466,6 +466,9 @@ - (NSString*) trigger
   return trigger;
 }
 
+#pragma clang diagnostic push
```
fredkiefer, line 469: I think all these pragma statements result in compiler warnings with gcc. In that case we now have three warnings instead of one. I need to test.

Source/NSMatrix.m
```diff
@@ -130,11 +130,13 @@ implementation here reflects OS42 more closely (as the original code
   NSInteger height;
 } MRect;
 
+/*
```
fredkiefer, line 133: If this code is not needed, just remove it. But I don't see the connection to NSSwitch.

Source/NSSwitch.m
```diff
@@ -0,0 +1,290 @@
...
+
+- (id) target
+{
+  return _target;
+}
+
+- (void) setEnabled: (BOOL)flag
+{
+  _enabled = flag;
+}
```
fredkiefer, line 72: Shouldn't this also call setNeedsDisplay?

Source/NSSwitch.m
gcasa: There are no bindings.

gcasa: Yes, it does have a tag.  It gets this from its parent NSView.

gcasa: I believe I have implemented everything that it needs now to be useful.

Source/GSThemeDrawing.m
gcasa: I don't think this is a concern right now.   As it is, only certain sizes can be generated.  I keep trying to think of what calculation might be best.   I will implement it at a later date if it becomes an issue.

Source/NSMatrix.m
gcasa: I know, there is no connection.  I just got really tired of seeing warnings about unused methods.

Source/NSSwitch.m
gcasa: Good point, it should.

Source/GSThemeDrawing.m
gcasa: Done

Source/GSThemeDrawing.m
gcasa: Removed.

Source/GSXib5KeyedUnarchiver.m
gcasa: It would be better to put it in a common place.  I was unaware of XmlBoolDefaultYes until now.

Source/GSThemeDrawing.m
gcasa: This causes some weirdnes. :) 
weirdness

gcasa: I think a change is needed in the knob rendering method....

This change, as it is, causes some weirdness...

weirdness

gcasa: I used your calculation (with slight modification) in the knob rendering.  looks better and it seems more correct.

Source/GSThemeTools.m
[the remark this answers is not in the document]
gcasa: Argh!

Source/GSXibLoading.m
gcasa: A high hearty "Dammit" is in order... I'll just remove the pragmas. :/

` Compiling file GSXibLoader.m ...
 Compiling file GSXibLoading.m ...
GSXibLoading.m:469:0: warning: ignoring #pragma clang diagnostic [-Wunknown-pragmas]
 #pragma clang diagnostic push
 ^
GSXibLoading.m:470:0: warning: ignoring #pragma clang diagnostic [-Wunknown-pragmas]
 #pragma clang diagnostic ignored "-Wcast-of-sel-type"
 ^
GSXibLoading.m:511:0: warning: ignoring #pragma clang diagnostic [-Wunknown-pragmas]
 #pragma clang diagnostic pop
 ^
 Compiling file GSXibKeyedUnarchiver.m ...
 Compiling file GSXib5KeyedUnarchiver.m `

Source/GSThemeTools.m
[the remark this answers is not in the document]
gcasa: Reverted... I checked out the copy from master. :)

Headers/AppKit/AppKit.h
```diff
@@ -216,6 +216,7 @@
 #import <AppKit/NSStepperTouchBarItem.h>
 #import <AppKit/NSStepper.h>
 #import <AppKit/NSStepperCell.h>
+#import <AppKit/NSSwitch.h>
```
fredkiefer, line 219: Should we include NSAccessibilityProtocols.h here as well?

Source/GSThemeDrawing.m
fredkiefer: That is weired, radius is frame.size.height / 2.0 multiplying that value with 2.0 should give  frame.size.height. This should just give the same result. Are you sure you did not mix up things?
But this is just a way to write things differently, it should result in the same appearance. So this is not important.

Source/GSThemeDrawing.m
```diff
@@ -816,6 +816,139 @@ - (void) drawStepperCell: (NSCell*)cell
...
+
+- (void) drawSwitchKnob: (NSRect)frame
+               forState: (NSControlStateValue)value
+                enabled: (BOOL)enabled
+{
+  NSColor *backgroundColor = enabled ? [NSColor windowBackgroundColor] : [NSColor disabledControlTextColor];
+  NSBezierPath *oval;
+  NSRect rect = NSZeroRect;
+  CGFloat w = (frame.size.width / 2) - 2;
+  CGFloat h = frame.size.height - 5;
```
fredkiefer, line 903: ON my computer the knob still has a few pixels outside of the bezel. With a value of 6 here it is completely inside. Of course this could also be achieved with other changes.

Source/NSSwitch.m
```diff
@@ -0,0 +1,288 @@
...
+   Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+   Boston, MA 02110 USA.
+*/
+
+#import "AppKit/NSSwitch.h"
+#import "GNUstepGUI/GSTheme.h"
+
+@implementation NSSwitch
+
++ (void) initialize
```
fredkiefer, line 30: I just compared with NSScroller, another NSControl subclass without a cell and in some regards we are already better here. But some methods are missing that would normally go to the NSCell, for example stuff like acceptsFirstResponder. And there are the takeXXXValueFrom type of methods, that are also missing in NSScroller. Should we add these?

Source/NSSwitch.m
```diff
@@ -0,0 +1,288 @@
...
+            {
+              NSString *s = [coder decodeObjectForKey: @"NSControlAction"];
+              [self setAction: NSSelectorFromString(s)];
+            }
+          if ([coder containsValueForKey: @"NSControlTarget"])
+            {
+              id t = [coder decodeObjectForKey: @"NSControlTarget"];
+              [self setTarget: t];
+            }
+         if ([coder containsValueForKey: @"NSEnabled"])
```
fredkiefer, line 248: I think NSControl already decodes that, we don't need it here.
The same is true for the encoding.

ivucica:
I have removed myself as I don't think I can add much to Fred's input. (My feedback on theme I shared previously is just a nitpick, anyway, and unrelated to existence of the class or correctness of its behavior, so it can and should be ignored at this time.)

Headers/AppKit/AppKit.h
gcasa: I believe so.

Source/GSXibLoading.m
gcasa: fixed

Source/NSSwitch.m
gcasa: See the comment I left.  The value is decoded in NSControl, but not used.

Source/GSXib5KeyedUnarchiver.m
gcasa: Done

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -2969,6 +2985,29 @@ - (id) decodeObjectForKey: (NSString *)key
   return object;
 }
 
+- (NSInteger) decodeIntegerForKey: (NSString *)key
```
fredkiefer, line 2988: Are you sure you need this method? When I took over this class I was trying to remove most of these specific methods and base them on decodeObjectForKey: as the super implementation will use that call and convert the result to an integer. I would even prefer to remove more of these specific type methods but for each of the remaining one there was some special issue.
I might be wrong here but this is how I remember things.

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -2788,7 +2804,7 @@ - (id) decodeObjectForXib: (GSXibElement*)element
       else if ([object respondsToSelector: @selector(setHeaderToolTip:)])
         [object setHeaderToolTip: [element attributeForKey: @"toolTip"]];
     }
-
+
```
fredkiefer, line 2807: Left over space.

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -381,7 +382,8 @@ + (void) initialize
                @"decodeDividerStyleForElement:", @"NSDividerStyle",
                @"decodeToolbarIdentifiedItemsForElement:", @"NSToolbarIBIdentifiedItems",
                @"decodeToolbarImageForElement:", @"NSToolbarItemImage",
-               nil];
+               @"decodeControlContentsForElement:",@"NSControlContents",
```
fredkiefer, line 385: Where does this key get used? I could not find it anywhere in the GNUstep gui code.

Source/NSSwitch.m
gcasa: I believe we don't need to.  The takeXXXValueFrom methods don't seem to exist on NSSwitch.  Also, some methods on this don't seem to have any effect (such as setAcceptsFirstResponder:) per tests.

Source/GSXib5KeyedUnarchiver.m
gcasa: I don't think it does.  I will remove it.  It was added earlier...  I was trying to understand what was going on in GSXib5KeyedUnarchiver.

gcasa: It is used in NSSwitch.   It decodes the "switch='on'" attribute.

Source/GSXib5KeyedUnarchiver.m
gcasa: Um... ok?

Source/GSXib5KeyedUnarchiver.m
gcasa: I removed it.  It doesn't seem necessary.

Source/GSThemeDrawing.m
gcasa: Changed... looks excellent.

Source/NSSwitch.m
gcasa: The test I wrote for this doesn't work... investigating.

Source/GSThemeDrawing.m
fredkiefer: Yes, but where is this change?

Source/GSXib5KeyedUnarchiver.m
fredkiefer: As I wrote I do not use 10.15 and won't be able to see that class in action myself. Could you please provide me with a key encoded file for NSSwitch? It would be best in XML format. 
The code I use for that looks something like this:

void encode(id object)
{
  NSMutableData *m;
  NSKeyedArchiver *a;
  NSString *aPath;
  
  aPath = @"/Users/fred/Documents/GNUstep/keyedcoding/keyed.xml";

  m = [[NSMutableData alloc] initWithCapacity: 10240];
  a = [[NSKeyedArchiver alloc] initForWritingWithMutableData: m];
  [a setOutputFormat: NSPropertyListXMLFormat_v1_0];
  [a encodeObject: object forKey: @"root"];
  [a finishEncoding];
  [m writeToFile: aPath atomically: YES];
  [a release];
  [m release];
}

Source/NSSwitch.m
fredkiefer: Just look through the code of NSScroller yourself and decide what we could use here.

Source/GSXib5KeyedUnarchiver.m
fredkiefer: Yes, but you did not push that change. I am not able to see it.

Source/GSXib5KeyedUnarchiver.m
fredkiefer: Also a space after the comma is missing.

gcasa: I am trying to attach it here, but it's not cooperating.   I will email them to you.

gcasa: I sent them.

Source/NSSwitch.m
gcasa: I reran it... it appears to work.  Also, in Xcode the takeXXX methods do not appear in the connection inspector (as they do for NSButton and others).

Source/GSThemeDrawing.m
gcasa: Committed.

gcasa:
Please let me know if you have any further comments, but I believe everything is covered now.

ChangeLog
```diff
@@ -1,3 +1,23 @@
...
+	* Headers/AppKit/NSSwitch.h: Header for NSSwitch
+	* MISSING: Remove NSSwitch
+	* Source/GNUmakefile: Add .h and .m files
+	* Source/GSThemeDrawing.m: Add drawing code for knob and
+        switch basin.
+	* Source/GSXib5KeyedUnarchiver.m: Add decoding for
+        NSControlContents.  Make sure enabled is YES when not
+        present.
+	* Source/NSColor.m: Minor changes
+	* Source/NSMatrix.m: Remove uneeded code
```
fredkiefer, line 17: I think it is "unneeded" but you are the native speaker.
This occurs again two lines below.

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -53,6 +53,7 @@
 #import "AppKit/NSScrollView.h"
 #import "AppKit/NSSliderCell.h"
 #import "AppKit/NSSplitView.h"
+#import "AppKit/NSSwitch.h"
```
fredkiefer, line 56: This import is no longer needed.

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -2725,6 +2728,19 @@ - (id) decodeToolbarImageForElement: (GSXibElement*)element
   return [self findResourceWithName: name];
 }
 
+- (id) decodeControlContentsForElement: (GSXibElement *)element
+{
+  NSNumber *num = [NSNumber numberWithInteger: 0];
```
fredkiefer, line 2733: We will most likely need to implement variants here for other NSControl subclasses. But for now this should do.

Source/NSSwitch.m
```diff
@@ -0,0 +1,291 @@
...
+}
+
+// NSCoding
+- (id) initWithCoder: (NSCoder *)coder
+{
+  if ((self = [super initWithCoder: coder]) != nil)
+    {
+      if ([coder allowsKeyedCoding])
+        {
+          if ([coder containsValueForKey: @"NSControlContents"])
```
fredkiefer, line 234: My feeling is that those should be moved to NSControl but for now it is OK. It looks like the Cocoa classes have changed their keyed encoding over the years.

Source/NSSwitch.m
fredkiefer: We might have to add a few more methods later on, when people start to use this class.

PR #62, opened 2020-05-05
NSPathControl branch

gcasa:
Implement NSPathControl.  NSPathControl shows a file path and allows the user to select parts of the path name.    This change is almost completed. GC

### DIFF
--- a/Headers/AppKit/NSPathCell.h
+++ b/Headers/AppKit/NSPathCell.h
@@ -48,7 +48,6 @@ typedef NSUInteger NSPathStyle;
 {
   NSPathStyle _pathStyle;
   NSColor *_backgroundColor;
-  NSArray *_pathItems;
   NSAttributedString *_placeholderAttributedString;
   NSArray *_allowedTypes;
   id<NSPathCellDelegate> _delegate;
--- a/Headers/AppKit/NSPathCell.h
+++ b/Headers/AppKit/NSPathCell.h
@@ -72,9 +72,6 @@ typedef NSUInteger NSPathStyle;
 - (NSPathStyle) pathStyle;
 - (void) setPathStyle: (NSPathStyle)pathStyle;
 
-// - (void) setControlSize: (NSControlSize)size;
-// - (void) setObjectValue: (id)obj;
-
 - (NSAttributedString *) placeholderAttributedString;
 - (void) setPlaceholderAttributedString: (NSAttributedString *)string;
 
--- a/Headers/AppKit/NSPathCell.h
+++ b/Headers/AppKit/NSPathCell.h
@@ -33,8 +33,6 @@
 extern "C" {
 #endif
 
-Class pathComponentCellClass;
-  
 enum {
   NSPathStyleStandard,
   NSPathStyleNavigationBar,  // deprecated
@@ -59,8 +57,8 @@ typedef NSUInteger NSPathStyle;
   SEL _doubleAction;
   NSArray *_pathComponentCells;
   NSPathComponentCell *_clickedPathComponentCell;
-  id _objectValue;
-  NSControlSize _controlSize;
+  // id _objectValue;
+  // NSControlSize _controlSize;
 }
   
 - (void)mouseEntered:(NSEvent *)event 
@@ -77,8 +75,8 @@ typedef NSUInteger NSPathStyle;
 - (NSPathStyle) pathStyle;
 - (void) setPathStyle: (NSPathStyle)pathStyle;
 
-- (void) setControlSize: (NSControlSize)size;
-- (void) setObjectValue: (id)obj;
+// - (void) setControlSize: (NSControlSize)size;
+// - (void) setObjectValue: (id)obj;
 
 - (NSAttributedString *) placeholderAttributedString;
 - (void) setPlaceholderAttributedString: (NSAttributedString *)string;
--- a/Headers/AppKit/NSPathCell.h
+++ b/Headers/AppKit/NSPathCell.h
@@ -55,6 +55,7 @@ typedef NSUInteger NSPathStyle;
   SEL _doubleAction;
   NSArray *_pathComponentCells;
   NSPathComponentCell *_clickedPathComponentCell;
+  id _objectValue;
 }
   
 - (void)mouseEntered:(NSEvent *)event 
--- a/Headers/AppKit/NSPathCell.h
+++ b/Headers/AppKit/NSPathCell.h
@@ -49,7 +49,6 @@ typedef NSUInteger NSPathStyle;
   NSPathStyle _pathStyle;
   NSColor *_backgroundColor;
   NSArray *_pathItems;
-  NSString *_placeholderString;
   NSAttributedString *_placeholderAttributedString;
   NSArray *_allowedTypes;
   id<NSPathCellDelegate> _delegate;
@@ -57,8 +56,6 @@ typedef NSUInteger NSPathStyle;
   SEL _doubleAction;
   NSArray *_pathComponentCells;
   NSPathComponentCell *_clickedPathComponentCell;
-  // id _objectValue;
-  // NSControlSize _controlSize;
 }
   
 - (void)mouseEntered:(NSEvent *)event 
--- a/Headers/AppKit/NSPathControlItem.h
+++ b/Headers/AppKit/NSPathControlItem.h
@@ -33,9 +33,27 @@
 extern "C" {
 #endif
 
+  @class NSURL, NSAttributedString, NSImage, NSString;
 
 @interface NSPathControlItem : NSObject
+{
+  NSURL *_url;
+  NSAttributedString *_attributedTitle;
+  NSImage *_image;
+  NSString *_title;
+}
+  
+- (NSURL *) URL;
+- (void) setURL: (NSURL *)url;
 
+- (NSAttributedString *) attributedTitle;
+- (void) setAttributedTitle: (NSAttributedString *)attributedTitle;
+
+- (NSImage *) image;
+- (void) setImage: (NSImage *)image;
+
+- (NSString *) title;
+- (void) setTitle: (NSString *)title;
 
 @end
 
--- a/Headers/AppKit/NSPathControlItem.h
+++ b/Headers/AppKit/NSPathControlItem.h
@@ -40,7 +40,6 @@ extern "C" {
   NSURL *_url;
   NSAttributedString *_attributedTitle;
   NSImage *_image;
-  NSString *_title;
 }
   
 - (NSURL *) URL;
--- a/Headers/AppKit/NSPathControlItem.h
+++ b/Headers/AppKit/NSPathControlItem.h
@@ -33,7 +33,7 @@
 extern "C" {
 #endif
 
-  @class NSURL, NSAttributedString, NSImage, NSString;
+@class NSURL, NSAttributedString, NSImage, NSString;
 
 @interface NSPathControlItem : NSObject
 {
@@ -64,4 +64,3 @@ extern "C" {
 #endif	/* GS_API_MACOSX */
 
 #endif	/* _NSPathControlItem_h_GNUSTEP_GUI_INCLUDE */
-
--- a/Source/GSThemeDrawing.m
+++ b/Source/GSThemeDrawing.m
@@ -969,6 +969,7 @@
   
   if (s == NSPathStyleStandard || s == NSPathStyleNavigationBar)
     {
+      /*
       if (s == NSPathStyleNavigationBar)
         {
           [[NSColor lightGrayColor] set];
@@ -976,7 +977,7 @@
           [[NSColor controlShadowColor] set];
           NSFrameRectWithWidth(f, 1);
         }
-      
+      */
       // Modify positions...
       textFrame.origin.x += 25.0; // the width of the image plus a few pixels.
       textFrame.origin.y += 3.0; // center with the image...
--- a/Source/GSThemeDrawing.m
+++ b/Source/GSThemeDrawing.m
@@ -985,6 +985,8 @@
   // Modify positions...
   imgFrame.size.width = 17.0;
   imgFrame.size.height = 17.0;
+  imgFrame.origin.x += 2.0;
+  imgFrame.origin.y += 2.0;
   textFrame.origin.x += imgFrame.size.width + 5.0; // the width of the image plus a few pixels.
   textFrame.origin.y += 5.0; // center with the image...
   arrowFrame.origin.x += newFrame.size.width - 17.0; 
--- a/Source/GSThemeDrawing.m
+++ b/Source/GSThemeDrawing.m
@@ -954,19 +954,19 @@
 // NSPathComponentCell
 
 - (void) drawPathComponentCellWithFrame: (NSRect)f
-                                 inView: (NSView *)v
-                                withURL: (NSURL *)u
-                                  image: (NSImage *)i
+                                 inView: (NSPathControl *)pc
+                               withCell: (NSPathComponentCell *)cell
                         isLastComponent: (BOOL)lc
 {
+  NSImage *i = [cell image];
+  NSURL *u = [cell URL];
   NSString *string = [[u path] lastPathComponent];
   NSRect textFrame = f;
   NSRect imgFrame = f;
   NSRect arrowFrame = f;
   NSImage *arrowImage = [NSImage imageNamed: @"NSMenuArrow"];
-  NSPathControl *pc = (NSPathControl *)v;
   NSPathStyle s = [pc pathStyle];
-
+  
   if (s == NSPathStyleStandard || s == NSPathStyleNavigationBar)
     {
       if (s == NSPathStyleNavigationBar)
@@ -1005,7 +1005,7 @@
     {
       if (lc == YES)
         {
-          NSRect newFrame = [v frame];
+          NSRect newFrame = [pc frame];
           
           arrowImage = [NSImage imageNamed: @"common_ArrowDown"];
                     
--- a/Source/GSThemeDrawing.m
+++ b/Source/GSThemeDrawing.m
@@ -953,43 +953,49 @@
 
 // NSPathComponentCell
 
-- (void) drawPathComponentCellWithFrame: (NSRect)f
+- (void) drawPathComponentCellWithFrame: (NSRect)frame
                                  inView: (NSPathControl *)pc
                                withCell: (NSPathComponentCell *)cell
-                        isLastComponent: (BOOL)lc
+                        isLastComponent: (BOOL)last
 {
-  NSImage *i = [cell image];
-  NSURL *u = [cell URL];
-  NSString *string = [[u path] lastPathComponent];
-  NSRect textFrame = f;
-  NSRect imgFrame = f;
-  NSRect arrowFrame = f;
+  NSImage *img = [cell image];
+  NSURL *url = [cell URL];
+  NSString *string = [[url path] lastPathComponent];
+  NSRect textFrame = frame;
+  NSRect imgFrame = frame;
+  NSRect arrowFrame = frame;
   NSImage *arrowImage = [NSImage imageNamed: @"NSMenuArrow"];
-  NSPathStyle s = [pc pathStyle];
-  
-  if (s == NSPathStyleStandard || s == NSPathStyleNavigationBar)
+  NSPathStyle style= [pc pathStyle];
+  NSRect newFrame = frame;
+
+  if (style == NSPathStylePopUp)
     {
-      /*
-      if (s == NSPathStyleNavigationBar)
-        {
-          [[NSColor lightGrayColor] set];
-          NSRectFill(f);
-          [[NSColor controlShadowColor] set];
-          NSFrameRectWithWidth(f, 1);
-        }
-      */
-      // Modify positions...
-      textFrame.origin.x += 25.0; // the width of the image plus a few pixels.
-      textFrame.origin.y += 3.0; // center with the image...
-      imgFrame.size.width = 20.0;
-      imgFrame.size.height = 20.0;
-      arrowFrame.origin.x += f.size.width - 20.0; 
-      arrowFrame.size.width = 8.0;
-      arrowFrame.size.height = 8.0;
-      arrowFrame.origin.y += 5.0;
+      newFrame = [pc frame];
       
+      // Reset coodinates.
+      newFrame.origin.x = 0.0;
+      newFrame.origin.y = 0.0;
+      
+      // Use control frame...
+      textFrame = newFrame;
+      imgFrame = newFrame;
+      arrowFrame = newFrame;    
+    }
+  
+  // Modify positions...
+  imgFrame.size.width = 17.0;
+  imgFrame.size.height = 17.0;
+  textFrame.origin.x += imgFrame.size.width + 5.0; // the width of the image plus a few pixels.
+  textFrame.origin.y += 5.0; // center with the image...
+  arrowFrame.origin.x += newFrame.size.width - 17.0; 
+  arrowFrame.size.width = 8.0;
+  arrowFrame.size.height = 8.0;
+  arrowFrame.origin.y += 5.0;
+      
+  if (style== NSPathStyleStandard || style== NSPathStyleNavigationBar)
+    {
       // Draw the image...
-      [i drawInRect: imgFrame];
+      [img drawInRect: imgFrame];
       
       // Draw the text...
       [[NSColor textColor] set]; 
@@ -997,42 +1003,23 @@
            withAttributes: nil];
       
       // Draw the arrow...
-      if (lc == NO)
+      if (last == NO)
         {
           [arrowImage drawInRect: arrowFrame];
         }
     }
-  else // can only be PopUp...
+  else if (style == NSPathStylePopUp)
     {
-      if (lc == YES)
+      if (last == YES)
         {
-          NSRect newFrame = [pc frame];
-          
           arrowImage = [NSImage imageNamed: @"common_ArrowDown"];
-                    
-          // Reset coodinates.
-          newFrame.origin.x = 0.0;
-          newFrame.origin.y = 0.0;
-          [[NSColor controlShadowColor] set];
-          NSFrameRectWithWidth(newFrame, .5);
-          
-          // Use control frame...
-          textFrame = newFrame;
-          imgFrame = newFrame;
-          arrowFrame = newFrame;
-          
-          // Modify positions...
-          textFrame.origin.x += 25.0; // the width of the image plus a few pixels.
-          textFrame.origin.y += 3.0; // center with the image...
-          imgFrame.size.width = 20.0;
-          imgFrame.size.height = 20.0;
-          arrowFrame.origin.x += newFrame.size.width - 20.0; 
-          arrowFrame.size.width = 8.0;
-          arrowFrame.size.height = 8.0;
-          arrowFrame.origin.y += 5.0;
 
+          // Draw border...
+          [[NSColor controlShadowColor] set];
+          NSFrameRectWithWidth(newFrame, 1.0);
+          
           // Draw the image...
-          [i drawInRect: imgFrame];
+          [img drawInRect: imgFrame];
           
           // Draw the text...
           [[NSColor textColor] set]; 
--- a/Source/GSThemeDrawing.m
+++ b/Source/GSThemeDrawing.m
@@ -59,7 +59,9 @@
 #import "AppKit/NSTabViewItem.h"
 #import "AppKit/PSOperators.h"
 #import "AppKit/NSSliderCell.h"
-
+#import "AppKit/NSPathCell.h"
+#import "AppKit/NSPathControl.h"
+#import "AppKit/NSPathComponentCell.h"
 #import "GNUstepGUI/GSToolbarView.h"
 #import "GNUstepGUI/GSTitleView.h"
 
@@ -947,6 +949,53 @@
               forState: state
                enabled: enabled];
         
+}
+
+// NSPathComponentCell
+
+- (void) drawPathComponentCellWithFrame: (NSRect)f
+                                 inView: (NSView *)v
+                                withURL: (NSURL *)u
+                                  image: (NSImage *)i
+                        isLastComponent: (BOOL)lc
+{
+  NSString *string = [[u path] lastPathComponent];
+  NSRect textFrame = f;
+  NSRect imgFrame = f;
+  NSRect arrowFrame = f;
+  NSImage *arrowImage = [NSImage imageNamed: @"NSMenuArrow"];
+  NSPathControl *pc = (NSPathControl *)v;
+  NSPathStyle s = [pc pathStyle];
+
+  if (s == NSPathStyleStandard || s == NSPathStyleNavigationBar)
+    {
+      // Modify positions...
+      textFrame.origin.x += 25.0; // the width of the image plus a few pixels.
+      textFrame.origin.y += 3.0; // center with the image...
+      imgFrame.size.width = 20.0;
+      imgFrame.size.height = 20.0;
+      arrowFrame.origin.x += f.size.width - 20.0; 
+      arrowFrame.size.width = 8.0;
+      arrowFrame.size.height = 8.0;
+      arrowFrame.origin.y += 5.0;
+      
+      // Draw the image...
+      [i drawInRect: imgFrame];
+      
+      // Draw the text...
+      [[NSColor textColor] set]; 
+      [string drawAtPoint: textFrame.origin
+           withAttributes: nil];
+      
+      // Draw the arrow...
+      if (lc == NO)
+        {
+          [arrowImage drawInRect: arrowFrame];
+        }
+    }
+  else
+    {
+    }
 }
 
 // NSSegmentedControl drawing methods
--- a/Source/GSThemeDrawing.m
+++ b/Source/GSThemeDrawing.m
@@ -1013,7 +1013,7 @@
           newFrame.origin.x = 0.0;
           newFrame.origin.y = 0.0;
           [[NSColor controlShadowColor] set];
-          NSFrameRectWithWidth(newFrame, 1);
+          NSFrameRectWithWidth(newFrame, .5);
           
           // Use control frame...
           textFrame = newFrame;
--- a/Source/GSThemeDrawing.m
+++ b/Source/GSThemeDrawing.m
@@ -969,6 +969,14 @@
 
   if (s == NSPathStyleStandard || s == NSPathStyleNavigationBar)
     {
+      if (s == NSPathStyleNavigationBar)
+        {
+          [[NSColor lightGrayColor] set];
+          NSRectFill(f);
+          [[NSColor controlShadowColor] set];
+          NSFrameRectWithWidth(f, 1);
+        }
+      
       // Modify positions...
       textFrame.origin.x += 25.0; // the width of the image plus a few pixels.
       textFrame.origin.y += 3.0; // center with the image...
@@ -993,8 +1001,46 @@
           [arrowImage drawInRect: arrowFrame];
         }
     }
-  else
+  else // can only be PopUp...
     {
+      if (lc == YES)
+        {
+          NSRect newFrame = [v frame];
+          
+          arrowImage = [NSImage imageNamed: @"common_ArrowDown"];
+                    
+          // Reset coodinates.
+          newFrame.origin.x = 0.0;
+          newFrame.origin.y = 0.0;
+          [[NSColor controlShadowColor] set];
+          NSFrameRectWithWidth(newFrame, 1);
+          
+          // Use control frame...
+          textFrame = newFrame;
+          imgFrame = newFrame;
+          arrowFrame = newFrame;
+          
+          // Modify positions...
+          textFrame.origin.x += 25.0; // the width of the image plus a few pixels.
+          textFrame.origin.y += 3.0; // center with the image...
+          imgFrame.size.width = 20.0;
+          imgFrame.size.height = 20.0;
+          arrowFrame.origin.x += newFrame.size.width - 20.0; 
+          arrowFrame.size.width = 8.0;
+          arrowFrame.size.height = 8.0;
+          arrowFrame.origin.y += 5.0;
+
+          // Draw the image...
+          [i drawInRect: imgFrame];
+          
+          // Draw the text...
+          [[NSColor textColor] set]; 
+          [string drawAtPoint: textFrame.origin
+               withAttributes: nil];
+          
+          // Draw the arrow...
+          [arrowImage drawInRect: arrowFrame];
+        }
     }
 }
 
--- a/Source/GSXib5KeyedUnarchiver.m
+++ b/Source/GSXib5KeyedUnarchiver.m
@@ -3236,10 +3236,6 @@ didStartElement: (NSString*)elementName
           hasValue  = [currentElement attributeForKey: @"alternateTitle"] != nil;
           hasValue |= [currentElement attributeForKey: @"alternateImage"] != nil;
         }
-      else if ([@"NSPathStyle" isEqualToString: key])
-        {
-          hasValue = [currentElement attributeForKey: @"pathStyle"] != nil;
-        }
       else if ([@"NSPathComponentCells" isEqualToString: key])
         {
           hasValue = [currentElement elementForKey: @"url"] != nil;
--- a/Source/GSXib5KeyedUnarchiver.m
+++ b/Source/GSXib5KeyedUnarchiver.m
@@ -2201,6 +2201,11 @@ didStartElement: (NSString*)elementName
     {
       object = [element attributeForKey: @"stringValue"];
     }
+  else if ([class isSubclassOfClass: [NSPathCell class]])
+    {
+      GSXibElement *el = [element elementForKey: @"url"];
+      object = [NSURL URLWithString: [el attributeForKey: @"string"]];
+    }
   else
     {
       // Try the title attribute first as it is the more common encoding...
@@ -3226,6 +3231,7 @@ didStartElement: (NSString*)elementName
         {
           hasValue  = [currentElement attributeForKey: @"title"] != nil;
           hasValue |= [currentElement attributeForKey: @"image"] != nil;
+          hasValue |= [currentElement attributeForKey: @"string"] != nil;
         }
       else if ([@"NSControlContents" isEqualToString: key])
         {
--- a/Source/GSXib5KeyedUnarchiver.m
+++ b/Source/GSXib5KeyedUnarchiver.m
@@ -1431,6 +1431,8 @@ didStartElement: (NSString*)elementName
         size = [NSFont labelFontSize];
       else if ([metaFont containsString: @"system"])
         size = [NSFont systemFontSize];
+      else if ([metaFont containsString: @"toolTip"])
+        size = [NSFont smallSystemFontSize];
       else if (metaFont)
         NSWarnMLog(@"unknown meta font value: %@", metaFont);
     }
--- a/Source/GSXib5KeyedUnarchiver.m
+++ b/Source/GSXib5KeyedUnarchiver.m
@@ -48,6 +48,7 @@
 #import "AppKit/NSNib.h"
 #import "AppKit/NSParagraphStyle.h"
 #import "AppKit/NSPathCell.h"
+#import "AppKit/NSPathComponentCell.h"
 #import "AppKit/NSPopUpButton.h"
 #import "AppKit/NSPopUpButtonCell.h"
 #import "AppKit/NSScroller.h"
@@ -60,6 +61,7 @@
 #import "AppKit/NSTabView.h"
 #import "AppKit/NSToolbarItem.h"
 #import "AppKit/NSView.h"
+#import "AppKit/NSWorkspace.h"
 #import "GSCodingFlags.h"
 
 #define DEBUG_XIB5 0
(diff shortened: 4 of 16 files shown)


gcasa:
There is still some work to do.   I just wanted to get this out there

gcasa:
To do:
- Add logic to get cell by frame or point
- Clean up implementation
- Clean up formatting

gcasa:
Here is an example of the keyed encoding... I have annotated it with positions so that it is more readable....
[control3.txt](https://github.com/gnustep/libs-gui/files/4580647/control3.txt)

Source/NSMenu.m
```diff
@@ -1626,6 +1626,22 @@ + (void) popUpContextMenu: (NSMenu *)menu
   [menu _rightMouseDisplay: event];
 }
 
+- (void) popUpMenuPositionItem: (NSMenuItem *)item
+                    atLocation: (NSPoint) point
+                        inView: (NSView *) view
+{
+  NSRect cellFrame = [[view window] convertRectToScreen: [view frame]];
```
fredkiefer, line 1633: This is only correct if the views frame is in window coordinates. You should first transform to the base view
cellFrame = [view convertRect: [view frame] toView: nil];

Source/NSMenu.m
```diff
@@ -1626,6 +1626,22 @@ + (void) popUpContextMenu: (NSMenu *)menu
   [menu _rightMouseDisplay: event];
 }
 
+- (void) popUpMenuPositionItem: (NSMenuItem *)item
+                    atLocation: (NSPoint) point
```
fredkiefer, line 1630: We don't seem to use the location here.

Source/NSPathCell.m
```diff
@@ -0,0 +1,309 @@
...
+}
+
+- (void) setPathStyle: (NSPathStyle)pathStyle
+{
+  _pathStyle = pathStyle;
+}
+
+- (void) setControlSize: (NSControlSize)size
+{
+  _controlSize = size;
```
fredkiefer, line 74: As I wrote it is better to reuse the super implementation here.

Source/NSPathCell.m
```diff
@@ -0,0 +1,309 @@
...
+}
+
+- (void) setBackgroundColor: (NSColor *)color
+{
+  ASSIGNCOPY(_backgroundColor, color);
+}
+
++ (Class) pathComponentCellClass
+{
+  return pathComponentCellClass;
```
fredkiefer, line 114: You need to initialise this variable somewhere.

Source/NSPathCell.m
```diff
@@ -0,0 +1,309 @@
...
++ (void) setPathComponentCellClass: (Class)clz
+{
+  pathComponentCellClass = clz;
+}
+
+- (NSRect)rectOfPathComponentCell:(NSPathComponentCell *)cell 
+                        withFrame:(NSRect)frame 
+                           inView:(NSView *)view
+{
+  return NSZeroRect;
```
fredkiefer, line 126: This method should be used in the drawing code.

Source/NSPathCell.m
```diff
@@ -0,0 +1,309 @@
...
+- (NSPathComponentCell *)pathComponentCellAtPoint:(NSPoint)point 
+                                        withFrame:(NSRect)frame 
+                                           inView:(NSView *)view
+{
+  return nil;
+}
+
+- (NSPathComponentCell *) clickedPathComponentCell
+{
+  return nil;
```
fredkiefer, line 138: You already have an instance variable for this _clickedPathComponentCell.

Source/NSPathCell.m
```diff
@@ -0,0 +1,309 @@
...
+}
+
+- (void) setControlSize: (NSControlSize)size
+{
+  _controlSize = size;
+}
+
+- (void) setObjectValue: (id)obj
+{
+  ASSIGN(_objectValue, obj);
```
fredkiefer, line 79: We already know that this is the url, so we should call setURL: here. Doing so will resolve the XIB decoding issue.

Source/NSPathCell.m
```diff
@@ -0,0 +1,309 @@
...
+}
+
+- (void) setDoubleAction: (SEL)action
+{
+  _doubleAction = action;
+}
+
+- (NSURL *) URL
+{
+  return _url;
```
fredkiefer, line 163: We rather should store this value in the super objectValue instance variable.

Source/NSPathCell.m
```diff
@@ -0,0 +1,309 @@
...
+  if (count > 0)
+    {
+      NSEnumerator *en = [_pathComponentCells objectEnumerator];
+      NSPathComponentCell *cell = nil;
+      CGFloat cell_width = (frame.size.width / (CGFloat)[_pathComponentCells count]);
+      CGFloat current_x = 0.0;
+      
+      while((cell = (NSPathComponentCell *)[en nextObject]) != nil)
+        {
+          NSRect f = NSMakeRect(current_x, 0.0, cell_width, frame.size.height);
```
fredkiefer, line 197: Use rectOfPathComponentCell:... here

Source/NSPathCell.m
```diff
@@ -0,0 +1,309 @@
...
+    }
+}
+
+- (id) initWithCoder: (NSCoder *)coder
+{
+  if ([coder allowsKeyedCoding])
+    {
+      [self setPathStyle: NSPathStyleStandard];
+      
+      // if ([coder containsValueForKey: @"NSPathStyle"])
```
fredkiefer: Why is this line commented out?

Source/NSPathCell.m
```diff
@@ -0,0 +1,309 @@
...
+    {
+      NSString *string = nil;
+      BOOL isDir = NO;
+      BOOL at_root = NO;
+      NSFileManager *fm = [NSFileManager defaultManager];
+            
+      // Decompose string...
+      do
+        {
+          NSPathComponentCell *cell = [[NSPathComponentCell alloc] init];
```
fredkiefer, line 269: The cell never gets released.

Source/NSPathCell.m
```diff
@@ -0,0 +1,309 @@
...
+
+// Private...
++ (NSArray *) _generateCellsForURL: (NSURL *)url
+{
+  NSMutableArray *array = [NSMutableArray arrayWithCapacity: 10];
+    
+  // Create cells
+  if (url != nil)
+    {
+      NSString *string = nil;
```
fredkiefer: I see no useful assignment to string. This will stay nil.
Also a better name would surely help to make clear what is going on here.

Source/NSPathControl.m
```diff
@@ -0,0 +1,396 @@
...
+#import "AppKit/NSOpenPanel.h"
+
+@implementation NSPathControl
+
++ (void) initialize
+{
+  if (self == [NSPathControl class])
+    {
+      [self setVersion: 1.0];
+      [self setCellClass: [NSPathCell class]];
```
fredkiefer, line 40: With the current implementation of NSControl you will need to override the setCellClass: method locally. You may consider this a bug in GNUstep gui. Or see it as a convention that you need to follow.
I am really not sure how Cocoa implements that. You would need to write a subclass of NSControl and test the behaviour.

Source/NSPathControl.m
```diff
@@ -0,0 +1,396 @@
...
+    }
+}
+
+- (instancetype) init
+{
+  self = [super init];
+  if (self != nil)
+    {
+      [self setPathStyle: NSPathStyleStandard];
+      [self setPathComponentCells: nil];
```
fredkiefer, line 50: Setting to nil is not needed, that is already the default.

Source/NSPathControl.m
```diff
@@ -0,0 +1,396 @@
...
+  [_cell setAllowedTypes: allowedTypes];
+  [self registerForDraggedTypes: allowedTypes];
+}
+
+- (NSPathControlItem *) clickedPathItem
+{
+  return nil;
+}
+
+- (NSArray *) pathItems
```
fredkiefer, line 165: Shouldn't this come from the cell?

Source/NSPathControl.m
```diff
@@ -0,0 +1,396 @@
...
+    {
+      _localMask = mask;
+    }
+  else
+    {
+      _remoteMask = mask;
+    }
+}
+
+- (NSMenu *) menu
```
fredkiefer, line 139: If we just call super these two methods could be removed.

Source/NSPathControl.m
```diff
@@ -0,0 +1,396 @@
...
+
+- (void) setPlaceholderString: (NSString *)string
+{
+  [_cell setPlaceholderString: string];
+  [self setNeedsDisplay];
+}
+
+- (NSColor *) backgroundColor
+{
+  return _backgroundColor;
```
fredkiefer, line 200: Again should come from the cell.

Source/NSPathControl.m
```diff
@@ -0,0 +1,396 @@
...
+- (void) _doChooseMenuAction: (id)sender
+{
+  NSOpenPanel *op = [NSOpenPanel openPanel];
+  int result = 0;
+  
+  [op setAllowsMultipleSelection: NO];
+  [op setCanChooseFiles: YES];
+  [op setCanChooseDirectories: YES];
+  
+  result = [op runModalForDirectory: nil
```
fredkiefer, line 249: Shouldn't we use the current URL here?

Source/NSPathControl.m
```diff
@@ -0,0 +1,396 @@
...
+                               file: nil
+                              types: nil];
+  if (result == NSOKButton)
+    {
+      NSArray *urls = [op URLs];
+      NSURL *url = [urls objectAtIndex: 0];
+      [self setURL: url];
+    }
+
+  [[sender menu] close];
```
fredkiefer, line 259: Is this line needed?

Source/NSPathControl.m
```diff
@@ -0,0 +1,396 @@
...
+    {
+      [super mouseDown: event];
+      return;
+    }
+
+  if ([self pathStyle] == NSPathStylePopUp)
+    {
+      NSPathCell *acell = (NSPathCell *)[self cell];
+      NSArray *array = [acell pathComponentCells];
+      NSMenu *menu = [[NSMenu alloc] initWithTitle: @"Select File"];
```
fredkiefer, line 274: The menu is never released.

Source/NSPathControl.m
```diff
@@ -0,0 +1,396 @@
...
+      NSPathComponentCell *c = nil;
+      NSEnumerator *en = [array objectEnumerator];
+      
+      while((c = [en nextObject]) != nil)
+        {
+          NSURL *u = [c URL];
+          NSString *s = [[u path] lastPathComponent];
+          NSMenuItem *i = [[NSMenuItem alloc] init];
```
fredkiefer, line 282: The menu item is never released. Using one letter variable names is very helpful.

Source/NSPathControl.m
```diff
@@ -0,0 +1,396 @@
...
+}
+
+- (id) target
+{
+  return _target;
+}
+
+- (void) _doMenuAction: (id)sender
+{
+  if (_action)
```
fredkiefer, line 231: Is this the correct behaviour? I would have expected that we need to look at the menu item that causes this click.

Source/NSPathControl.m
```diff
@@ -0,0 +1,396 @@
...
+      while((c = [en nextObject]) != nil)
+        {
+          NSURL *u = [c URL];
+          NSString *s = [[u path] lastPathComponent];
+          NSMenuItem *i = [[NSMenuItem alloc] init];
+
+          [i setTitle: s];
+          [i setTarget: self];
+          [i setAction: @selector(_doMenuAction:)];
+
```
fredkiefer, line 287: Having a represented item on the menu item might help.

Source/NSPathControl.m
```diff
@@ -0,0 +1,396 @@
...
+          [menu insertItem: i
+                   atIndex: 0]; 
+        }
+
+      // Add separator
+      [menu insertItem: [NSMenuItem separatorItem]
+               atIndex: 0];
+
+      // Add choose menu option
+      NSMenuItem *i = [[NSMenuItem alloc] init];
```
fredkiefer, line 297: Again never released.

Source/NSPathControl.m
```diff
@@ -0,0 +1,396 @@
...
+        {
+          [self sendAction: _action
+                        to: _target];
+        }
+    }
+}
+
+- (NSDragOperation) draggingEntered: (id<NSDraggingInfo>)sender
+{
+  // if (_delegate != nil)
```
fredkiefer, line 329: Why is this line commented out?

Source/NSPathControl.m
```diff
@@ -0,0 +1,396 @@
...
+              NSArray *files = [pb propertyListForType: NSFilenamesPboardType];
+              if ([files count] > 0)
+                {
+                  NSString *file = [files objectAtIndex: 0];
+                  NSURL *u = [NSURL URLWithString: file];
+                  BOOL accept = [_delegate pathControl: self
+                                            acceptDrop: sender];
+                  if (accept)
+                    {
+                      [self setURL: u];
```
fredkiefer, line 347: This should only be called in performDragOperation:
Please have a look at the implementation in NSImageView.

gcasa:
Just FYI, the test is here...  my convention is to always create a test with the ${CLASSNAME}_test

https://github.com/gcasa/NSPathControl_test

gcasa:
I will start addressing each of the concerns once I "finish" :)

Source/NSPathControl.m
gcasa: No, it actually returns the associated cell.  If you look at the test...

2020-05-05 16:01:02.469748-0400 NSPathControl_test[17687:4249934] clickedCell on dropdown <NSPathComponentCell: 0x600003300500> 'Macintosh SSD' _minWidth: 29.000000, _fullWidth: 104.000000, _resizedWidth: 0.000000, _currentWidth: 0.000000

It looks like it returns the cell.

Headers/AppKit/NSPathCell.h
```diff
@@ -0,0 +1,136 @@
...
+
+@class NSEvent, NSView, NSArray, NSString, NSAttributeString, NSColor, NSPathComponentCell, NSOpenPanel, NSURL;    
+  
+@interface NSPathCell : NSActionCell
+{
+  NSPathStyle _pathStyle;
+  NSColor *_backgroundColor;
+  NSArray *_pathItems;
+  NSString *_placeholderString;
+  NSAttributedString *_placeholderAttributedString;
```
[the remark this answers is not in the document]
gcasa: done

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -3176,6 +3236,14 @@ - (BOOL) containsValueForKey: (NSString *)key
           hasValue  = [currentElement attributeForKey: @"alternateTitle"] != nil;
           hasValue |= [currentElement attributeForKey: @"alternateImage"] != nil;
         }
+      else if ([@"NSPathStyle" isEqualToString: key])
```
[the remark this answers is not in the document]
gcasa: I'm not sure what you mean.

Source/NSPathCell.m
gcasa: This is implemented now.

Headers/AppKit/NSPathCell.h
```diff
@@ -0,0 +1,136 @@
...
+
+#import <AppKit/NSActionCell.h>
+
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_5, GS_API_LATEST)
+
+#if	defined(__cplusplus)
+extern "C" {
+#endif
+
+Class pathComponentCellClass;
```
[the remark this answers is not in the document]
fredkiefer: I prefer you leave comments visible until you actually commit a solution.

Headers/AppKit/NSPathCell.h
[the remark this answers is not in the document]
fredkiefer: Yes, but the instance variable _placeholderString  ist still there.

Source/GSXib5KeyedUnarchiver.m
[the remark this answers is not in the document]
fredkiefer: The check in line 3212 (// Try a key minus a (potential) NS prefix...) should take care of this simple case.

Source/NSPathControl.m
```diff
@@ -0,0 +1,562 @@
...
+  if (_action)
+    {
+      [self sendAction: _action
+                    to: _target];
+    }
+  
+  // Tested on OSX it doesn't do this...  it apparently only chooses
+  // the cell, and sends the action.  It doesn't reset the URL.
+  [[sender menu] close];
+  AUTORELEASE([sender menu]);
```
fredkiefer, line 305: Why not autorelease the menu on creation? That would make the code clearer.

Source/NSPathControl.m
```diff
@@ -0,0 +1,562 @@
...
+                              types: nil];
+  if (result == NSOKButton)
+    {
+      NSArray *urls = [op URLs];
+      NSURL *url = [urls objectAtIndex: 0];
+      [self setURL: url];
+    }
+
+  [[sender menu] close];
+  AUTORELEASE([sender menu]);
```
fredkiefer, line 340: Again, this autorelease belongs into the menu creation

Source/NSPathControl.m
```diff
@@ -0,0 +1,562 @@
...
+               atIndex: 0];
+
+      // Add choose menu option
+      NSMenuItem *i = [[NSMenuItem alloc] init];
+      [i setTitle: @"Choose..."];
+      [i setTarget: self];
+      [i setAction: @selector(_doChooseMenuAction:)];
+      [menu insertItem: i
+               atIndex: 0];
+      AUTORELEASE(i);
```
fredkiefer, line 385: Why do you call AUTORELEASE instead of RELEASE?

Source/NSPathControlItem.m
```diff
@@ -0,0 +1,76 @@
...
+
+- (NSString *) title
+{
+  return [_attributedTitle string];
+}
+
+- (void) setTitle: (NSString *)title
+{
+  NSAttributedString *attrTitle = [[NSAttributedString alloc] initWithString: title];
+  AUTORELEASE(attrTitle);
```
fredkiefer, line 71: If you move this one line down you could use RELEASE instead.

Headers/AppKit/NSPathCell.h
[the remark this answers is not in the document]
gcasa: Ok, apologies.  I was trying to fix something else as well. :) Done now.

Source/GSThemeDrawing.m
```diff
@@ -949,6 +951,99 @@ - (void) drawSwitchInRect: (NSRect)rect
         
 }
 
+// NSPathComponentCell
+
+- (void) drawPathComponentCellWithFrame: (NSRect)f
+                                 inView: (NSView *)v
```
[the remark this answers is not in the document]
gcasa: Done.

Source/GSThemeDrawing.m
```diff
@@ -949,6 +951,99 @@ - (void) drawSwitchInRect: (NSRect)rect
...
+                                withURL: (NSURL *)u
```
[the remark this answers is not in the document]
gcasa: Done.

Headers/AppKit/NSPathCell.h
```diff
@@ -0,0 +1,136 @@
...
+  NSArray *_pathItems;
+  NSString *_placeholderString;
+  NSAttributedString *_placeholderAttributedString;
+  NSArray *_allowedTypes;
+  id<NSPathCellDelegate> _delegate;
+  NSURL *_url;
+  SEL _doubleAction;
+  NSArray *_pathComponentCells;
+  NSPathComponentCell *_clickedPathComponentCell;
+  id _objectValue;
```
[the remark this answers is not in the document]
gcasa: For some reason when "NSContents" (the key which corresponds to _object_value is not unarchiving as I expect it to.  It should be the URL.

Source/NSPathControl.m
gcasa: Done.

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -2740,6 +2749,57 @@ - (id) decodeControlContentsForElement: (GSXibElement *)element
...
+    }
+  else // if not specified then assume standard...
+    {
+      num = [NSNumber numberWithInteger: NSPathStyleStandard];
+    }
+
+  return num;  
+}
+
+- (id) decodePathComponentCells: (GSXibElement *)element
```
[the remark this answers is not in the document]
gcasa: See my comment about NSContents in NSCell.m initWithCoder:

Source/GSXib5KeyedUnarchiver.m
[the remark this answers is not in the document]
gcasa: Done

Source/NSMenu.m
gcasa: I believe it always should be.

Source/NSPathCell.m
gcasa: As I said, I cannot seem to get NSContents to unarchive.

Source/NSPathCell.m
gcasa: Doing this.

Source/NSPathCell.m
gcasa: Works now, due to changes above.

Source/NSPathCell.m
gcasa: Added release

Source/NSPathCell.m
gcasa: This has been removed.

Source/NSPathControl.m
gcasa: Does now.

Source/NSPathControl.m
gcasa: done

Headers/AppKit/NSPathControlItem.h
```diff
@@ -0,0 +1,67 @@
...
+
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_10, GS_API_LATEST)
+
+#if	defined(__cplusplus)
+extern "C" {
+#endif
+
+  @class NSURL, NSAttributedString, NSImage, NSString;
+
+@interface NSPathControlItem : NSObject
```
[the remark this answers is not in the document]
gcasa: Do you think I should have NSPathComponentCell contain an NSPathControlItem?  They are very similar so having an NSPathControlItem provide the information for the cell might be useful.

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -3176,6 +3236,14 @@ - (BOOL) containsValueForKey: (NSString *)key
           hasValue  = [currentElement attributeForKey: @"alternateTitle"] != nil;
           hasValue |= [currentElement attributeForKey: @"alternateImage"] != nil;
         }
+      else if ([@"NSPathStyle" isEqualToString: key])
+        {
+          hasValue = [currentElement attributeForKey: @"pathStyle"] != nil;
+        }
+      else if ([@"NSPathComponentCells" isEqualToString: key])
```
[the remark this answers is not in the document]
gcasa: I'm having trouble getting NSContents.

Source/NSPathControl.m
gcasa: done

Source/NSPathControl.m
gcasa: Done.

Source/NSPathControl.m
gcasa: It is now.

Source/NSPathControl.m
gcasa: the close?  I believe so.

Headers/AppKit/NSPathCell.h
```diff
@@ -0,0 +1,131 @@
...
+
+@protocol NSPathCellDelegate;
+
+@class NSEvent, NSView, NSArray, NSString, NSAttributeString, NSColor, NSPathComponentCell, NSOpenPanel, NSURL;    
+  
+@interface NSPathCell : NSActionCell
+{
+  NSPathStyle _pathStyle;
+  NSColor *_backgroundColor;
+  NSArray *_pathItems;
```
fredkiefer, line 51: You are no longer using this.

Source/GSXib5KeyedUnarchiver.m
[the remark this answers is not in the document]
fredkiefer: You need a setObjectValue: method on NSPathCell that calls the setURL: method with the handed in URL.

Source/NSMenu.m
fredkiefer: Why? If there are multiple views with different coordinate systems between the window and the NSPathControl this assumption is not true. Believe me, having struggled with all these transformations for years.

Source/NSPathCell.m
```diff
@@ -0,0 +1,357 @@
...
+  NSUInteger count = [_pathComponentCells count];
+
+  [super drawInteriorWithFrame: frame
+                        inView: controlView];
+  if (count > 0)
+    {
+      NSEnumerator *en = [_pathComponentCells objectEnumerator];
+      NSPathComponentCell *cell = nil;
+      
+      while((cell = (NSPathComponentCell *)[en nextObject]) != nil)
```
fredkiefer, line 211: Space after while is missing.

Source/NSPathCell.m
```diff
@@ -0,0 +1,357 @@
...
+        {
+          [self setPathStyle: [coder decodeIntegerForKey: @"NSPathStyle"]];
+        }
+      
+      if ([coder containsValueForKey: @"NSPathComponentCells"])
+        {
+          [self setPathComponentCells: [coder decodeObjectForKey: @"NSPathComponentCells"]];
+        }
+
+      if ([coder containsValueForKey: @"NSContents"])
```
fredkiefer, line 239: This already gets decoded by NSCell.

Source/NSPathControl.m
```diff
@@ -0,0 +1,565 @@
...
+#import "AppKit/NSPasteboard.h"
+#import "AppKit/NSMenu.h"
+#import "AppKit/NSOpenPanel.h"
+#import "AppKit/NSPathComponentCell.h"
+#import "AppKit/NSPathControlItem.h"
+#import "AppKit/NSEvent.h"
+#import "AppKit/NSTrackingArea.h"
+
+static NSNotificationCenter *nc = nil;
+Class pathCellClass;
```
fredkiefer, line 40: Better make this static.

Source/NSPathControl.m
fredkiefer: I thought that you had removed these methods from the cell, as Apple does not declare them there.
I am fine with both ways, we just should not keep the values twice.

Source/NSPathCell.m
gcasa: RIGHT... but it's not getting decoded properly by NSCell. :)   That's the problem.

Source/NSPathControl.m
```diff
@@ -0,0 +1,565 @@
...
+{
+  return pathCellClass;
+}
+
++ (void) setCellClass: (Class)classId
+{
+  pathCellClass = classId;
+}
+
+- (void) _resetTrackingRect
```
fredkiefer, line 73: If you name this method resetCursorRects you get it called automatically.

Source/NSPathCell.m
fredkiefer: Could you just write a locale setObjectValue: method and log the parameter you get handed in? It should be the URL.

Source/NSPathControl.m
gcasa: Done

Source/NSPathCell.m
gcasa: Do you have any clue as to why?

gcasa: I tried that, I am working on day job stuff now.   I will return to this in a few hours.

gcasa: fixed... I also figured out why it is implemented separately.  If we setObjectValue on NSPathComponentCell, it will show because it's parent is NSTextField...  we don't want this.  So it's implemented there separately.

I also got the NSContents stuff working.  I am going to commit this and then clean up the path components stuff.  I am tempted to keep it only because that key IS used in the nib stuff.   It might be useful.

gcasa:
BTW, I have been unable to test drag and drop of files since GWorkspace appears to have a broken build.

gcasa:
My changes are complete.  Please let me know if you have any further comments.

Headers/AppKit/NSPathCell.h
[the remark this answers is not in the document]
gcasa: fixed this as well

Headers/AppKit/NSPathControlItem.h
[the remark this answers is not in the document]
gcasa: Deferred

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -2192,6 +2201,11 @@ - (id) decodeCellContentsForElement: (GSXibElement*)element
     {
       object = [element attributeForKey: @"stringValue"];
     }
+  else if ([class isSubclassOfClass: [NSPathCell class]])
+    {
+      GSXibElement *el = [element elementForKey: @"url"];
```
fredkiefer, line 2205: As URLs may appear in other context as well we should add a general way to decode it. I think the way to go here is the dictionary XmlTagToObjectClassMap, where we could define a mapping from "url" to NSURL. But I am unsure whether this shouldn't work automatically?

Source/GSXib5KeyedUnarchiver.m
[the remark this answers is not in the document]
fredkiefer: After your other change this really should be obsolete.

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -3166,6 +3231,7 @@ - (BOOL) containsValueForKey: (NSString *)key
         {
           hasValue  = [currentElement attributeForKey: @"title"] != nil;
           hasValue |= [currentElement attributeForKey: @"image"] != nil;
+          hasValue |= [currentElement attributeForKey: @"string"] != nil;
```
fredkiefer, line 3207: Shouldn't this be "url"?

Source/NSPathCell.m
```diff
@@ -0,0 +1,359 @@
...
+      
+      // Decompose string...
+      do
+        {
+          NSPathComponentCell *cell = [[NSPathComponentCell alloc] init];
+          NSImage *image = nil;
+          NSString *string = [u path];
+          
+          [cell setURL: u];
+          [fm fileExistsAtPath: [url path]
```
fredkiefer, line 298: You could be using "string" here and in the line 301.

Source/NSPathCell.m
```diff
@@ -0,0 +1,359 @@
...
+            {
+              image = [NSImage imageNamed: @"NSComputer"];
+            }
+          else
+            {
+              image = [[NSWorkspace sharedWorkspace] iconForFile: [[url path] lastPathComponent]];
+            }
+
+          [cell setImage: image];
+          string = [string stringByDeletingLastPathComponent];
```
fredkiefer, line 320: I would move this line down to 334

Source/NSPathComponentCell.m
```diff
@@ -0,0 +1,86 @@
...
+#import "GNUstepGUI/GSTheme.h"
+
+@implementation NSPathComponentCell
+
+- (instancetype) init
+{
+  self = [super init];
+  if (self != nil)
+    {
+      _image = nil;
```
fredkiefer, line 41: All three are the default.

Source/NSPathControl.m
```diff
@@ -0,0 +1,565 @@
...
+}
+
+- (void) _createPathItems
+{
+  NSArray *a = [_cell pathComponentCells];
+  NSEnumerator *en = [a objectEnumerator];
+  NSPathComponentCell *c = nil;
+  NSMutableArray *items = [NSMutableArray arrayWithCapacity: [a count]];
+  
+  while((c = [en nextObject]) != nil)
```
fredkiefer, line 160: Again a space after the "while" is missing.

Source/NSPathControl.m
```diff
@@ -0,0 +1,565 @@
...
+  if ([[self cell] action])
+    {
+      [self sendAction: [[self cell] action]
+                    to: [[self cell] target]];
+    }
+  
+  // Tested on OSX it doesn't do this...  it apparently only chooses
+  // the cell, and sends the action.  It doesn't reset the URL.
+  [[sender menu] close];
+  // AUTORELEASE([sender menu]);
```
fredkiefer, line 288: The comment above and this line should be removed.

Source/NSPathControl.m
```diff
@@ -0,0 +1,565 @@
...
+                              types: nil];
+  if (result == NSOKButton)
+    {
+      NSArray *urls = [op URLs];
+      NSURL *url = [urls objectAtIndex: 0];
+      [self setURL: url];
+    }
+
+  [[sender menu] close];
+  // AUTORELEASE([sender menu]);
```
fredkiefer, line 340: Another line to remove.

Source/NSPathControl.m
```diff
@@ -0,0 +1,565 @@
...
+- (NSDragOperation) draggingEntered: (id<NSDraggingInfo>)sender
+{
+  if (_delegate != nil)
+    {
+      NSDragOperation dop = [_delegate pathControl: self
+                                      validateDrop: sender];
+      return dop;
+    }
+  
+  return NSDragOperationNone;
```
fredkiefer, line 437: Why none? Even without a delegate dragging should be possible.

Source/NSPathControl.m
```diff
@@ -0,0 +1,565 @@
...
+{
+  NSPasteboard *pb = [sender draggingPasteboard];
+  if ([[pb types] containsObject: NSFilenamesPboardType])
+    {
+      NSArray *files = [pb propertyListForType: NSFilenamesPboardType];
+      if ([files count] > 0)
+        {
+          NSString *file = [files objectAtIndex: 0];
+          NSURL *u = [NSURL URLWithString: file];
+          BOOL accept = [_delegate pathControl: self
```
fredkiefer, line 450: Again, even without a delegate dropping should be possible.

Source/NSPathControl.m
```diff
@@ -0,0 +1,565 @@
...
+            }
+        }
+    }
+  return YES;
+}
+
+- (void) setFrame: (NSRect)frame
+{
+  [super setFrame: frame];
+  [self resetCursorRects];
```
fredkiefer, line 468: This should not be needed as it already happens in NSView. That was the idea of renaming the method,

Source/NSPathControl.m
```diff
@@ -0,0 +1,565 @@
...
+              NSString *s = [coder decodeObjectForKey: @"NSControlAction"];
+              [self setAction: NSSelectorFromString(s)];
+            }
+          if ([coder containsValueForKey: @"NSControlTarget"])
+            {
+              id t = [coder decodeObjectForKey: @"NSControlTarget"];
+              [self setTarget: t];
+            }
+          
+          [self resetCursorRects];
```
fredkiefer, line 503: Again this should not be needed.

Source/NSPathControl.m
```diff
@@ -0,0 +1,565 @@
...
+{
+  NSPathControlItem *pi = [[NSPathControlItem alloc] init];
+  NSURL *u = [self URL];
+  NSString *path = [u path];
+  
+  [pi setImage: [self image]];
+  [pi setURL: u];
+  [pi setTitle: [path lastPathComponent]];
+
+  return pi;
```
fredkiefer, line 562: AUTORELEASE is missing here.

Source/GSXib5KeyedUnarchiver.m
gcasa: No.  Url is an element.  Its value is string.

Source/NSPathCell.m
gcasa: Fixed

Source/GSXib5KeyedUnarchiver.m
gcasa: This does work.   Should we leave it or come up with something better?

Source/NSPathControl.m
gcasa: I just figured that might be the default..

gcasa: Changed to Copy until I can test it more.

Source/NSPathControl.m
gcasa: Removed NO

gcasa:
All changes made.  Let me know if you have any additional comments.

Source/GSThemeDrawing.m
```diff
@@ -949,6 +951,100 @@ - (void) drawSwitchInRect: (NSRect)rect
         
 }
 
+// NSPathComponentCell
+
+- (void) drawPathComponentCellWithFrame: (NSRect)f
```
fredkiefer, line 956: Could you please use more meaningful variable names in this method?

Source/GSThemeDrawing.m
```diff
@@ -949,6 +951,100 @@ - (void) drawSwitchInRect: (NSRect)rect
...
+  NSString *string = [[u path] lastPathComponent];
+  NSRect textFrame = f;
+  NSRect imgFrame = f;
+  NSRect arrowFrame = f;
+  NSImage *arrowImage = [NSImage imageNamed: @"NSMenuArrow"];
+  NSPathStyle s = [pc pathStyle];
+  
+  if (s == NSPathStyleStandard || s == NSPathStyleNavigationBar)
+    {
+      /*
```
fredkiefer, line 972: If this code is not needed it would be better to remove it.

Source/GSThemeDrawing.m
```diff
@@ -949,6 +951,100 @@ - (void) drawSwitchInRect: (NSRect)rect
...
+      if (s == NSPathStyleNavigationBar)
+        {
+          [[NSColor lightGrayColor] set];
+          NSRectFill(f);
+          [[NSColor controlShadowColor] set];
+          NSFrameRectWithWidth(f, 1);
+        }
+      */
+      // Modify positions...
+      textFrame.origin.x += 25.0; // the width of the image plus a few pixels.
```
fredkiefer, line 982: Don't all these values depend on the size of the image?

Source/GSThemeDrawing.m
```diff
@@ -949,6 +951,100 @@ - (void) drawSwitchInRect: (NSRect)rect
...
+          [[NSColor controlShadowColor] set];
+          NSFrameRectWithWidth(newFrame, .5);
+          
+          // Use control frame...
+          textFrame = newFrame;
+          imgFrame = newFrame;
+          arrowFrame = newFrame;
+          
+          // Modify positions...
+          textFrame.origin.x += 25.0; // the width of the image plus a few pixels.
```
fredkiefer, line 1025: Again here

Source/GSXib5KeyedUnarchiver.m
fredkiefer: I really would prefer to change it. The idea of this class is that we decode each element for itself and construct bigger entities from these components. Here we are reaching through. Sometimes there is no way around that, but here there is.

Source/GSXib5KeyedUnarchiver.m
fredkiefer: Here is an extract from your own XIB file:
<pathCell key="cell" controlSize="small" selectable="YES" editable="YES" alignment="left" id="hSq-Cw-QmP">
                            <font key="font" metaFont="controlContent" size="11"/>
                            <url key="url" string="file:///usr/"/>
</pathCell>
When looking for NSContents we are on the pathCell element and we should check whether there is an element for the key url. At least that is my understanding of what goes on here. Did you add some NSLog statement to check?

Source/NSMenu.m
gcasa: The change you suggested puts the menu off window

Source/GSXib5KeyedUnarchiver.m
gcasa: Handling this in a separate PR.

Source/GSXib5KeyedUnarchiver.m
gcasa: Handling this in a separate PR.

gcasa:
Fred, I am going to handle the Xib related changes in a separate PR.  Please let me know if you approve of these changes outside of that.

Source/GSXib5KeyedUnarchiver.m
fredkiefer: Does this just mean you want to merge the code as is and clean up later?

Source/NSMenu.m
fredkiefer: What I suggested would look something like this:

 NSRect cellFrame = [view convertRect: [view frame] toView: nil];
 cellFrame = [[view window] convertRectToScreen: cellFrame];

But looking at it again I think that we need to use [view bounds]  instead of [view frame], as the frame is in the parent view coordinate system (Which is why your code works for the simple case). Coordinate transformations are always confusing.

Source/NSPathCell.m
```diff
@@ -0,0 +1,359 @@
...
+  _delegate = delegate;
+}
+
+- (void) drawInteriorWithFrame: (NSRect)frame inView: (NSView *)controlView
+{
+  NSUInteger count = [_pathComponentCells count];
+
+  [super drawInteriorWithFrame: frame
+                        inView: controlView];
+  if (count > 0)
```
fredkiefer, line 217: We do not need to special case 0, in that case the enumerator would just return no elements.

Source/NSPathCell.m
```diff
@@ -0,0 +1,359 @@
...
+{
+  NSMutableArray *array = [NSMutableArray arrayWithCapacity: 10];
+  
+  // Create cells
+  if (url != nil)
+    {
+      BOOL isDir = NO;
+      BOOL at_root = NO;
+      NSFileManager *fm = [NSFileManager defaultManager];
+      NSURL *u = [url copy];
```
fredkiefer, line 288: Why copy, we never change u.

Source/NSPathCell.m
```diff
@@ -0,0 +1,359 @@
...
+      
+      // Decompose string...
+      do
```
fredkiefer, line 291: I prefer while loops and in this specific case it would make no difference to move the condition up as at_root is NO at the start.

Source/NSPathComponentCell.m
```diff
@@ -0,0 +1,74 @@
...
+
+#import "AppKit/NSPathComponentCell.h"
+#import "AppKit/NSImage.h"
+#import "AppKit/NSStringDrawing.h"
+
+#import "GNUstepGUI/GSTheme.h"
+
+@implementation NSPathComponentCell
+
+- (void) dealloc
```
fredkiefer, line 36: This reminds me, we seem to be missing dealloc methods in the other classes.

Source/NSPathComponentCell.m
```diff
@@ -0,0 +1,74 @@
...
+   Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+   Boston, MA 02110 USA.
+*/
+
+#import <Foundation/NSURL.h>
+#import <Foundation/NSGeometry.h>
+
+#import "AppKit/NSPathComponentCell.h"
+#import "AppKit/NSImage.h"
+#import "AppKit/NSStringDrawing.h"
```
fredkiefer, line 30: This import is not needed.

Source/NSPathControlItem.m
```diff
@@ -0,0 +1,76 @@
...
+   License along with this library; if not, write to the Free
+   Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+   Boston, MA 02110 USA.
+*/
+
+#import <Foundation/NSString.h>
+#import <Foundation/NSURL.h>
+#import "AppKit/NSPathControlItem.h"
+#import "AppKit/NSImage.h"
+#import "AppKit/NSAttributedString.h"
```
fredkiefer, line 29: I think NSAttributedString has been mostly moved to base, which is still confusing to me.

Source/NSPathControl.m
fredkiefer: Sorry, but the space is still missing. Why do you resolve the conversation without resolving the issue?

Source/NSPathControl.m
```diff
@@ -0,0 +1,549 @@
...
+}
+
+- (NSArray *) pathItems
+{
+  return _pathItems;
+}
+
+- (void) setPathItems: (NSArray *)items
+{
+  ASSIGNCOPY(_pathItems, items);
```
fredkiefer, line 236: If this gets called from the outside the cell and all its components will be out of sync.

Source/NSPathControl.m
fredkiefer: Did you test it?

Source/NSPathControl.m
```diff
@@ -0,0 +1,549 @@
...
+          RELEASE(i);
+        }
+
+      // Add separator
+      [menu insertItem: [NSMenuItem separatorItem]
+               atIndex: 0];
+
+      // Add choose menu option
+      NSMenuItem *i = [[NSMenuItem alloc] init];
+      [i setTitle: @"Choose..."];
```
fredkiefer, line 376: This should be internationalised by putting _() around the string.

Source/NSPathControl.m
```diff
@@ -0,0 +1,549 @@
...
+      [i setTarget: self];
+      [i setAction: @selector(_doChooseMenuAction:)];
+      [menu insertItem: i
+               atIndex: 0];
+      RELEASE(i);
+
+      [self setMenu: menu];
+      // RELEASE(menu);
```
fredkiefer, line 384: Please remove this line as you already autorelease the menu above.

Source/NSPathControl.m
fredkiefer: When delegate is nil this call will result in NO and no dropping happens. You should set accept to YES if delegate is nil.

Source/NSPathControl.m
```diff
@@ -0,0 +1,549 @@
...
+
+- (instancetype) initWithCoder: (NSKeyedUnarchiver *)coder
+{
+  self = [super initWithCoder: coder];
+  if (self != nil)
+    {
+      if ([coder allowsKeyedCoding])
+        {
+          // Defaults for some values which aren't encoded unless they are non-default.
+          [self setBackgroundColor: [NSColor windowBackgroundColor]];
```
fredkiefer, line 464: It is better to put these into the else branch of the following if's.

Source/NSPathControl.m
gcasa: I dont know how i missed it.  Fixed.

Source/NSMenu.m
gcasa: Got it and fixed.

gcasa:
Last of your suggestions has been integrated.  Please let me know if I'm clear to merge.

Source/NSMenu.m
```diff
@@ -1619,13 +1619,33 @@ + (void) popUpContextMenu: (NSMenu*)menu
...
+                        inView: (NSView *) view
+{
+  NSRect cellFrame = [view convertRect: [view bounds] toView: nil];
+  NSWindow *w = [view window];
+  NSMenuView *mr = [self menuRepresentation];
+  NSUInteger selectedItem = [self indexOfItem: item];
+
+  cellFrame = [[view window] convertRectToScreen: cellFrame];
+  cellFrame.origin.x += point.x;
```
fredkiefer, line 1639: Do you know in which coordinate system point is? We may have to fix this later on.

gcasa: I believe it is the point within the given view.  I will do more testing and fix in a later PR.

gcasa:
Thanks!  Merging!  I will start the XIB branch right after this merge.

PR #66, opened 2020-06-24
Add setIsVisible method.  It is 10.0 and was missing for some reason.

gcasa:
This is a simple change.

### DIFF
--- a/Headers/AppKit/NSWindow.h
+++ b/Headers/AppKit/NSWindow.h
@@ -528,6 +528,7 @@ PACKAGE_SCOPE
 - (void) orderWindow: (NSWindowOrderingMode)place
 	  relativeTo: (NSInteger)otherWin;
 - (BOOL) isVisible;
+- (void) setIsVisible: (BOOL)flag;
 - (NSInteger) level;
 - (void) setLevel: (NSInteger)newLevel;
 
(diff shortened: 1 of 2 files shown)


fredkiefer:
> The last time I looked this was a read only property. Where did you find this method?

I was wrong here. There is the read only property visible and a method setIsVisible. https://developer.apple.com/documentation/appkit/nswindow/1449570-setisvisible?language=objc

The implementation here looks OK, the only thing we need to still think about are miniaturised windows. Should those be handled here as well?

gcasa:
> > The last time I looked this was a read only property. Where did you find this method?
> 
> I was wrong here. There is the read only property visible and a method setIsVisible. https://developer.apple.com/documentation/appkit/nswindow/1449570-setisvisible?language=objc
> 
> The implementation here looks OK, the only thing we need to still think about are miniaturised windows. Should those be handled here as well?

I can add something to deminiaturize as well.   It's a one line change.

gcasa:
> > > The last time I looked this was a read only property. Where did you find this method?
> > 
> > 
> > I was wrong here. There is the read only property visible and a method setIsVisible. https://developer.apple.com/documentation/appkit/nswindow/1449570-setisvisible?language=objc
> > The implementation here looks OK, the only thing we need to still think about are miniaturised windows. Should those be handled here as well?
> 
> I can add something to deminiaturize as well. It's a one line change.

Correction: The documentation doesn't mention this.  I think we should go ahead with the implementation as-is.

PR #73, opened 2020-08-02
NSText* classes remaining in MISSING

gcasa:
Create either template implementations or full implementations of these classes.

### DIFF
--- a/Headers/AppKit/NSTextCheckingClient.h
+++ b/Headers/AppKit/NSTextCheckingClient.h
@@ -1,0 +1,47 @@
+/* Definition of class NSTextCheckingClient
+   Copyright (C) 2020 Free Software Foundation, Inc.
+   
+   By: Gregory John Casamento
+   Date: 02-08-2020
+
+   This file is part of the GNUstep Library.
+   
+   This library is free software; you can redistribute it and/or
+   modify it under the terms of the GNU Lesser General Public
+   License as published by the Free Software Foundation; either
+   version 2.1 of the License, or (at your option) any later version.
+   
+   This library is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+   Lesser General Public License for more details.
+   
+   You should have received a copy of the GNU Lesser General Public
+   License along with this library; if not, write to the Free
+   Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+   Boston, MA 02110 USA.
+*/
+
+#ifndef _NSTextCheckingClient_h_GNUSTEP_GUI_INCLUDE
+#define _NSTextCheckingClient_h_GNUSTEP_GUI_INCLUDE
+
+#import <Foundation/NSObject.h>
+
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_15, GS_API_LATEST)
+
+#if	defined(__cplusplus)
+extern "C" {
+#endif
+
+@protocol NSTextCheckingClient
+
+@end
+
+#if	defined(__cplusplus)
+}
+#endif
+
+#endif	/* GS_API_MACOSX */
+
+#endif	/* _NSTextCheckingClient_h_GNUSTEP_GUI_INCLUDE */
+
--- a/Headers/AppKit/NSTextInputClient.h
+++ b/Headers/AppKit/NSTextInputClient.h
@@ -1,0 +1,47 @@
+/* Definition of class NSTextInputClient
+   Copyright (C) 2020 Free Software Foundation, Inc.
+   
+   By: Gregory John Casamento
+   Date: 02-08-2020
+
+   This file is part of the GNUstep Library.
+   
+   This library is free software; you can redistribute it and/or
+   modify it under the terms of the GNU Lesser General Public
+   License as published by the Free Software Foundation; either
+   version 2.1 of the License, or (at your option) any later version.
+   
+   This library is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+   Lesser General Public License for more details.
+   
+   You should have received a copy of the GNU Lesser General Public
+   License along with this library; if not, write to the Free
+   Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+   Boston, MA 02110 USA.
+*/
+
+#ifndef _NSTextInputClient_h_GNUSTEP_GUI_INCLUDE
+#define _NSTextInputClient_h_GNUSTEP_GUI_INCLUDE
+
+#import <Foundation/NSObject.h>
+
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_5, GS_API_LATEST)
+
+#if	defined(__cplusplus)
+extern "C" {
+#endif
+
+@protocol NSTextInputClient
+
+@end
+
+#if	defined(__cplusplus)
+}
+#endif
+
+#endif	/* GS_API_MACOSX */
+
+#endif	/* _NSTextInputClient_h_GNUSTEP_GUI_INCLUDE */
+
--- a/Headers/AppKit/NSTextInputContext.h
+++ b/Headers/AppKit/NSTextInputContext.h
@@ -1,0 +1,47 @@
+/* Interface of class NSTextInputContext
+   Copyright (C) 2020 Free Software Foundation, Inc.
+   
+   By: Gregory John Casamento
+   Date: 02-08-2020
+
+   This file is part of the GNUstep Library.
+   
+   This library is free software; you can redistribute it and/or
+   modify it under the terms of the GNU Lesser General Public
+   License as published by the Free Software Foundation; either
+   version 2.1 of the License, or (at your option) any later version.
+   
+   This library is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+   Lesser General Public License for more details.
+   
+   You should have received a copy of the GNU Lesser General Public
+   License along with this library; if not, write to the Free
+   Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+   Boston, MA 02110 USA.
+*/
+
+#ifndef _NSTextInputContext_h_GNUSTEP_GUI_INCLUDE
+#define _NSTextInputContext_h_GNUSTEP_GUI_INCLUDE
+
+#import <Foundation/NSObject.h>
+
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_6, GS_API_LATEST)
+
+#if	defined(__cplusplus)
+extern "C" {
+#endif
+
+@interface NSTextInputContext : NSObject
+
+@end
+
+#if	defined(__cplusplus)
+}
+#endif
+
+#endif	/* GS_API_MACOSX */
+
+#endif	/* _NSTextInputContext_h_GNUSTEP_GUI_INCLUDE */
+
--- a/Source/NSTextCheckingController.m
+++ b/Source/NSTextCheckingController.m
@@ -1,0 +1,30 @@
+/* Implementation of class NSTextCheckingController
+   Copyright (C) 2020 Free Software Foundation, Inc.
+   
+   By: Gregory John Casamento
+   Date: 02-08-2020
+
+   This file is part of the GNUstep Library.
+   
+   This library is free software; you can redistribute it and/or
+   modify it under the terms of the GNU Lesser General Public
+   License as published by the Free Software Foundation; either
+   version 2.1 of the License, or (at your option) any later version.
+   
+   This library is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+   Lesser General Public License for more details.
+   
+   You should have received a copy of the GNU Lesser General Public
+   License along with this library; if not, write to the Free
+   Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+   Boston, MA 02110 USA.
+*/
+
+#import "AppKit/NSTextCheckingController.h"
+
+@implementation NSTextCheckingController
+
+@end
+
--- a/Source/NSTextFinder.m
+++ b/Source/NSTextFinder.m
@@ -1,0 +1,30 @@
+/* Implementation of class NSTextFinder
+   Copyright (C) 2020 Free Software Foundation, Inc.
+   
+   By: Gregory John Casamento
+   Date: 02-08-2020
+
+   This file is part of the GNUstep Library.
+   
+   This library is free software; you can redistribute it and/or
+   modify it under the terms of the GNU Lesser General Public
+   License as published by the Free Software Foundation; either
+   version 2.1 of the License, or (at your option) any later version.
+   
+   This library is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+   Lesser General Public License for more details.
+   
+   You should have received a copy of the GNU Lesser General Public
+   License along with this library; if not, write to the Free
+   Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+   Boston, MA 02110 USA.
+*/
+
+#import "AppKit/NSTextFinder.h"
+
+@implementation NSTextFinder
+
+@end
+
--- a/Headers/AppKit/AppKit.h
+++ b/Headers/AppKit/AppKit.h
@@ -246,6 +246,11 @@
 #import <AppKit/NSTextAlternatives.h>
 #import <AppKit/NSTextAttachment.h>
 #import <AppKit/NSTextContainer.h>
+#import <AppKit/NSTextCheckingClient.h>
+#import <AppKit/NSTextCheckingController.h>
+#import <AppKit/NSTextFinder.h>
+#import <AppKit/NSTextInputClient.h>
+#import <AppKit/NSTextInputContext.h>
 #import <AppKit/NSTextList.h>
 #import <AppKit/NSTextStorage.h>
 #import <AppKit/NSTextTable.h>
--- a/Headers/AppKit/NSTextCheckingController.h
+++ b/Headers/AppKit/NSTextCheckingController.h
@@ -1,0 +1,47 @@
+/* Interface of class NSTextCheckingController
+   Copyright (C) 2020 Free Software Foundation, Inc.
+   
+   By: Gregory John Casamento
+   Date: 02-08-2020
+
+   This file is part of the GNUstep Library.
+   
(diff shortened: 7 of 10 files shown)


Headers/AppKit/NSTextCheckingClient.h
```diff
@@ -0,0 +1,113 @@
...
+ NSTextInputTraitTypeNo,
+ NSTextInputTraitTypeYes
+};
+typedef NSInteger NSTextInputTraitType;
+
+@class NSDictionary, NSCandidateListTouchBarItem;
+
+@protocol NSTextInputTraits
+
+@optional
```
fredkiefer, line 53: Normally we use something like this for optional protocols

`#if GS_PROTOCOLS_HAVE_OPTIONAL
 @optional
#else
@end
@interface NSObject (NSFilePresenter)
#endif`

Not sure for which compilers this is still needed.

gcasa: It seems to work ok on clang and gcc in ci. Ill add those guards

gcasa: done.

Source/NSTextCheckingController.m
```diff
@@ -0,0 +1,130 @@
...
+
+// initializer
+- (instancetype) initWithClient: (id<NSTextCheckingClient>)client
+{
+  self = [super init];
+  if (self != nil)
+    {
+      _client = client;
+      _spellCheckerDocumentTag = 0;
+      _menu = nil;
```
fredkiefer, line 37: What is this instance variable for? It never gets used anywhere.

Source/NSTextView.m
```diff
@@ -6078,6 +6081,96 @@ - (void) performFindPanelAction: (id)sender
...
+  NSValue *r = [[self selectedRanges] objectAtIndex: 0];
+  return [r rangeValue];
+}
+
+- (BOOL) shouldReplaceCharactersInRanges: (NSArray *)ranges withStrings: (NSArray *)strings
+{
+  return YES;
+}
+
+- (void) replaceCharactersInRange: (NSRange)range withString: (NSString *)string
```
fredkiefer, line 6116: I think this method is already implemented in Lin 2285 and as you can see there it is a bit more complicated to change anything inside an NSTextView.

Source/NSTextCheckingController.m
gcasa: Based on some of the methods I thought it might be needed... if it's not, I will remove it.  I'm assuming you're referring to "_menu"

Source/NSTextView.m
gcasa: Removed

Source/NSTextCheckingController.m
gcasa: removed

Headers/AppKit/NSTextCheckingClient.h
fredkiefer: Somehow I am not seeing this change. What is wrong?

Headers/AppKit/NSTextInputClient.h
```diff
@@ -0,0 +1,85 @@
...
+
+#if	defined(__cplusplus)
+extern "C" {
+#endif
+
+@class NSArray;
+
+@protocol NSTextInputClient
+
+@required
```
fredkiefer, line 43: Again a protocol where required and @optional are not protected by guards.

Headers/AppKit/NSTextInputContext.h
```diff
@@ -0,0 +1,79 @@
...
+@interface NSTextInputContext : NSObject
+
++ (NSTextInputContext *) currentInputContext;
+
+- (instancetype) initWithClient: (id<NSTextInputClient>)client;
+
+- (id<NSTextInputClient>) client; 
+
+- (BOOL) acceptsGlyphInfo;
+- (void) setAccessGlyphInfo: (BOOL)flag;
```
fredkiefer, line 57: This method name is wrong.

Source/NSTextCheckingController.m
fredkiefer: No you didn't.

Source/NSTextFinder.m
```diff
@@ -0,0 +1,305 @@
...
+
+@implementation NSTextFinder
+
+- (instancetype) init
+{
+  self = [super init];
+  if (self != nil)
+    {
+      // initialize finder...
+      _finder = [GSTextFinder sharedTextFinder];
```
fredkiefer, line 41: Could you please explain the concept here? Why have all the instances of NSTextFinder, which may have different instance values, point to the same GSTextFinder? That way you loose all the differences based on the instance variables.

Headers/AppKit/NSTextCheckingClient.h
fredkiefer: Now I see it, sorry

Source/NSTextFinder.m
```diff
@@ -0,0 +1,305 @@
...
+{
+  [_finder showFindPanel];
+}
+
+- (void) _hideReplaceInterface
+{
+}
+
+// Validating and performing
+- (void) performAction: (NSTextFinderAction)op
```
fredkiefer, line 115: This class needs also to implement the NSResponder method performTextFinderAction: (and for compatibly with old versions also performFindPanelAction:) which gets the tag of the sender and calls this method here.

Source/NSTextFinder.m
```diff
@@ -0,0 +1,305 @@
...
+    case NSTextFinderActionHideReplaceInterface:
+      [self _hideReplaceInterface];
+      break;
+    default:
+      NSLog(@"Unknown operation: %ld", op);
+      break;
+    }
+}
+
+- (BOOL) validateAction: (NSTextFinderAction)op
```
fredkiefer, line 164: The same here. We need to implement validateUserInterfaceItem: and call this method from there when the action is performTextFinderAction: or performFindPanelAction:.

Source/NSTextView.m
```diff
@@ -6078,6 +6081,88 @@ - (void) performFindPanelAction: (id)sender
...
+// NSTextFinder methods implementation...
+// isSelectable, isEditable, string, selectedRanges, setSelectedRanges, replaceCharactersInRange:withString:
+// implemented by NSTextView already...
+
+- (BOOL) allowsMultipleSelection
+{
+  return NO;
+}
+
+- (NSString *) stringAtIndex: (NSUInteger)characterIndex
```
fredkiefer, line 6093: For NSTextView it would actually be better not to implement the next two methods. In that case the NSTextFinder would just use the "string" method. But if you really want to implement those, this method here should return the full string of the NSTextView.

Source/NSTextView.m
```diff
@@ -6078,6 +6081,88 @@ - (void) performFindPanelAction: (id)sender
...
+
+- (NSRange) firstSelectedRange
+{
+  NSValue *r = [[self selectedRanges] objectAtIndex: 0];
+  return [r rangeValue];
+}
+
+- (BOOL) shouldReplaceCharactersInRanges: (NSArray *)ranges withStrings: (NSArray *)strings
+{
+  return YES;
```
fredkiefer, line 6113: Maybe this should call the method shouldChangeTextInRangereplacementString:?

Source/NSTextView.m
```diff
@@ -6078,6 +6081,88 @@ - (void) performFindPanelAction: (id)sender
...
+}
+
+- (void) didReplaceCharacters
+{
+  // nothing...
```
fredkiefer, line 6118: And this should most likely call didChangeText.

Source/NSTextView.m
```diff
@@ -6078,6 +6081,88 @@ - (void) performFindPanelAction: (id)sender
...
+}
+
+- (NSView *) contentViewAtIndex: (NSUInteger)index effectiveCharacterRange: (NSRangePointer)outRange
+{
+  return nil;
+}
+
+- (NSArray *) rectsForCharacterRange: (NSRange)range
+{
+  NSValue *value = [NSValue valueWithRect: [self rectForCharacterRange: range]];
```
fredkiefer, line 6128: Please have a look at the implementation of firstRectForCharacterRange in this class.

Source/NSTextView.m
```diff
@@ -6078,6 +6081,88 @@ - (void) performFindPanelAction: (id)sender
...
+      NSValue *value = [NSValue valueWithRange: visibleRange];
+      result = [NSArray arrayWithObject: value];
+    }
+ 
+  return result;
+}
+
+- (void) drawCharactersInRange: (NSRange)range forContentView: (NSView *)view
+{
+  NSArray *rectsArray = [self rectsForCharacterRange: range];
```
fredkiefer, line 6156: I think the code here should rather contain the two calls to the layout manager from the method drawRect:

Headers/AppKit/NSTextView.h
```diff
@@ -247,6 +248,9 @@ therefore be stored in the NSLayoutManager to avoid problems.
   // Text checking (spelling/grammar)
   NSTimer *_textCheckingTimer;
   NSRect _lastCheckedRect;
+
+  // NSTextFinder vars...
+  NSArray *_selectedRanges;
```
fredkiefer, line 253: You are currently not using this new instance variable. Either implement something based on this or remove the declaration again.

Headers/AppKit/NSTextCheckingClient.h
```diff
@@ -0,0 +1,117 @@
...
+
+enum
+{
+ NSTextInputTraitTypeDefault,
+ NSTextInputTraitTypeNo,
+ NSTextInputTraitTypeYes
+};
+typedef NSInteger NSTextInputTraitType;
+
+@class NSDictionary, NSCandidateListTouchBarItem;
```
fredkiefer, line 49: At the moment the class NSCandidateListTouchBarItem isn't implemented anywhere in GNUstep base.

Source/NSTextCheckingController.m
gcasa: I'm not sure why you're not seeing this...
// initializer
- (instancetype) initWithClient: (id<NSTextCheckingClient>)client
{
  self = [super init];
  if (self != nil)
    {
      _client = client;
      _spellCheckerDocumentTag = 0;
    }
  return self;
}

And I have pushed.

Headers/AppKit/NSTextInputContext.h
gcasa: Fixed.

Source/NSTextView.m
gcasa: Done

Headers/AppKit/NSTextView.h
gcasa: Removed.

Source/NSTextCheckingController.m
gcasa: This should be gone.  Please pull and check on your end.  Re-open this if you still don't see the change.

Source/NSTextFinder.m
gcasa: Done.

Source/NSTextFinder.m
gcasa: I don't know what the validateUserInterfaceItem method should look like.  Any suggestions?

Source/NSTextView.m
gcasa: Done.

Source/NSTextView.m
gcasa: Done

Headers/AppKit/NSTextCheckingClient.h
gcasa: It's implemented as a "mac specific" class in gui currently.  I've been meaning to speak with you about the touch bar stuff in specific.   I will open an issue on the subject and assign it to both of us.

gcasa: Done

Source/NSTextView.m
gcasa: Done.

Source/NSTextView.m
gcasa: Please elaborate, if possible.

Source/NSTextFinder.m
```diff
@@ -0,0 +1,332 @@
...
+{
+  [_finder showFindPanel];
+}
+
+- (void) _hideReplaceInterface
+{
+}
+
+// Validating and performing
+- (void) performTextFinderAction: (id)sender
```
fredkiefer, line 116: I think this method should have the same code as performFindPanelAction: (or rather call that).

Source/NSTextFinder.m
```diff
@@ -0,0 +1,332 @@
...
+      if (_finder == nil)
+        {
+          _finder = [[GSTextFinder alloc] init];
+        }
+    }
+}
+
+- (void) performFindPanelAction: (id)sender
+{
+  BOOL valid = [self validateUserInterfaceAction: sender];
```
fredkiefer, line 130: Normally validation happens independently of the action. Are you sure this should be called here?

Source/NSTextFinder.m
```diff
@@ -0,0 +1,332 @@
...
+    case NSTextFinderActionHideReplaceInterface:
+      [self _hideReplaceInterface];
+      break;
+    default:
+      NSLog(@"Unknown operation: %ld", op);
+      break;
+    }
+}
+
+- (BOOL) validateUserInterfaceAction: (id<NSValidatedUserInterfaceItem>)item
```
fredkiefer, line 186: The name of this method is validateUserInterfaceItem: The code should look something like this
SEL action = [item action];
if (sel_isEqual(action, @selector(performTextFinderAction:)) || sel_isEqual(action, @selector(performFindPanelAction:)))
    return [self validateAction: [item tag]]

return YES;

gcasa: Done

Source/NSTextFinder.m
gcasa: probably not.  I will change it.

Source/NSTextFinder.m
gcasa: Done

Source/NSTextFinder.m
gcasa: Done

Source/NSTextView.m
gcasa:
```
  /* Then draw the special background of the new glyphs.  */
  [_layoutManager drawBackgroundForGlyphRange: drawnRange
                  atPoint: _textContainerOrigin];

  [_layoutManager drawGlyphsForGlyphRange: drawnRange 
                  atPoint: _textContainerOrigin];
```
Do you mean these?

gcasa: On the assumption that that is what you meant, done.

Source/NSTextCheckingController.m
```diff
@@ -0,0 +1,137 @@
...
+
+// instance methods...
+- (void) changeSpelling: (id)sender
+{
+}
+
+- (void) checkSpelling: (id)sender
+{
+    int wordCount = 0;
+  [[NSSpellChecker sharedSpellChecker] checkSpellingOfString: [_client string]
```
fredkiefer, line 66: Looks like something with the indentation is wrong here.

Source/NSTextFinder.m
```diff
@@ -0,0 +1,333 @@
...
+
+@implementation NSTextFinder
+
+- (instancetype) init
+{
+  self = [super init];
+  if (self != nil)
+    {
+      // initialize finder...
+      _finder = [[GSTextFinder alloc] init];
```
fredkiefer, line 42: In the long run I would expect that this class should replace GSTextFinder, but for now this reuse might work. Not really sure.

Source/NSTextFinder.m
```diff
@@ -0,0 +1,333 @@
...
+
+// Validating and performing
+- (void) performTextFinderAction: (id)sender
+{
+  [self performFindPanelAction: sender];
+}
+
+- (void) performFindPanelAction: (id)sender
+{
+  if (_finder == nil)
```
fredkiefer, line 123: How could _finder ever be nil here?

Source/NSTextFinder.m
```diff
@@ -0,0 +1,333 @@
...
+    {
+      _finder = [[GSTextFinder alloc] init];
+    }
+  [self performAction: [sender tag]];
+  [self validateUserInterfaceAction: sender];
```
fredkiefer, line 128: The call to validateUserInterfaceAction: should rather be done by the standard UI mechanism. We should not do it here.

Source/NSTextFinder.m
```diff
@@ -0,0 +1,333 @@
...
+    {
+      return [self validateAction: [item tag]];
+    }
+
+  return YES;
+}
+
+- (BOOL) validateAction: (NSTextFinderAction)op
+{
+  return YES;
```
fredkiefer, line 194: The actual code is missing here. See GSTextFinder for an example.

Source/NSTextView.m
```diff
@@ -89,6 +93,8 @@
 #import "AppKit/NSTextStorage.h"
 #import "AppKit/NSTextView.h"
 #import "AppKit/NSWindow.h"
+#import "AppKit/NSStringDrawing.h"
```
fredkiefer, line 96: This import should no longer be needed.

Source/NSTextView.m
```diff
@@ -6078,6 +6080,116 @@ - (void) performFindPanelAction: (id)sender
...
+  
+  if (_layoutManager)
+    {
+      NSUInteger idx = 0;
+      rects = [_layoutManager rectArrayForCharacterRange: range
+                            withinSelectedCharacterRange: NSMakeRange(NSNotFound, 0)
+                                         inTextContainer: _textContainer
+                                               rectCount: &rectCount];
+      
+      for(idx = 0; idx < rectCount; idx++)
```
fredkiefer, line 6151: Here a space after "for" is missing.

Source/NSTextView.m
```diff
@@ -6078,6 +6080,116 @@ - (void) performFindPanelAction: (id)sender
...
+  return nil;
+}
+
+- (NSArray *) rectsForCharacterRange: (NSRange)range
+{
+  NSUInteger rectCount = 0;
+  NSRect *rects;
+  NSMutableArray *result = [NSMutableArray array];
+  
+  if (_layoutManager)
```
fredkiefer, line 6143: Why this check? In all other places we expect that the layout manager exists, and of course it does.

Source/NSTextCheckingController.m
gcasa: Corrected, check on your side to make sure.

Source/NSTextFinder.m
gcasa: That was my read on this change as well.  It also seems that NSTextCheckingController is meant as a replacement for NSSpellChecker's role as a controller to the spell checking panel which was always awkward in the first place.

Source/NSTextFinder.m
gcasa: Done

Source/NSTextFinder.m
gcasa: Removed

Source/NSTextFinder.m
gcasa: I'll do it until it no longer makes sense, I want to introduce as little risk as possible since this is a rather sensitive portion of the system.

Source/NSTextView.m
gcasa: Removed

Source/NSTextView.m
gcasa: fixed.

Source/NSTextView.m
gcasa: Removed.

Source/NSTextFinder.m
gcasa: Done.

PR #74, opened 2020-08-09
NSGridView branch

gcasa:
Implement NSGridView and NSStackView.


Headers/AppKit/NSGridView.h
```diff
@@ -0,0 +1,179 @@
...
+};
+typedef NSInteger NSGridRowAlignment;
+
+APPKIT_EXPORT const CGFloat NSGridViewSizeForContent;
+
+@class NSGridColumn, NSGridRow, NSGridCell;
+  
+@interface NSGridView : NSView
+
+- (instancetype) initWithFrame: (NSRect)frameRect;
```
fredkiefer, line 63: These two methods are already defined in the super classes.

Source/NSGridView.m
```diff
@@ -0,0 +1,511 @@
...
+  return self;
+}
+
+- (instancetype) initWithViews: (NSArray *)rows
+{
+  self = [self initWithFrame: NSZeroRect];
+
+  if (self != nil)
+    {
+      NSMutableArray *mutableRows = [rows mutableCopy];
```
fredkiefer, line 58: The mutable copy here is completely useless. You might need it two lines below, where you add to _rows.

gnustep-admin: I will fix this.  Forgive my lag in response lately, I have taken on a new contract and it's absorbing some of my extra time.  Thanks. GC  I will get back to this sometime later today.

gcasa:
attaching the enumerated nib for reference....
[MainMenu-enumerated.txt](https://github.com/gnustep/libs-gui/files/5541887/MainMenu-enumerated.txt)

Source/NSGridView.m
gcasa: Removed.

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -287,6 +288,18 @@ + (void) initialize
                                            @"string", @"NS.relative",
                                            @"canPropagateSelectedChildViewControllerTitle",
                                                    @"NSTabViewControllerCanPropagateSelectedChildViewControllerTitle",
+                                           @"rowAlignment", @"NSGrid_alignment",
```
fredkiefer, line 291: We should rather add a mechanism to do this mapping automatically. similar to the one I already included to add the class name without underscore.

Source/NSGridView.m
```diff
@@ -0,0 +1,639 @@
...
+}
+
+- (NSInteger) numberOfRows
+{
+  return [_rows count];
+}
+
+- (NSInteger) numberOfColumns
+{
+  return [_columns count];
```
fredkiefer, line 101: The only place where you set _columns is in the initWithCoder: method.

Source/GSXib5KeyedUnarchiver.m
gcasa: Good suggestion.  I was thinking about that, but since this is the first class I have seen that does it I wasn't sure if I should add such a thing.   I will make the changes.

Source/NSGridView.m
gcasa: True.  Implementation is ongoing.  I still need to implement the methods which are used to programmatically add rows and columns.

Source/GSXib5KeyedUnarchiver.m
gcasa: Where is this happening in the Xib code?

fredkiefer: I would suggest to put that code into the method alternateName:, we could for example scan for "_" and remove everything before that. You just need to make sure this won't interfere with other decoding mechanisms. What ever you do, please document it on the method. The behaviour is already complicated enough.

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -2930,6 +2933,79 @@ - (id) decodeTransitionStyle: (GSXibElement *)element
...
+
+- (id) decodeYPlacementForElement: (GSXibElement *)element
+{
+  id obj = [element attributeForKey: @"yPlacement"];
+  return [self _decodePlacementForObject: obj];
+}
+
+- (id) decodeRowAlignmentForElement: (GSXibElement *)element
+{
+  id obj = [element attributeForKey: @"xPlacement"];
```
fredkiefer, line 2988: I think this should be "rowAlignment"

gcasa: Fixed

Source/GSXib5KeyedUnarchiver.m
gcasa: okay, I will do this.   I am going to get it a little closer to working and then optimize in the way you suggest.

gcasa:
> Could you please explain why you think this method was needed? The inherited method from NSKeyedUnarchiver should be usable here. And if we really need a different behaviour the most likely approach would be to add a method to GSXibKeyedUnarchiver similar to the other methods there.

In the xib, the relevant portion is this:

`                    <gridView fixedFrame="YES" xPlacement="leading" yPlacement="bottom" rowAlignment="none" rowSpacing="9" columnSpacing="9" translatesAutoresizingMaskIntoConstraints="NO" `

This maps to the following in the nib file:

`
	15	<dict>
			<key>$class</key>
			<dict>
				<key>CF$UID</key>
				<integer>162</integer>
			</dict>
			<key>IBNSClipsToBounds</key>
			<integer>0</integer>
			<key>IBNSLayoutMarginsGuide</key>
			<dict>
				<key>CF$UID</key>
				<integer>0</integer>
			</dict>
			<key>IBNSSafeAreaLayoutGuide</key>
			<dict>
				<key>CF$UID</key>
				<integer>0</integer>
			</dict>
			<key>NSDoNotTranslateAutoresizingMask</key>
			<true/>
			<key>NSFrame</key>
			<dict>
				<key>CF$UID</key>
				<integer>104</integer>
			</dict>
			<key>NSGrid_alignment</key>
			<integer>1</integer>
			<key>NSGrid_columnSpacing</key>
			<real>9</real>
			<key>NSGrid_columns</key>
			<dict>
				<key>CF$UID</key>
				<integer>161</integer>
			</dict>
			<key>NSGrid_rowSpacing</key>
			<real>9</real>
			<key>NSGrid_rows</key>
			<dict>
				<key>CF$UID</key>
				<integer>105</integer>
			</dict>
			<key>NSGrid_xPlacement</key>
			<integer>2</integer>
			<key>NSGrid_yPlacement</key>
			<integer>3</integer>
			<key>NSNextResponder</key>
			<dict>
				<key>CF$UID</key>
				<integer>13</integer>
			</dict>
			<key>NSNibTouchBar</key>
			<dict>
				<key>CF$UID</key>
				<integer>0</integer>
			</dict>
			<key>NSSubviews</key>
			<dict>
				<key>CF$UID</key>
				<integer>16</integer>
			</dict>
			<key>NSSuperview</key>
			<dict>
				<key>CF$UID</key>
				<integer>13</integer>
			</dict>
			<key>NSViewWantsBestResolutionOpenGLSurface</key>
			<true/>
			<key>NSvFlags</key>
			<integer>268</integer>
		</dict>
`

The issue is that alignment and x/yPlacement are enumerated types and I must map these types to the proper values using the methods listed in xmlKeyToDecoderSelectorMap.   Is this happening anywhere else so that I don't need this method?

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -3201,6 +3290,42 @@ - (BOOL) decodeBoolForKey: (NSString *)key
   return flag;
 }
 
+- (NSInteger) decodeIntegerForKey: (NSString *)key
```
fredkiefer, line 3293: There should be no reason for this method. Please try to remove it and test again. The code in the super implementation and in decodeObjectForKey: should do the trick.
If there is still an issue it might be the order of the checks in the decodeObjectForKey: method.

   `if ([XmlKeyMapTable objectForKey: key])
        {
          object = [self decodeObjectForKey: [XmlKeyMapTable objectForKey: key]];
        }
      else if ([XmlKeyToDecoderSelectorMap objectForKey: key])
        {`

You could try to change that order, in this case.

gcasa: ok

gcasa: It's not decoding an object... it's decoding an integer.

fredkiefer: Did you actually look at the implementation of the other int decoding methods in GSXibKeyedUnarchiver.m?

`- (int64_t) decodeInt64ForKey: (NSString*)aKey
{
  id o = [self decodeObjectForKey: aKey];

  if (o != nil)
    {
      if (([o isKindOfClass: [NSNumber class]] == YES) ||
          ([o isKindOfClass: [NSString class]] == YES))
	{
	  long long l = [o longLongValue];

	  return l;
	}
      else
	{
	  [NSException raise: NSInvalidUnarchiveOperationException
		      format: @"[%@ +%@]: value for key(%@) is '%@'",
	    NSStringFromClass([self class]), NSStringFromSelector(_cmd),
	    aKey, o];
	}
    }
  return 0;
}`

gcasa: > Did you actually look at the implementation of the other int decoding methods in GSXibKeyedUnarchiver.m?
> 
> `- (int64_t) decodeInt64ForKey: (NSString*)aKey
> {
> id o = [self decodeObjectForKey: aKey];
> 
> if (o != nil)
> {
> if (([o isKindOfClass: [NSNumber class]] == YES) ||
> ([o isKindOfClass: [NSString class]] == YES))
> {
> long long l = [o longLongValue];
> 
> ```
>   return l;
> }
>   else
> {
>   [NSException raise: NSInvalidUnarchiveOperationException
> 	      format: @"[%@ +%@]: value for key(%@) is '%@'",
>     NSStringFromClass([self class]), NSStringFromSelector(_cmd),
>     aKey, o];
> }
> }
> ```
> 
> return 0;
> }`

Yes, I looked.  Please look at the .xib vs. the .nib?   We are going from a text-based value like "none" to an integer which is mapped via an enumerated value to something like NS....None which is 0.   This is not as simple as going from a string value which is a number to an integer.   If it were the existing method would be sufficient.   I am mapping from a text-word to an integer-value.    This is similar to the reason why decodeFromBool exists in GSXib5KeyedUnarchiver.

gcasa: The method you copied above does NOT cover the use case I am trying to implement.

gcasa: > Did you actually look at the implementation of the other int decoding methods in GSXibKeyedUnarchiver.m?
> 
> `- (int64_t) decodeInt64ForKey: (NSString*)aKey
> {
> id o = [self decodeObjectForKey: aKey];
> 
> if (o != nil)
> {
> if (([o isKindOfClass: [NSNumber class]] == YES) ||
> ([o isKindOfClass: [NSString class]] == YES))
> {
> long long l = [o longLongValue];
> 
> ```
>   return l;
> }
>   else
> {
>   [NSException raise: NSInvalidUnarchiveOperationException
> 	      format: @"[%@ +%@]: value for key(%@) is '%@'",
>     NSStringFromClass([self class]), NSStringFromSelector(_cmd),
>     aKey, o];
> }
> }
> ```
> 
> return 0;
> }`

I see it.  I will remove the method and try again.

gcasa: Changed to use decodeIntForKey: this works.  Thanks for the feedback.

Headers/AppKit/NSStackView.h
```diff
@@ -0,0 +1,75 @@
...
+    NSStackViewDistributionFill = 0,
+    NSStackViewDistributionFillEqually,
+    NSStackViewDistributionFillProportionally,
+    NSStackViewDistributionEqualSpacing,
+    NSStackViewDistributionEqualCentering
+};
+typedef NSInteger NSStackViewDistribution;
+
+typedef float NSStackViewVisibilityPriority;
+static const NSStackViewVisibilityPriority NSStackViewVisibilityPriorityMustHold = 1000;
```
fredkiefer, line 58: The last time you did add ".0" to all the floatCGFloat values to make the type clearer.

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -2917,6 +2943,79 @@ - (id) decodeTransitionStyle: (GSXibElement *)element
...
+- (id) decodeYPlacementForElement: (GSXibElement *)element
+{
+  id obj = [element attributeForKey: @"yPlacement"];
+  return [self _decodePlacementForObject: obj];
+}
+
+- (id) decodeRowAlignmentForElement: (GSXibElement *)element
+{
+  id obj = [element attributeForKey: @"rowAlignment"];
+  NSNumber *num = [NSNumber numberWithInteger: 0];
```
fredkiefer, line 2999: Why not something like this:

NSGridRowAlignment align;
if ([obj isEqualToString: @"inherited"])
     {
       align = NSGridRowAlignmentInherited;
     }
   else if ([obj isEqualToString: @"firstBaseline"])
     {
       align = NSGridRowAlignmentFirstBaseline;
     }
   else if ([obj isEqualToString: @"lastBaseline"])
     {
       align = NSGridRowAlignmentLastBaseline;
     }
   else
     {
       align = NSGridRowAlignmentNone;
     }
   return [NSNumber numberWithInteger: align];

Source/NSGridView.m
```diff
@@ -0,0 +1,957 @@
...
+  [_rows insertObject: gr atIndex: index];
+  RELEASE(gr);
+
+  // Insert views...
+  FOR_IN(NSView*, v, views)
+    {
+      NSGridCell *c = [[NSGridCell alloc] init];
+      [c setContentView: v];
+      RELEASE(v);
+      [gr _addCell: c];
```
fredkiefer, line 204: IN your concept you would also need to add the new cell to a column.

Source/NSGridView.m
```diff
@@ -0,0 +1,957 @@
...
+
+- (void) moveRowAtIndex: (NSInteger)fromIndex toIndex: (NSInteger)toIndex
+{
+  [self _refreshCells];
+}
+
+- (void) removeRowAtIndex: (NSInteger)index
+{
+  [_rows removeObjectAtIndex: index];
+  [self _refreshCells];
```
fredkiefer, line 221: This also needs to change the columns by removing the corresponding cells.

Source/NSGridView.m
```diff
@@ -0,0 +1,957 @@
...
+
+- (void) moveColumnAtIndex: (NSInteger)fromIndex toIndex: (NSInteger)toIndex
+{
+  [self _refreshCells]; 
+}
+
+- (void) removeColumnAtIndex: (NSInteger)index
+{
+  [_columns removeObjectAtIndex: index];
+  [self _refreshCells];
```
fredkiefer, line 245: This also needs to change the rows.

Source/NSGridView.m
```diff
@@ -0,0 +1,957 @@
...
+      [coder encodeBool: _isHidden forKey: @"NSGrid_hidden"];
+      [coder encodeFloat: _leadingPadding forKey: @"NSGrid_leadingPadding"];
+      [coder encodeObject: _gridView forKey: @"NSGrid_owningGrid"]; // weak
+      [coder encodeFloat: _trailingPadding forKey: @"NSGrid_trailingPadding"];
+      [coder encodeFloat: _width forKey: @"NSGrid_width"];
+      [coder encodeInteger: _xPlacement forKey: @"NSGrid_xPlacement"];
+    }
+  else
+    {
+      [coder encodeValueOfObjCType:@encode(BOOL)
```
fredkiefer, line 685: Could you please add spaces after the colons to all these methods?

Headers/AppKit/NSGridView.h
```diff
@@ -0,0 +1,212 @@
...
+APPKIT_EXPORT const CGFloat NSGridViewSizeForContent;
+
+@class NSGridColumn, NSGridRow, NSGridCell, NSArray, NSMutableArray;
+  
+@interface NSGridView : NSView
+{
+  NSGridRowAlignment _rowAlignment;
+  NSMutableArray *_rows;
+  NSMutableArray *_columns;
+  NSMutableArray *_cells;
```
fredkiefer, line 66: If I were to implement these classes I would only store references to the cells here. In NSGridRow and NSGridColumn there wouldn't be any cell references and all the managing of the cells would happen here. Otherwise you have three data structures that you have to keep in sync. All the code you need to have on _cells you need anyway and for each row/column you always have access to the NSGridView to get the required information from there.

fredkiefer:
This PR should stillte labeled as draft as it is not functional in its current state.

Headers/AppKit/NSStackView.h
gcasa: Sure, I'll do that, sorry.

Source/GSXib5KeyedUnarchiver.m
gcasa: Yep, doing that now.

gcasa: Fixed there and one other place where I was doing the same thing.

gcasa: Pushed

Source/NSGridView.m
gcasa: Yes, I am aware, working on that.

Source/NSGridView.m
gcasa: Yep, similarly as above

Source/NSGridView.m
gcasa: > This PR should stillte labeled as draft as it is not functional in its current state.

Okay.

Headers/AppKit/NSGridView.h
gcasa: Yep, understood.  I will take the approach you suggest here.

Source/NSGridView.m
gcasa: I am currently refactoring to how you suggested it be done.  Your suggested approach reduces the code significantly.

Source/NSGridView.m
gcasa: Following your suggestion, so this is out of date.

Headers/AppKit/NSGridView.h
gcasa: Working on this now.

gcasa:
[MainMenu-enumerated.nib.txt](https://github.com/gnustep/libs-gui/files/6014316/MainMenu-enumerated.nib.txt)

This is the nib attached.  This is a window containing an NSGridView.  The project containing the test is here:
[https://github.com/gcasa/NSGridView_test.git](https://github.com/gcasa/NSGridView_test.git)

gcasa (commented):
Please find NSGrid_width and NSGrid_height in the nib attached to this PR for an example of what I am referring to in this comment.  The values are very small for some reason.  This nib is the one generated by the build process under macOS.

Source/NSGridView.m
```diff
@@ -0,0 +1,1055 @@
...
+      NSRect rect = NSZeroRect;
+      
+      // Get row and column index...
+      ci = i % num_col;
+      ri = i / num_col;
+      
+      if (v != nil)
+        {
+          rect = [v frame];
+          // NOTE: I am not sure why this is needed, the row and column heights are coming in via the nib
```
fredkiefer, line 63: Float laues that small should be treated as zero. Most likely the programmer that implemented coding on Cocoa was not aware of this and instead of not encoding that value encoded it. The best we can do now is to add a small delta to our comparisons. Base is already doing this in a few places (See almostEqual in NSGeometry.m). You could copy the code from there, but if the simple "> 0.0" is sufficient maybe that is already good enough? I am a bit unsure here as this may depend on the specific CGFloat implementation.

gcasa: I think you're correct.  The values for these are not in the XIB file.  I'm guessing the XIB->NIB compiler is simply instantiating the objects and then using the non-keyed methods to save them and since they are not treated as zero by the non-keyed code it is simply dumping them out.   

I've attached the XIB file here: [MainMenu.xib.txt](https://github.com/gnustep/libs-gui/files/6014922/MainMenu.xib.txt)

AS AN ASIDE: I have toyed with the idea of doing a XIB->gorm "compiler".  The trick would be to create an NSWindow subclass that would allow you to add views and such in memory so that no GUI would appear on the display during compilation, but since we can read XIBs directly it's hardly a priority at this point.

gcasa: Another thought: It's confusing that they are not in the xib, since they are clearly something that the InterfaceBuilder in XCode is setting up...  please see this here: 

column_editing

If you notice, the column editor has "100" and the "explicit" flag is unchecked.   Similarly if I click on a row I get the value of "30" for height.  I believe these MIGHT be defaults, but I am not sure.  I am going to operate on the assumption that they are and if the method I wrote for finding a "prototype view" to base the height/width on, I will use that.     

Also, at this point, I may leave the "merge..." method unimplemented as I am not sure I understand yet how that will work.  I believe it may be just as simple as adding the two views together and just letting the math skip over the merged view (either vertically or horizontally as the case may be), but I am not sure.

Headers/AppKit/NSGridView.h
gcasa: This is done.  Everything is done in terms of the array in _cells.

Source/NSGridView.m
gcasa: done

gcasa:
I am going to move the changes for NSStackView to another branch as they are sufficiently complex.  I am not sure why I joined these two to begin with.

gcasa:
I am moving this to a non-draft PR as it is now functional and ready for review.

Source/NSGridView.m
```diff
@@ -0,0 +1,1280 @@
...
+        }
+    }
+  END_FOR_IN(_cells);
+  
+  return result;
+}
+
+- (NSArray *) _viewsForRowAtIndex: (NSUInteger)rowIndex
+{
+  NSArray *cells = [self _cellsForColumnAtIndex: rowIndex];
```
fredkiefer, line 62: Shouldn't that be _cellsForRowAtIndex?

Source/NSGridView.m
```diff
@@ -0,0 +1,1280 @@
...
+      if (rc > 0)
+        {
+          cc = [[rows objectAtIndex: 0] count];
+        }
+      
+      _cells = [[NSMutableArray alloc] initWithCapacity: rc * cc];
+      _rows = [[NSMutableArray alloc] initWithCapacity: rc];
+      _columns = [[NSMutableArray alloc] initWithCapacity: cc];
+
+      FOR_IN(NSArray*, row, rows)
```
fredkiefer, line 295: This whole code makes no sense to me. Could you please write some test code that exercises this method with more than one row and column?

Source/NSGridView.m
```diff
@@ -0,0 +1,1280 @@
...
+      for (c = 0; c < columnCount; c++)
+        {
+          NSGridCell *gc = [[NSGridCell alloc] init];
+          [col addObject: gc];
+          RELEASE(gc);
+        }
+      [rows addObject: col];
+    }
+
+  return AUTORELEASE([self gridViewWithViews: rows]);
```
fredkiefer, line 350: You add an array of arrays here where the elements are of the class NSGridCell. Shouldn't this be views?

Source/NSGridView.m
```diff
@@ -0,0 +1,1280 @@
...
+}
+
++ (instancetype) gridViewWithViews: (NSArray *)rows
+{
+  return [[self alloc] initWithViews: rows];
```
fredkiefer, line 355: Here the autorelease is missing. Please remember to remove it from -gridViewWithNumberOfColumns:rows:, when you add it here.

Source/NSGridView.m
```diff
@@ -0,0 +1,1280 @@
...
+
+  NSRect f = [self _findProtypeView];
+  NSUInteger i = 0;
+  NSUInteger pos = index * [self numberOfColumns];
+  FOR_IN(NSView*, v, views)
+    {
+      NSGridCell *c = [[NSGridCell alloc] init];
+      NSGridColumn *col = [_columns objectAtIndex: i];
+      
+      [col setGridView: self];
```
fredkiefer, line 457: Why are you setting anything on the already existing column here

Source/NSGridView.m
```diff
@@ -0,0 +1,1280 @@
...
+      [c _setOwningRow: gr];
+      [c _setOwningColumn: col];
+      [v setFrame: f];
+      [c setContentView: v];
+      RELEASE(v);
```
fredkiefer, line 462: This release looks wrong to me. Could you please explain why you think it is needed?

Source/NSGridView.m
```diff
@@ -0,0 +1,1280 @@
...
+      [_cells insertObject: c
+                   atIndex: pos + i];
+      RELEASE(c);
+      i++;
+    }
+  END_FOR_IN(views);
+
+  // Insert remaineder of cells for views not present..
```
fredkiefer, line 470: Typo

Source/NSGridView.m
```diff
@@ -0,0 +1,1280 @@
...
+  NSUInteger r = [self numberOfColumns] - i;
+  NSUInteger idx = 0;
+  pos += i;
+  for(idx = 0; idx < r; idx++)
+    {
+      NSGridCell *c = [[NSGridCell alloc] init];
```
fredkiefer, line 476: You need to set all the values on the cell you add in the loop above. This code would be easier if you change to first loop to be over all columns and get the view form views as long as the index is smaller. That way you only have one loop and everything should be consistent.

Source/NSGridView.m
```diff
@@ -0,0 +1,1280 @@
...
+{
+  return [self insertColumnAtIndex: [self numberOfColumns]
+                         withViews: views];
+}
+
+- (NSGridColumn *) insertColumnAtIndex: (NSInteger)index withViews: (NSArray *)views
+{
+  NSGridColumn *gc = [[NSGridColumn alloc] init];
+
+  // Insert the row and release...
```
fredkiefer, line 512: column

Source/NSGridView.m
```diff
@@ -0,0 +1,1280 @@
...
+                   forKey: @"NSGrid_rows"];
+      [coder encodeInteger: _xPlacement
+                    forKey: @"NSGrid_xPlacement"];
+      [coder encodeInteger: _yPlacement forKey: @"NSGrid_yPlacement"];
+      [coder encodeObject: _cells
+                   forKey: @"NSGrid_cells"];
+    }
+  else
+    {
+      [coder encodeValueOfObjCType:@encode(NSUInteger)
```
fredkiefer, line 649: Here and in the following lines spaces are missing.

Source/NSGridView.m
```diff
@@ -0,0 +1,1280 @@
...
+                                at:&_yPlacement];
+    }
+}
+
+- (instancetype) initWithCoder: (NSCoder *)coder
+{
+  self = [super initWithCoder: coder];
+  if (self != nil)
+    {
+      NSDebugLog(@"%@ %@",NSStringFromClass([self class]), NSStringFromSelector(_cmd));
```
fredkiefer, line 670: Before the merge most of these log statements need to be removed.

Source/NSGridView.m
```diff
@@ -0,0 +1,1280 @@
...
+    }
+  
+  // Refresh...
+  [self _refreshCells];
+  return gr;
+}
+
+- (void) moveRowAtIndex: (NSInteger)fromIndex toIndex: (NSInteger)toIndex
+{
+  NSArray *views = [self _viewsForRowAtIndex: fromIndex];
```
fredkiefer, line 487: This is a possible way to implement this method, but that way we loose all the information that was set on a specific cell.

Source/NSGridView.m
```diff
@@ -0,0 +1,1280 @@
...
+}
+
+- (NSGridView *) gridView
+{
+  return _gridView;
+}
+
+- (NSInteger) numberOfCells
+{
+  NSUInteger ci = [_gridView indexOfColumn: self];
```
fredkiefer, line 958: Don't all columns have the same amount of cells and that value is numberOfRows of the grid view?

Source/NSGridView.m
```diff
@@ -0,0 +1,1280 @@
...
+}
+
+- (BOOL) isHidden
+{
+  return _isHidden;
+}
+
+- (void) mergeCellsInRange: (NSRange)range
+{
+  NSLog(@"Method %@ unimplemented.", NSStringFromSelector(_cmd));
```
fredkiefer, line 1017: Shouldn't you be calling mergeCellsInHorizontalRange:verticalRange: on the grid view? I really don't know.

Source/NSGridView.m
```diff
@@ -0,0 +1,1280 @@
...
+    }
+}
+
+- (instancetype) initWithCoder: (NSCoder *)coder
+{
+  self = [super init];
+  if (self != nil)
+    {
+      NSDebugLog(@"%@ %@",NSStringFromClass([self class]), NSStringFromSelector(_cmd));
+      // _cells = [[NSMutableArray alloc] initWithCapacity: 10];
```
fredkiefer, line 1060: Please remove this line

Source/NSGridView.m
```diff
@@ -0,0 +1,1280 @@
...
+}
+
+- (NSGridView *) gridView
+{
+  return _gridView;
+}
+
+- (NSInteger) numberOfCells
+{
+  NSUInteger ri = [_gridView indexOfRow: self];
```
fredkiefer, line 1132: Same question as above

Source/NSGridView.m
```diff
@@ -0,0 +1,1280 @@
...
+{
+  NSUInteger ci = [_gridView indexOfColumn: self];
+  NSArray *arr = [_gridView _cellsForColumnAtIndex: ci];
+  return [arr count];
+}
+
+- (NSGridCell *) cellAtIndex:(NSInteger)index
+{
+  NSUInteger ci = [_gridView indexOfColumn: self];
+  NSArray *arr = [_gridView _cellsForColumnAtIndex: ci];
```
fredkiefer, line 966: You do not need this intermediate array, you could compute the index of the cell directly.

Headers/AppKit/NSGridView.h
```diff
@@ -0,0 +1,213 @@
...
+   License along with this library; if not, write to the Free
+   Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+   Boston, MA 02110 USA.
+*/
+
+#ifndef _NSGridView_h_GNUSTEP_GUI_INCLUDE
+#define _NSGridView_h_GNUSTEP_GUI_INCLUDE
+
+#import <AppKit/NSView.h>
+#import <AppKit/NSLayoutAnchor.h>
```
fredkiefer, line 29: I don't see where you require this import.

MISSING
```diff
@@ -12,12 +12,10 @@ MISSING HEADERS ( * = difficult, - = quick, + = placeholder )
 > NSDraggingSession.h -
 > NSFilePromiseProvider.h -
 > NSFilePromiseReceiver.h -
-> NSGlyphInfo.h
-> NSGridView.h *
+> NSGlyphInfo.h -
 > NSItemProvider.h + 
 > NSMenuToolbarItem.h -
-> NSOpenGLLayer.h
-> NSStackView.h *
```
fredkiefer, line 20: NSStackView.h was removed from this PR.

gcasa: Fixed

Headers/AppKit/NSGridView.h
gcasa: Removed NSLayoutAnchor.h

Source/NSGridView.m
gcasa: Modified to use -[NSGridView cellAtColumnIndex:rowIndex: I have left a comment in there which explains why it is done this way.

Source/NSGridView.m
gcasa: Done

Source/NSGridView.m
gcasa: Since they can be merged, I didn't want to rely on [NSGridView numberOfColumns/Rows] for this data.   While this way of doing it is not as fast it is more accurate if the cells are merged.  Currently the code doesn't support merging, but when it does I, hopefully, won't need to revisit these methods.

Source/NSGridView.m
gcasa: Done.  I believe I have this correct.

Source/NSGridView.m
gcasa: Since merging is possible some might have a different count.  That is why I did this this way.  I need to test on Cocoa/macOS to be sure.  Once I do I will make the appropriate changes here.

Source/NSGridView.m
gcasa: Corrected

Source/NSGridView.m
gcasa: Removed.  Only needed them while I was working on decoding.

Source/NSGridView.m
gcasa: Corrected.

Source/NSGridView.m
gcasa: FIxed

Source/NSGridView.m
gcasa: Removed... the views are already autoreleased.

Source/NSGridView.m
gcasa: Removed.

Source/NSGridView.m
fredkiefer: Please test on Cocoa. I think your argument is wrong and if it where true most methods in the current implementation of NSGridView would fail. To be consistent you should rely on the equal size here as well.

Source/NSGridView.m
fredkiefer: Again, take a look at cellAtColumnIndex:rowIndex:. This method would fail if your argument would be true.

fredkiefer:
We are really getting closer now!

Source/NSGridView.m
gcasa: Tested on macOS and changed implementation.

Source/NSGridView.m
gcasa: Tested on macOS... changed implementation.

Source/NSGridView.m
gcasa: Done

Source/NSGridView.m
gcasa: In the docs it is rows, but I will change it to views.

gcasa: Done

Source/NSGridView.m
```diff
@@ -0,0 +1,1262 @@
...
+      for (c = 0; c < columnCount; c++)
+        {
+          NSGridCell *gc = [[NSGridCell alloc] init];
+          [col addObject: gc];
+          RELEASE(gc);
+        }
+      [rows addObject: col];
+    }
+
+  return [self gridViewWithViews: rows];
```
fredkiefer, line 357: My comment here seems to be lost, so I add it again.
The method gridViewWithViews: expects an array of arrays of views. What you create here is an array of arrays pf NSGridCells. This is something completely different and won't work.

gcasa: Ah, thank you for pointing that out.  I have changed this to generate an array of arrays of NSViews and then call gridViewWIthViews:.   Sorry if this seems to have been lost before.

Source/NSGridView.m
gcasa: Fixed.  The methods now deal with the cells directly.

Source/GSXib5KeyedUnarchiver.m
gcasa: This has been done.

Source/NSGridView.m
gcasa: Corrected.

gcasa:
@fredkiefer I believe we are at a Release Candidate level right now.

Source/GSXib5KeyedUnarchiver.m
fredkiefer: Are you going to work on this or leave it that way?

Source/NSGridView.m
```diff
@@ -0,0 +1,1232 @@
...
+          // NOTE: I am not sure why this is needed, the row and column heights are coming in via the nib
+          // as VERY small numbers (1.175... e-38) (see PR for this work, I have attached the XML nib there)
+          // so it is necessary to figure out the width of each column so that things can be properly spaced.
+          if (ci == 0)
+            {
+              current_y -= p.size.height; // rect.size.height;
+              current_x = 0.0;
+            }
+          
+          current_y -= [c yPlacement] - [row topPadding];
```
fredkiefer, line 201: Shouldn't topPadding be added here? You subtract from current_y here an currently it seems to me you do it wrongly. The same is true for height further down.

Source/NSGridView.m
```diff
@@ -0,0 +1,1232 @@
...
+          current_x += [c xPlacement] + [col leadingPadding] + [col width];
```
fredkiefer, line 202: I think width should not be included here.

Source/NSGridView.m
```diff
@@ -0,0 +1,1232 @@
...
+          if (v != nil)
+            {
+              rect = p;
+              rect.origin.x = current_x;
+              rect.origin.y = current_y;
+              [v setFrame: rect];
+              [self addSubview: v];
+            }
+          
+          current_x += [col trailingPadding] + [col width] + p.size.width + _columnSpacing;
```
fredkiefer, line 213: Having both [col width]  and p.size.width here seems to be too much. Did you ever run this method and had some values output from it?

Source/NSGridView.m
```diff
@@ -0,0 +1,1232 @@
...
+  if (self != nil)
+    {
+      NSUInteger rc = [views count], cc = 0;
+
+      if (rc > 0)
+        {
+          cc = [[views objectAtIndex: 0] count];
+        }
+      
+      _cells = [[NSMutableArray alloc] initWithCapacity: rc * cc];
```
fredkiefer, line 255: These instance variables got already set in initWithFrame:.

Source/NSGridView.m
```diff
@@ -0,0 +1,1232 @@
...
+      _rows = [[NSMutableArray alloc] initWithCapacity: rc];
+      _columns = [[NSMutableArray alloc] initWithCapacity: cc];
+
+      FOR_IN(NSArray*, row, views)
+        {
+          NSGridRow *gr = [[NSGridRow alloc] init];
+
+          [gr setGridView: self];
+          [_rows addObject: gr];
```
fredkiefer, line 264: You should release gr here.

Source/NSGridView.m
```diff
@@ -0,0 +1,1232 @@
...
+          c = 0;
+          FOR_IN(NSView*, v, row)
+            {
+              NSGridColumn *gc = nil; 
+              NSGridCell *cell = [[NSGridCell alloc] init];
+
+              if (r == 0) // first row, create all of the columns...
+                {
+                  gc = [[NSGridColumn alloc] init];
+                  [_columns addObject: gc];
```
fredkiefer, line 275: You should release gc here.

Source/NSGridView.m
```diff
@@ -0,0 +1,1232 @@
...
+  self = [super initWithFrame: frameRect];
+
+  if (self != nil)
+    {
+      _rows = [[NSMutableArray alloc] init];
+      _columns = [[NSMutableArray alloc] init];
+      _cells = [[NSMutableArray alloc] init];
+    }
+
+  [self _refreshCells];
```
fredkiefer, line 234: Why this call? You do not have any cells here.

Source/NSGridView.m
```diff
@@ -0,0 +1,1232 @@
...
+  NSGridRow *gr = [[NSGridRow alloc] init];
+
+  // Insert the row and release...
+  [_rows insertObject: gr atIndex: index];
+  [gr setGridView: self];
+  RELEASE(gr);
+
+  NSUInteger i = 0;
+  NSUInteger pos = index * [self numberOfColumns];
+  for(i = 0; i < [self numberOfColumns]; i++)
```
fredkiefer, line 402: A space is missing here

Source/NSGridView.m
```diff
@@ -0,0 +1,1232 @@
...
+    {
+      NSGridCell *c = [[NSGridCell alloc] init];
```
fredkiefer, line 404: You should not create a cell in this line.

Source/NSGridView.m
```diff
@@ -0,0 +1,1232 @@
...
+{
+  NSRect f = [self _prototypeFrame];
+  NSMutableArray *cells = [NSMutableArray arrayWithCapacity: [views count]];
+  FOR_IN(NSView*, v, views)
+    {
+      NSGridCell *c = [[NSGridCell alloc] init];
+
+      [v setFrame: f];
+      [c setContentView: v];
+      [cells addObject: c];
```
fredkiefer, line 438: You need to release c here.

Source/NSGridView.m
```diff
@@ -0,0 +1,1232 @@
...
+  NSGridColumn *gc = [[NSGridColumn alloc] init];
+
+  // Insert the row and release...
+  [_columns insertObject: gc atIndex: index];
+  [gc setGridView: self];
+  RELEASE(gc);
+
+  NSUInteger i = 0;
+  NSUInteger pos = index;
+  for(i = 0; i < [self numberOfRows]; i++)
```
fredkiefer, line 478: Space missing

Source/NSGridView.m
```diff
@@ -0,0 +1,1232 @@
...
+    {
+      NSGridCell *c = nil;
+      NSGridRow *row = [_rows objectAtIndex: i];
+
+      if (i > ([cells count] - 1))
+        {
+          c = [[NSGridCell alloc] init];
```
fredkiefer, line 485: You need to (auto)release this cell.

Source/NSGridView.m
```diff
@@ -0,0 +1,1232 @@
...
+  NSUInteger i = 0;
+  NSUInteger pos = index * [self numberOfColumns];
+  for(i = 0; i < [self numberOfColumns]; i++)
+    {
+      NSGridCell *c = [[NSGridCell alloc] init];
+      NSGridColumn *col = [_columns objectAtIndex: i];
+
+      if (i > ([cells count] - 1))
+        {
+          c = [[NSGridCell alloc] init];
```
fredkiefer, line 409: You need to (auto)release this cell.

Source/NSGridView.m
```diff
@@ -0,0 +1,1232 @@
...
+{
+  NSRect f = [self _prototypeFrame];
+  NSMutableArray *cells = [NSMutableArray arrayWithCapacity: [views count]];
+  FOR_IN(NSView*, v, views)
+    {
+      NSGridCell *c = [[NSGridCell alloc] init];
+
+      [v setFrame: f];
+      [c setContentView: v];
+      [cells addObject: c];
```
fredkiefer, line 513: You need to release c.

gcasa: fixed.

Source/NSGridView.m
gcasa: done

Source/NSGridView.m
gcasa: done

Source/NSGridView.m
gcasa: done

Source/NSGridView.m
gcasa: done

Source/NSGridView.m
gcasa: removed

Source/NSGridView.m
gcasa: Added

Source/NSGridView.m
gcasa: Removed

Source/NSGridView.m
gcasa: Done

Source/NSGridView.m
gcasa: Done

Source/NSGridView.m
gcasa: Done

Source/NSGridView.m
gcasa: I believe p.size.width is the correct thing to use here.

Source/NSGridView.m
gcasa: Removed.

Source/NSGridView.m
gcasa: Should this be added?

gcasa: Took them out.  I tried adding and it caused it to do a "staircase" layout (each column slightly lower).

Source/NSGridView.m
```diff
@@ -0,0 +1,1233 @@
...
+          NSGridRow *row = [self rowAtIndex: ri];
+          NSGridColumn *col = [self columnAtIndex: ci];
+          NSRect rect = NSZeroRect;
+          
+          if (v != nil)
+            {
+              rect = [v frame];
+            }
+          
+          // Do the math for the frame...
```
fredkiefer, line 191: Not sure whether this comment is still related to the code below.

Source/NSGridView.m
```diff
@@ -0,0 +1,1233 @@
...
+            {
+              rect = p;
+              rect.origin.x = current_x;
+              rect.origin.y = current_y;
+              [v setFrame: rect];
+              [self addSubview: v];
+            }
+          
+          current_x += [col trailingPadding] + p.size.width + _columnSpacing;
+          current_y -= [row bottomPadding];// + _rowSpacing;
```
fredkiefer, line 214: I think changes to y should only happen within the ci == 0 clause.

Source/NSGridView.m
```diff
@@ -0,0 +1,1233 @@
...
+}
+
+- (instancetype) initWithViews: (NSArray *)views // an array of arrays
+{
+  self = [self initWithFrame: NSZeroRect];
+
+  NSUInteger c = 0;
+  NSUInteger r = 0;
+
+  if (self != nil)
```
fredkiefer, line 244: This code would be easier to read if you moved the check for the case of self being nil into line 240 and just returned nil there.

Source/NSGridView.m
```diff
@@ -0,0 +1,1233 @@
...
+
+// coding
+- (void) encodeWithCoder: (NSCoder *)coder
+{
+  [super encodeWithCoder: coder];
+  if ([coder allowsKeyedCoding])
+    {
+      [coder encodeInteger: _rowAlignment
+                    forKey: @"NSGrid_alignment"];
+      [coder encodeFloat: _columnSpacing
```
fredkiefer, line 600: Why float and not double?

Source/NSGridView.m
```diff
@@ -0,0 +1,1233 @@
...
+}
+
+// coding
+- (void) encodeWithCoder: (NSCoder *)coder
+{
+  if ([coder allowsKeyedCoding])
+    {
+      [coder encodeBool: _isHidden
+                 forKey: @"NSGrid_hidden"];
+      [coder encodeFloat: _leadingPadding
```
fredkiefer, line 980: Again, why float an not double?

Source/NSGridView.m
```diff
@@ -0,0 +1,1233 @@
...
+    {
+      return;
+    }
+  else
+    {
+      NSUInteger i = 0;
+      NSUInteger num_col = [self numberOfColumns];
+      NSRect f = [self frame];
+      CGFloat current_x = 0.0, current_y = f.size.height;
+      NSRect p = [self _prototypeFrame];
```
fredkiefer, line 169: The way you compute the prototype frame does not leave any space for padding. This calculation here can never add up.

Source/NSGridView.m
```diff
@@ -0,0 +1,1233 @@
...
+          // NOTE: I am not sure why this is needed, the row and column heights are coming in via the nib
+          // as VERY small numbers (1.175... e-38) (see PR for this work, I have attached the XML nib there)
+          // so it is necessary to figure out the width of each column so that things can be properly spaced.
+          if (ci == 0)
+            {
+              current_y -= p.size.height;
+              current_x = 0.0;
+            }
+          
+          current_y -= [c yPlacement] + [row topPadding];
```
fredkiefer, line 201: The values in placement are about alignment not absolute values.

Source/NSGridView.m
```diff
@@ -0,0 +1,1233 @@
...
+   Lesser General Public License for more details.
+   
+   You should have received a copy of the GNU Lesser General Public
+   License along with this library; if not, write to the Free
+   Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+   Boston, MA 02110 USA.
+*/
+
+#import <Foundation/NSArray.h>
+#import "AppKit/NSGridView.h"
```
fredkiefer, line 26: I do not see any drawRect: methods in this file, how is this supposed to work out?

Source/GSXib5KeyedUnarchiver.m
gcasa: I am not sure if it's really a clean mapping.   It's not just a matter of putting the class name in front like so @"%@_%@", className, attributeName...   the attribute in the XIB is alignment, but the one in the nib is rowAlignment.   Granted there are not that many exceptions, but I was going to ask you if this seemed worth it.  If you think it is, I will do it.

Source/NSGridView.m
gcasa: Ah, good point.

Source/NSGridView.m
gcasa: Done

Source/NSGridView.m
gcasa: fixed

Source/NSGridView.m
gcasa: Fixed

Source/NSGridView.m
gcasa: I should add the x padding and y padding in that method so that the calculation is better.

Source/GSXib5KeyedUnarchiver.m
fredkiefer: OK, so we leave it for now.

Source/NSGridView.m
fredkiefer: OK, this "works" because you add the cell views as subviews of the grid view in _refreshCells. This is a rather simplistic way of implementing this. That won't allow you to have cell merging and it makes padding and alignment of cells rather hard. Are you sure to go ahead with that implementation?

gcasa:
How else can i have functional views within the gridview?   What
alternative do you have in mind?

Source/NSGridView.m
gcasa: How else can I have functional views within the NSGridView?   What alternate implementation do you have in mind?  It is worth mentioning that when you list the subviews of the NSGridView that they are the actual views in the cells.   I would guess that suggests that my implementation is similar to the one done by Apple.

Source/GSXib5KeyedUnarchiver.m
fredkiefer: OK

Source/NSGridView.m
fredkiefer: Seeing this views as subview of the NSGridView is a good argument. I am surprised they did it that way. I expected that all the magic would happen in the drawRect: method. This has a few consequences I need to think about. The first being that you need to handle view removal. At the moment all cell views ever added remain as subviews of the NSGridView.
The next is the frame calculation for the views. Most likely you should start from your _prototypeFrame position that into the row/column and then subtract all the different padding values according to the placement.

Source/NSGridView.m
fredkiefer: No, rather you need to subtract these values before you set the frame of the cell view.

Source/NSGridView.m
```diff
@@ -0,0 +1,1233 @@
...
+}
+
+- (void) setColumnSpacing: (CGFloat)f
+{
+  _columnSpacing = f;
+}
+  
+- (void) mergeCellsInHorizontalRange: (NSRange)hRange verticalRange: (NSRange)vRange
+{
+  NSLog(@"Method %@ unimplemented.", NSStringFromSelector(_cmd));
```
fredkiefer, line 589: I think what needs to be done here is to set the _mergeHead of all the affected cells (sparing the first one). The interesting question is how this should show up in view drawing. Most likely views where the cell is merged should not be displayed and the first cell of such an area should cover the whole range.

Source/NSGridView.m
gcasa: Just to document this in the PR.  The test for this class is here.. [https://github.com/gcasa/NSGridView_test.git](https://github.com/gcasa/NSGridView_test.git)

2021-03-26 02:24:57.844359-0400 NSGridView_test[62461:8943425] NSGridView has 13 subviews.
2021-03-26 02:24:57.844463-0400 NSGridView_test[62461:8943425] NSGridViewSizeForContent = 0.000000
2021-03-26 02:24:57.844530-0400 NSGridView_test[62461:8943425] Nunber of cells in row 1 = 9
2021-03-26 02:24:57.844860-0400 NSGridView_test[62461:8943425] Subviews = (
    "<NSButton: 0x7f81bb707aa0>",
    "<NSButton: 0x7f81bb40a6e0>",
    "<NSTextField: 0x7f81bb40b590>",
    "<NSColorWell: 0x7f81bb708b50>",
    "<NSStepper: 0x7f81bb709c80>",
    "<NSSwitch: 0x7f81bb70a8f0>",
    "<NSLevelIndicator: 0x7f81bb70b6c0>",
    "<NSSlider: 0x7f81bb523020>",
    "<NSView: 0x7f81bb40e4d0>",
    "<NSButton: 0x7f81bb40e9b0>",
    "<NSPopUpButton: 0x7f81bb40fa30>",
    "<NSButton: 0x7f81bb720e30>",
    "<NSButton: 0x7f81bc908d80>"
)

Source/NSGridView.m
gcasa: removed this.

Source/NSGridView.m
gcasa: Removed

gcasa:
Just to illustrate some weirdness going on with [self.gridView mergeCellsInHorizontalRange:NSMakeRange(0, 2) verticalRange:NSMakeRange(0, 2)];

weirdness

Source/NSGridView.m
gcasa: Done.

Source/NSGridView.m
gcasa: I won't be implementing merging right now.  I need to understand how it works as it is doing funny things (I attached an example in a comment).

gcasa:
Fred, I am going to merge this unless there is an objection.

PR #81, opened 2020-09-01
Fix issues found in GNUstep specific to running on Windows/WIN32 platforms.

gnustep-admin:
This branch is to fix windows specific problems in GUI.  I will do my best to provide a detailed explanation of why I am making each change.

### DIFF
--- a/Source/GSToolTips.m
+++ b/Source/GSToolTips.m
@@ -156,7 +156,10 @@
       NSRect frame = [self frame];
       NSRect textRect = NSInsetRect(frame, 2, 2);
 
+      [[NSColor whiteColor] set];
+      NSRectFill(bounds);
       NSDrawColorTiledRects(bounds, bounds, sides, colors, 4);
+      [[NSColor blackColor] set];
       [_text drawInRect: textRect];
     }
 }
@@ -182,6 +185,7 @@
                               defer: flag];
   if (self)
     {
+      [self setBackgroundColor: [NSColor whiteColor]];
       [self setContentView: [[[GSTTView alloc] initWithFrame: contentRect] autorelease]];
     }
   return self;
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,13 @@
+2020-09-12 Gregory John Casamento <greg.casamento@gmail.com>
+
+	* Source/GSToolTips.m: Fix issue where setting the theme causes
+	tooltips to become all black.  This is happening on windows and other
+	platforms.
+	* Source/NSCell.m: Fix for issue on Windows were contents are reset
+	when initWithCoder: coder is called.  If the contents are set, they
+	determine the type.  Currently investigating why this is occuring
+	ONLY on Windows.
+
 2020-09-12 Fred Kiefer <FredKiefer@gmx.de>
 
 	* Source/NSTextView.m (-initWithCoder:): Make sure text container is
--- a/Source/NSCell.m
+++ b/Source/NSCell.m
@@ -2647,7 +2647,15 @@ static NSColor *dtxtCol;
           [self setSelectable: ((cFlags & 0x200000) == 0x200000)];
           [self setBezeled: ((cFlags & 0x400000) == 0x400000)];
           [self setBordered: ((cFlags & 0x800000) == 0x800000)];
-          [self setType: ((cFlags & 0xC000000) >> 26)];
+	  if (contents == nil)
+	    {
+	      //
+	      // If the contents aren't set (the contents determine the type), 
+	      // get it from the flags.  This prevents the type from being accidentally
+	      // reset on some platforms (mainly WIN32) after the contents are set.
+	      //
+	      [self setType: ((cFlags & 0xC000000) >> 26)];
+	    }
           [self setEditable: ((cFlags & 0x10000000) == 0x10000000)];
           // This bit flag is the other way around!
           [self setEnabled: ((cFlags & 0x20000000) != 0x20000000)];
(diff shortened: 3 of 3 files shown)


gnustep-admin:
> I think this analysis is only partly correct. The interesting question would be, why the value that we send to setType: here does not match the already set one? (setType ignores setting the same value again) This would mean that within XIB decoding we compute a wrong value for the flags that go into NSCellFlags. It surely would be more worthwhile to fix that place instead of working around here.

If that's the case, then why is it working on Linux?   That's what I can't figure out.  Why doesn't this happen anywhere else?

fredkiefer:
> If that's the case, then why is it working on Linux? That's what I can't figure out. Why doesn't this happen anywhere else?

That is the question, regardless of what the fix should be. To answer that we really need to investigate the values passed on to setType:. I currently don't have the time to look into this. Maybe on the weekend. But the best approach here is really to dig into the conversion done in the XIB decoding.

fredkiefer:
I added a few NSLog statements to NSCell setType: and the GSXib5KeyedUnarchiver method decodeCellFlags1ForElement: and was not able to reproduce this behaviour with your NSStoryboard_test on Linux. I would have expected to see at least a duplicated setType: call with a different value than the current one.
My impression is that there is difference in the compiler that causes this strange behaviour. But as I said previously I don't have an MS Windows environment and would only set up one if somebody pays me for that. This means somebody else needs to investigate the issue.

gcasa:
> I added a few NSLog statements to NSCell setType: and the GSXib5KeyedUnarchiver method decodeCellFlags1ForElement: and was not able to reproduce this behaviour with your NSStoryboard_test on Linux. I would have expected to see at least a duplicated setType: call with a different value than the current one.
> My impression is that there is difference in the compiler that causes this strange behaviour. But as I said previously I don't have an MS Windows environment and would only set up one if somebody pays me for that. This means somebody else needs to investigate the issue.

This is a WOW moment for me.  I don't know how to address this.   A *temporary* and I stress *TEMPORARY* because I don't like it and it's not a perfect solution is to put #ifdefs around the change I made and leave a long comment about this issue as it is truly strange.   What *could* be happening is that some optimization is done by the compiler that is rearranging the order of the calls here for whatever reason (this is a CLUELESS GUESS).   I will test that by disabling optimization and seeing what results.

fredkiefer:
> This is a WOW moment for me. I don't know how to address this. A _temporary_ and I stress _TEMPORARY_ because I don't like it and it's not a perfect solution is to put #ifdefs around the change I made and leave a long comment about this issue as it is truly strange. What _could_ be happening is that some optimization is done by the compiler that is rearranging the order of the calls here for whatever reason (this is a CLUELESS GUESS). I will test that by disabling optimization and seeing what results.

Did you try with a few log statements on Windows? A guess should always be the last resort after going through the other available information sources first.

gcasa:
> > This is a WOW moment for me. I don't know how to address this. A _temporary_ and I stress _TEMPORARY_ because I don't like it and it's not a perfect solution is to put #ifdefs around the change I made and leave a long comment about this issue as it is truly strange. What _could_ be happening is that some optimization is done by the compiler that is rearranging the order of the calls here for whatever reason (this is a CLUELESS GUESS). I will test that by disabling optimization and seeing what results.
> 
> Did you try with a few log statements on Windows? A guess should always be the last resort after going through the other available information sources first.

I did.   It showed that setType: was being called after setting the contents.  I also ran in the debugger.   For some reason, the type being set was different than the contents would demand.  The only thing I can think of is that there is some difference in how the flags are being set on Windows vs on Linux.   The funny thing is that this happens on no other platform.  I am going to dive back into it this afternoon and, hopefully, I will have a little more data for us to consider.

fredkiefer:
The interesting bit would be the old and the new value of type at the start of setType:. And add the end of decodeCellFlags1ForElement: the value of mask.value and especially that of (mask.value &  0xC000000) >> 26) as the should be the type.

Source/GSToolTips.m
```diff
@@ -156,7 +156,10 @@ - (void) drawRect: (NSRect)dirtyRect
       NSRect frame = [self frame];
       NSRect textRect = NSInsetRect(frame, 2, 2);
 
+      [[NSColor whiteColor] set];
```
fredkiefer, line 159: This should use the background colour you set below.

Source/GSToolTips.m
```diff
@@ -156,7 +156,10 @@ - (void) drawRect: (NSRect)dirtyRect
...
+      NSRectFill(bounds);
       NSDrawColorTiledRects(bounds, bounds, sides, colors, 4);
+      [[NSColor blackColor] set];
```
fredkiefer, line 162: This should use the text colour from NSColor

Source/GSToolTips.m
```diff
@@ -182,6 +185,7 @@ - (id) initWithContentRect: (NSRect)contentRect
                               defer: flag];
   if (self)
     {
+      [self setBackgroundColor: [NSColor whiteColor]];
```
fredkiefer, line 188: Again a colour from NSColor should be used here

fredkiefer (changes requested):
I am strongly against merging all these changes as they are just workarounds without addressing the real issues.

fredkiefer:
Why did you merge this PR? You did not address any of the comments I made and in my last comment I strongly advised against merging. If you want to take over the maintainership of GNUstep gui just say so. But as long as I am the maintainer you should at least listen to my  comments and if you do not agree say so. Nobody else in the GNUstep community would behave that way, why do you?
I will revert this commit and we try again.

gcasa:
> Why did you merge this PR? You did not address any of the comments I made and in my last comment I strongly advised against merging. If you want to take over the maintainership of GNUstep gui just say so. But as long as I am the maintainer you should at least listen to my comments and if you do not agree say so. Nobody else in the GNUstep community would behave that way, why do you?
> I will revert this commit and we try again.

My apologies, you're correct.   I will address the issues immediately.

PR #82, opened 2020-09-11
NSWindow input focus switch fix

trunkmaster:
Unhide/activate application after ordering and making it key. This prevents focus flickering between current key window and this window during application activation.  Do not try to set key and main if window can't be set as key.


fredkiefer (approved):
Looks OK, even though the order of a few calls did change

PR #85, opened 2020-10-04
Protect against theme colour list missing out on default colours

fredkiefer:
As I am not actually using themes, somebody doing so should check and test this change.

### DIFF
--- a/Source/NSColor.m
+++ b/Source/NSColor.m
@@ -2046,6 +2046,7 @@ systemColorWithName(NSString *name)
   NSColorList	*list = [theme colors];
   NSEnumerator	*enumerator;
   NSString	*name;
+  NSString	*key;
 
   if (list == nil)
     {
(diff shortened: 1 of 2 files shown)


Source/NSColor.m
```diff
@@ -2065,6 +2066,22 @@ + (void) themeDidActivate: (NSNotification*)notification
     {
       [[systemDict objectForKey: name] recache];
     }
+
+  if (list != defaultSystemColors)
+    {
+      // Check that all default colours are definied in the theme
+      enumerator = [colorStrings keyEnumerator];
+      while ((key = [enumerator nextObject]) != nil)
```
gcasa, line 2074: Why not use fast enumeration here?  Just wondering.

gcasa (approved):
Approved.  Thank you.

rfm:
This patch makes sense to me (though I haven't actually tried it, as my GUI is not currently working, most likely because I'm playing with David's latest libobj which has linker issues).

As for fast enumeration (which I love), I was not aware (or maybe I once knew but had forgotten) that we had decided to move to requiring compilers that supported it.
Assuming we are retaining support older compilers (I don't actually know when fast enumeration was introducd):  I like the fast enumeration syntax as well as performance, but in most situations (like this) neither is important enough for me to want to use preprocessor directives to implement support for both normal enumeration and fast enumeration.

fredkiefer:
> As for fast enumeration (which I love), I was not aware (or maybe I once knew but had forgotten) that we had decided to move to requiring compilers that supported it.

We already have a set of macros that allow us to use fast enumeration when the compiler supports it and fall back to the old code if it doesn't. I had not used it here as the old code around it doesn't and I wanted to stay close to that code, but there is no reason for not changing over.

But before merging I would like a definite answer from Greg whether this resolves the issue he had with tool tips in themes.

gcasa:
I tested it and it works.

PR #86, opened 2020-10-06
NSDictionaryController branch

gcasa:


gcasa:
> Why not NSExcludedKeysBinding?

The dangers of coding while tired.  I fixed the spelling.

gcasa:
I have an example that tests bindings.  It uses NSTableView.  One is NSArrayController_test and the other is NSDictionaryController_test.  NSArrayController_test works on macOS10.15 but doesn't function on GNUstep at all.  I believe the issue is in NSTableView.

fredkiefer:
> I have an example that tests bindings. It uses NSTableView. One is NSArrayController_test and the other is NSDictionaryController_test. NSArrayController_test works on macOS10.15 but doesn't function on GNUstep at all. I believe the issue is in NSTableView.

"doesn't function" is not a very helpful description. What happens and what should be happening instead? There are quite a few missing features in NSTableView and as for NSArrayController I only got that working with a horrible hack, I wrote about that last winter. You are welcome to improve on both classes.

gcasa:
> > I have an example that tests bindings. It uses NSTableView. One is NSArrayController_test and the other is NSDictionaryController_test. NSArrayController_test works on macOS10.15 but doesn't function on GNUstep at all. I believe the issue is in NSTableView.
> 
> "doesn't function" is not a very helpful description. What happens and what should be happening instead? There are quite a few missing features in NSTableView and as for NSArrayController I only got that working with a horrible hack, I wrote about that last winter. You are welcome to improve on both classes.

I will work on both of them.   I saw your comment about the hack.  I wrote a bug on this:
https://github.com/gnustep/libs-gui/issues/87

Headers/AppKit/NSKeyValueBinding.h
```diff
@@ -86,10 +86,16 @@
 
 @end
 
-// Keys in options dictionary
+// typedefs
+typedef NSString* NSBindingName;
```
fredkiefer, line 90: Now you define this type twice, it is already on line 95. BTW, the placement of the star is correct on line 95 and incorrect here.

Source/externs.m
```diff
@@ -754,14 +754,21 @@
 NSString *NSAlignmentBinding = @"alignment";
```
fredkiefer, line 754: Shouldn't we change the types here as well?

gcasa:
This PR uses a branch that has been superseded.

PR #90, opened 2021-01-08
Change encoding/decoding to protect against differences in integer size.

fredkiefer:
I went through most of the cell classes in base and fixed the encoding/decoding there. Of course there will be issues in the other classes as well, but I wanted your opinion on these changes before going ahead.

### DIFF
--- a/Source/GSGuiPrivate.h
+++ b/Source/GSGuiPrivate.h
@@ -31,6 +31,7 @@
 #define _GNUstep_H_GSGuiPrivate
 
 #import <Foundation/NSBundle.h>
+#import <Foundation/NSCoder.h>
 #include "GNUstepBase/GSConfig.h"
 #include <math.h>
 
@@ -102,6 +103,25 @@ static inline CGFloat GSRoundTowardsNegativeInfinity(CGFloat x)
 {
   return ceil(x - 0.5);
 }
+
+#if !defined(GS_DECODER)
+#define GS_DECODER(type) \
+static inline void decode_##type(NSCoder *coder, type *value) \
+{ \
+[coder decodeValueOfObjCType: @encode(type) at: value]; \
+}
+
+#define GS_ENCODER(type) \
+static inline void encode_##type(NSCoder *coder, type *value) \
+{ \
+[coder encodeValueOfObjCType: @encode(type) at: value]; \
+}
+#endif
+
+GS_ENCODER(NSUInteger)
+GS_DECODER(NSUInteger)
+GS_ENCODER(NSInteger)
+GS_DECODER(NSInteger)
 
 #endif /* _GNUstep_H_GSGuiPrivate */
 
--- a/Source/NSActionCell.m
+++ b/Source/NSActionCell.m
@@ -31,6 +31,7 @@
 
 #import "AppKit/NSActionCell.h"
 #import "AppKit/NSControl.h"
+#import "GSGuiPrivate.h"
 
 @implementation NSActionCell
 
@@ -386,10 +387,7 @@ static Class controlClass;
     }
   else
     {
-      int32_t tmp;
-
-      tmp = _tag;
-      [aCoder encodeValueOfObjCType: @encode(int32_t) at: &tmp];
+      encode_NSInteger(aCoder, &_tag);
       [aCoder encodeConditionalObject: _target];
       [aCoder encodeValueOfObjCType: @encode(SEL) at: &_action];
       // This is only encoded for backward compatibility and won't be decoded.
@@ -422,10 +420,8 @@ static Class controlClass;
   else
     {
       id dummy;
-      int32_t tmp;
 
-      [aDecoder decodeValueOfObjCType: @encode(int32_t) at: &tmp];
-      _tag = tmp;
+      decode_NSInteger(aDecoder, &_tag);
       _target = [aDecoder decodeObject];
       [aDecoder decodeValueOfObjCType: @encode(SEL) at: &_action];
       // Don't decode _control_view, as this may no longer be valid.
--- a/Source/NSActionCell.m
+++ b/Source/NSActionCell.m
@@ -386,7 +386,10 @@ static Class controlClass;
     }
   else
     {
-      [aCoder encodeValueOfObjCType: @encode(NSInteger) at: &_tag];
+      int32_t tmp;
+
+      tmp = _tag;
+      [aCoder encodeValueOfObjCType: @encode(int32_t) at: &tmp];
       [aCoder encodeConditionalObject: _target];
       [aCoder encodeValueOfObjCType: @encode(SEL) at: &_action];
       // This is only encoded for backward compatibility and won't be decoded.
@@ -419,12 +422,15 @@ static Class controlClass;
   else
     {
       id dummy;
+      int32_t tmp;
 
-      [aDecoder decodeValueOfObjCType: @encode(NSInteger) at: &_tag];
+      [aDecoder decodeValueOfObjCType: @encode(int32_t) at: &tmp];
+      _tag = tmp;
       _target = [aDecoder decodeObject];
       [aDecoder decodeValueOfObjCType: @encode(SEL) at: &_action];
       // Don't decode _control_view, as this may no longer be valid.
-      dummy = [aDecoder decodeObject];
+      [aDecoder decodeValueOfObjCType: @encode(id) at: &dummy];
+      RELEASE(dummy);
     }
 
   return self;
--- a/Source/NSButtonCell.m
+++ b/Source/NSButtonCell.m
@@ -1695,7 +1695,7 @@
   else
     {
       BOOL tmp;
-      uint32_t tmp2;
+      NSUInteger tmp2;
 
       [aCoder encodeObject: _keyEquivalent];
       [aCoder encodeObject: _keyEquivalentFont];
@@ -1705,16 +1705,11 @@
       [aCoder encodeValueOfObjCType: @encode(BOOL)
               at: &tmp];
 
-      tmp2 = _keyEquivalentModifierMask;
-      [aCoder encodeValueOfObjCType: @encode(uint32_t)
-                                 at: &tmp2];
-
+      encode_NSUInteger(aCoder, &_keyEquivalentModifierMask);
       tmp2 = _highlightsByMask;
-      [aCoder encodeValueOfObjCType: @encode(uint32_t)
-              at: &tmp2];
+      encode_NSUInteger(aCoder, &tmp2);
       tmp2 = _showAltStateMask;
-      [aCoder encodeValueOfObjCType: @encode(unsigned int)
-              at: &tmp2];
+      encode_NSUInteger(aCoder, &tmp2);
 
       [aCoder encodeObject: _sound];
       [aCoder encodeObject: _backgroundColor];
@@ -1723,11 +1718,9 @@
       [aCoder encodeValueOfObjCType: @encode(float)
                                  at: &_repeatInterval];
       tmp2 = _bezel_style;
-      [aCoder encodeValueOfObjCType: @encode(uint32_t)
-                                 at: &tmp2];
+      encode_NSUInteger(aCoder, &tmp2);
       tmp2 = _gradient_type;
-      [aCoder encodeValueOfObjCType: @encode(uint32_t)
-                                 at: &tmp2];
+      encode_NSUInteger(aCoder, &tmp2);
       tmp = _image_dims_when_disabled;
       [aCoder encodeValueOfObjCType: @encode(BOOL)
                                  at: &tmp];
@@ -1880,7 +1873,7 @@
   else
     {
       BOOL tmp;
-      uint32_t tmp2;
+      NSUInteger tmp2;
       int version = [aDecoder versionForClassName: @"NSButtonCell"];
       NSString *key = nil;
 
@@ -1897,18 +1890,14 @@
       [aDecoder decodeValueOfObjCType: @encode(id) at: &_altImage];
       [aDecoder decodeValueOfObjCType: @encode(BOOL) at: &tmp];
       _buttoncell_is_transparent = tmp;
-      [aDecoder decodeValueOfObjCType: @encode(uint32_t)
-                                   at: &tmp2];
-      _keyEquivalentModifierMask = (NSUInteger)tmp2;
+      decode_NSUInteger(aDecoder, &_keyEquivalentModifierMask);
       if (version <= 2)
         {
           _keyEquivalentModifierMask = _keyEquivalentModifierMask << 16;
         }
-      [aDecoder decodeValueOfObjCType: @encode(uint32_t)
-                                   at: &tmp2];
+      decode_NSUInteger(aDecoder, &tmp2);
       _highlightsByMask = (NSInteger)tmp2;
-      [aDecoder decodeValueOfObjCType: @encode(uint32_t)
-                                   at: &tmp2];
+      decode_NSUInteger(aDecoder, &tmp2);
       _showAltStateMask = (NSInteger)tmp2;
 
       if (version >= 2)
@@ -1917,11 +1906,9 @@
           [aDecoder decodeValueOfObjCType: @encode(id) at: &_backgroundColor];
           [aDecoder decodeValueOfObjCType: @encode(float) at: &_delayInterval];
           [aDecoder decodeValueOfObjCType: @encode(float) at: &_repeatInterval];
-          [aDecoder decodeValueOfObjCType: @encode(uint32_t)
-                                       at: &tmp2];
+          decode_NSUInteger(aDecoder, &tmp2);
           _bezel_style = (NSBezelStyle)tmp2;
-          [aDecoder decodeValueOfObjCType: @encode(uint32_t)
-                                       at: &tmp2];
+          decode_NSUInteger(aDecoder, &tmp2);
           _gradient_type = (NSGradientType)tmp2;
           [aDecoder decodeValueOfObjCType: @encode(BOOL) at: &tmp];
           _image_dims_when_disabled = tmp;
--- a/Source/NSButtonCell.m
+++ b/Source/NSButtonCell.m
@@ -1579,9 +1579,8 @@
  */
 - (void) encodeWithCoder: (NSCoder*)aCoder
 {
-  BOOL tmp;
-
   [super encodeWithCoder: aCoder];
+
   if ([aCoder allowsKeyedCoding])
     {
       GSButtonCellFlags buttonCellFlags;
@@ -1695,6 +1694,9 @@
     }
   else
     {
+      BOOL tmp;
+      uint32_t tmp2;
+
       [aCoder encodeObject: _keyEquivalent];
       [aCoder encodeObject: _keyEquivalentFont];
       [aCoder encodeObject: _altContents];
@@ -1703,42 +1705,35 @@
       [aCoder encodeValueOfObjCType: @encode(BOOL)
               at: &tmp];
 
-      if([NSButtonCell version] <= 2)
-	{
-	  unsigned int ke = _keyEquivalentModifierMask << 16;
-	  [aCoder encodeValueOfObjCType: @encode(unsigned int)
-		  at: &ke];
-	}
-      else
-	{
-	  [aCoder encodeValueOfObjCType: @encode(unsigned int)
-		  at: &_keyEquivalentModifierMask];
-	}
+      tmp2 = _keyEquivalentModifierMask;
+      [aCoder encodeValueOfObjCType: @encode(uint32_t)
+                                 at: &tmp2];
 
+      tmp2 = _highlightsByMask;
+      [aCoder encodeValueOfObjCType: @encode(uint32_t)
+              at: &tmp2];
+      tmp2 = _showAltStateMask;
       [aCoder encodeValueOfObjCType: @encode(unsigned int)
-              at: &_highlightsByMask];
-      [aCoder encodeValueOfObjCType: @encode(unsigned int)
-              at: &_showAltStateMask];
+              at: &tmp2];
 
-      if([NSButtonCell version] >= 2)
-	{
-	  [aCoder encodeObject: _sound];
-	  [aCoder encodeObject: _backgroundColor];
-	  [aCoder encodeValueOfObjCType: @encode(float)
-		  at: &_delayInterval];
-	  [aCoder encodeValueOfObjCType: @encode(float)
-		  at: &_repeatInterval];
-	  [aCoder encodeValueOfObjCType: @encode(unsigned int)
-		  at: &_bezel_style];
-	  [aCoder encodeValueOfObjCType: @encode(unsigned int)
-		  at: &_gradient_type];
-	  tmp = _image_dims_when_disabled;
-	  [aCoder encodeValueOfObjCType: @encode(BOOL)
-		  at: &tmp];
-	  tmp = _shows_border_only_while_mouse_inside;
-	  [aCoder encodeValueOfObjCType: @encode(BOOL)
-		  at: &tmp];
-	}
+      [aCoder encodeObject: _sound];
+      [aCoder encodeObject: _backgroundColor];
+      [aCoder encodeValueOfObjCType: @encode(float)
+                                 at: &_delayInterval];
+      [aCoder encodeValueOfObjCType: @encode(float)
+                                 at: &_repeatInterval];
+      tmp2 = _bezel_style;
+      [aCoder encodeValueOfObjCType: @encode(uint32_t)
+                                 at: &tmp2];
+      tmp2 = _gradient_type;
+      [aCoder encodeValueOfObjCType: @encode(uint32_t)
+                                 at: &tmp2];
+      tmp = _image_dims_when_disabled;
+      [aCoder encodeValueOfObjCType: @encode(BOOL)
+                                 at: &tmp];
+      tmp = _shows_border_only_while_mouse_inside;
+      [aCoder encodeValueOfObjCType: @encode(BOOL)
+                                 at: &tmp];
     }
 }
 
@@ -1885,27 +1880,36 @@
   else
     {
       BOOL tmp;
+      uint32_t tmp2;
       int version = [aDecoder versionForClassName: @"NSButtonCell"];
       NSString *key = nil;
 
       [aDecoder decodeValueOfObjCType: @encode(id) at: &key];
-      [self setKeyEquivalent: key]; // Set the key equivalent...
+      // Hack to correct a Gorm problem, there "\n" is used instead of "\r".
+      if ([key isEqualToString: @"\n" ])
+        {
+          key = @"\r";
+        }
+      [self setKeyEquivalent: key];
 
       [aDecoder decodeValueOfObjCType: @encode(id) at: &_keyEquivalentFont];
       [aDecoder decodeValueOfObjCType: @encode(id) at: &_altContents];
       [aDecoder decodeValueOfObjCType: @encode(id) at: &_altImage];
       [aDecoder decodeValueOfObjCType: @encode(BOOL) at: &tmp];
       _buttoncell_is_transparent = tmp;
-      [aDecoder decodeValueOfObjCType: @encode(unsigned int)
-                                   at: &_keyEquivalentModifierMask];
+      [aDecoder decodeValueOfObjCType: @encode(uint32_t)
+                                   at: &tmp2];
+      _keyEquivalentModifierMask = (NSUInteger)tmp2;
       if (version <= 2)
         {
           _keyEquivalentModifierMask = _keyEquivalentModifierMask << 16;
         }
-      [aDecoder decodeValueOfObjCType: @encode(unsigned int)
-                                   at: &_highlightsByMask];
-      [aDecoder decodeValueOfObjCType: @encode(unsigned int)
-                                   at: &_showAltStateMask];
+      [aDecoder decodeValueOfObjCType: @encode(uint32_t)
+                                   at: &tmp2];
+      _highlightsByMask = (NSInteger)tmp2;
+      [aDecoder decodeValueOfObjCType: @encode(uint32_t)
+                                   at: &tmp2];
+      _showAltStateMask = (NSInteger)tmp2;
 
       if (version >= 2)
         {
@@ -1913,10 +1917,12 @@
           [aDecoder decodeValueOfObjCType: @encode(id) at: &_backgroundColor];
           [aDecoder decodeValueOfObjCType: @encode(float) at: &_delayInterval];
           [aDecoder decodeValueOfObjCType: @encode(float) at: &_repeatInterval];
-          [aDecoder decodeValueOfObjCType: @encode(unsigned int)
-                                       at: &_bezel_style];
-          [aDecoder decodeValueOfObjCType: @encode(unsigned int)
-                                       at: &_gradient_type];
+          [aDecoder decodeValueOfObjCType: @encode(uint32_t)
+                                       at: &tmp2];
+          _bezel_style = (NSBezelStyle)tmp2;
+          [aDecoder decodeValueOfObjCType: @encode(uint32_t)
+                                       at: &tmp2];
+          _gradient_type = (NSGradientType)tmp2;
           [aDecoder decodeValueOfObjCType: @encode(BOOL) at: &tmp];
           _image_dims_when_disabled = tmp;
           [aDecoder decodeValueOfObjCType: @encode(BOOL) at: &tmp];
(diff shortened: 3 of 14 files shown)


rfm (commented):
The original problem was a straightforward bug: code telling the encoding/decoding methods that a pointer to type X was being passed when in fact what was being passed was a pointer to type Y.  The correct fix IMO is to fix the bug so that when a pointer to Y is passed the the encoding says it's a pointer to Y.  There is generally no problem with the type being one whose size varies between architectures as long as the value passed will fit, and since the code using the instance variable has to work on all architectures you can be pretty confident that the data will fit in the variable on the architecture where the type is smallest.

The aim of the proposed patch seems not to be just one of ensuring that the type specified is the same as the type pointed to, but rather to always decode/encode values that are 32bits in size (fixing the bug by making sure the type specified and the type used are the same seems almost a byproduct).  I don't think it's actually harmful, but this does mean a lot of unnecessary extra code copying ivars to/from temporary variables.
If the rationale is to try to reduce the chance of coding errors then I'm not sure adding intermediate variables is the way to go.

IMO the cleanest solution is to fix the bug by ensuring that the type passed to the methods (using @encode()) is the type of the variable being encoded/decoded.  In an ideal world, we could minimise errors encoding a variable foo by doing
[aCoder encodeValueOfObjCType: @encode(__typeof__(foo)) at &foo];
However I'm pretty sure the compilers don't support applying @encode() to __typeof__()

Another approach would be to add and use wrapper methods for each type so you could do
[aCoder encodeNSInteger: _myInteger]
and
_myInteger = [aCoder decodeNSInteger] 
then the compiler would warn about incorrect types.

rmottola:
I was working in the meanwhile too and prepared beyond NSButtonCell.m also NSMenuItem.m, I just pushed it. While Fred worked on other classes, the approach is similar. However, beyond being careful in uintTmp and intTmp naming, I was extra careful to add an explicit cast - since the explicit casts can take care of size and sign conversion much better.

I do like however Richard's suggestion! With all these tmp and casts, an error is easy.
Apple already puts in nscoder beyond - decodeIntForKey: also:
- decodeInt32ForKey:
- decodeInt64ForKey:
- decodeIntegerForKey:

the last method is exactly the method you propose as decodeNSInteger if I understand correctly just with a Key, we could provide the equivalent decodeInteger (and decodeUInteger?) and the corresponding encoders, what do you think?

They unsigned versions are completely missing. Why? I think that if a mask is encoded, it should be treated as unsigned. I worry that if to/from a signed type of the same size, some values might get lost.

I will try to do that as a proof in a new branch of base+gui.

rfm:
> On 8 Jan 2021, at 18:45, Riccardo <notifications@github.com> wrote:
> 
> 
> I was working in the meanwhile too and prepared beyond NSButtonCell.m also NSMenuItem.m, I just pushed it. While Fred worked on other classes, the approach is similar. However, beyond being careful in uintTmp and intTmp naming, I was extra careful to add an explicit cast - since the explicit casts can take care of size and sign conversion much better.
> 
> I do like however Richard's suggestion! With all these tmp and casts, an error is easy.

Maybe, to get compile-time checking of types, a private header with wrappers like this:

static inline void encodeNSInteger(NSCoder *aCoder, NSInteger value)
{
  [aCoder encodeValueOfObjCType: @encode(NSInteger) at: &value];
}

static inline void decodeNSInteger(NSCoder *aCoder, NSInteger *value;)
{
  [aCoder decodeValueOfObjCType: @encode(NSInteger) at: value];
}

fredkiefer:
The original bug in NSButtonCell isn't that easy to solve without copying the values. Or at least I don't see a way. There we have unsigned int stored in the Gorm file while the actual instance variable has to be of type NSInteger, according to the Apple documentation. We had a bug there even before I made the change in the header, which made it a lot worse.

So sometimes we need copying if we don't want to change the format in the archive. Which is of course something that we can do with versions, but it will break forward compatibility with old code unable to read newer archives. (No matter what strange code Greg and Riccardo are writing for that case)

Also we already have a lot of copying going on for most of the boolean flags that we store in bit fields. And actually a few of the places I changed in the PR are also not fully correct in the size of the int values. I really do not like this new code but it is the safest I could come up with that is not too easy to break. Quite a few of the instance variables have the wrong type, so they may (and should) change in the future.

One weaker argument is also that now we start what range of numbers we guarantee in the encoding. We always promise to deliver 32 bit and nothing more, even if the types used on a specific machine would provide more. I think this is the right way to go. This argument is weak as for most of the data encoded here, even less would do.

Not sure how adding new helpers are a way out of this mess. They would be GNUstep specific and were would we put them?

rfm:
> On 8 Jan 2021, at 20:54, Fred Kiefer <notifications@github.com> wrote:
> 
> 
> The original bug in NSButtonCell isn't that easy to solve without copying the values. Or at least I don't see a way. There we have unsigned int stored in the Gorm file while the actual instance variable has to be of type NSInteger, according to the Apple documentation. We had a bug there even before I made the change in the header, which made it a lot worse.

Sure ... that's a bug that requires a new archive version to fix, but its cause was simply archiving specifying a type different from the type of the variable pointed to.

> So sometimes we need copying if we don't want to change the format in the archive. Which is of course something that we can do with versions, but it will break forward compatibility with old code unable to read newer archives. (No matter what strange code Greg and Riccardo are writing for that case)

I think the existing code in that case creates broken archives on big-endian machines, so there's no way change things so that we can read those archives (though quite possibly nobody ever createsarchives on big endian machines, so it's not an issue).
We should therefore create a new archive version that's not broken, so that in future the archives would contain the correct information irrespective of the architecture on which they were created.

> Also we already have a lot of copying going on for most of the boolean flags that we store in bit fields. And actually a few of the places I changed in the PR are also not fully correct in the size of the int values. I really do not like this new code but it is the safest I could come up with that is not too easy to break. Quite a few of the instance variables have the wrong type, so they may (and should) change in the future.
> 
> One weaker argument is also that now we start what range of numbers we guarantee in the encoding. We always promise to deliver 32 bit and nothing more, even if the types used on a specific machine would provide more. I think this is the right way to go. This argument is weak as for most of the data encoded here, even less would do.

I prefer to NOT limit to 32bit ... if we have a 64bit NSInteger/NSUInteger variable which actually has data in the top 32 bits, and we store that in an archive, that archive would be unreadable on a 32bit system (the unarchiving would raise an exception).  While that could be seen as a bad thing, my feeling is that if the variable needed to be that big than its handling something that is beyond the capacity of a 32bit machine, so it doesn;t matter that we can't unmarchive on a 32bit system.  On the other hand, allowing 64bit values in an archive means that the archive is future-proof for a world where we *do* use that kind of amount of data.  Fundamentally though, it doesn't really matter; the choice is down to the encoding/decoding of individual classes.

> Not sure how adding new helpers are a way out of this mess. They would be GNUstep specific and were would we put them?

I think a private header.

rmottola:
Richard, Fred - I made a proof of concept, I added three convenience methods to NSCoder. Right now in the public header, before completing, Richard can tell me (or change it himself in the branch) to a different private header, gnustep extension, an addition class... I don't know

I left out any other fixes (we can do it in a second pass) - no further cleanup, I just replaced instruction by instruction.
I think the new code is a lot more concise and cleaner and comes out with less temp vaiables, exactly as @rfm suggested. I like it.

Small bonus: by just doing this with the correct type and passing a cast, it actually fixes my decoding issue on the G5 - decoding NSInteger into NSUinteger coming from an unsigned int works.

Source/NSButtonCell.m
```diff
@@ -1703,42 +1705,35 @@ - (void) encodeWithCoder: (NSCoder*)aCoder
       [aCoder encodeValueOfObjCType: @encode(BOOL)
               at: &tmp];
 
-      if([NSButtonCell version] <= 2)
```
gcasa, line 1706: This code was originally put in so that users could save older versions of .gorm files.   This change was very speculative as .gorm file development was not far along at this point and I wanted to provide users a way to create archives for older versions of the libraries.   This hasn't seen much use to my knowledge, so it is okay to remove it as doing so makes the code a lot more straightforward and simple.   Just FYI there is a corresponding change in Gorm which sets the version of certain classes to facilitate this.

Source/NSButtonCell.m
```diff
@@ -1703,42 +1705,35 @@ - (void) encodeWithCoder: (NSCoder*)aCoder
...
+      [aCoder encodeValueOfObjCType: @encode(uint32_t)
+              at: &tmp2];
+      tmp2 = _showAltStateMask;
       [aCoder encodeValueOfObjCType: @encode(unsigned int)
-              at: &_highlightsByMask];
-      [aCoder encodeValueOfObjCType: @encode(unsigned int)
-              at: &_showAltStateMask];
-
-      if([NSButtonCell version] >= 2)
-	{
```
gcasa, line 1724: The same story here as the previous comment.  I believe removing the versioning code in encodeWithCoder: will make things much easier to understand.

Source/NSButtonCell.m
```diff
@@ -1885,38 +1880,49 @@ - (id) initWithCoder: (NSCoder*)aDecoder
...
         {
           [aDecoder decodeValueOfObjCType: @encode(id) at: &_sound];
           [aDecoder decodeValueOfObjCType: @encode(id) at: &_backgroundColor];
           [aDecoder decodeValueOfObjCType: @encode(float) at: &_delayInterval];
           [aDecoder decodeValueOfObjCType: @encode(float) at: &_repeatInterval];
-          [aDecoder decodeValueOfObjCType: @encode(unsigned int)
-                                       at: &_bezel_style];
-          [aDecoder decodeValueOfObjCType: @encode(unsigned int)
-                                       at: &_gradient_type];
+          [aDecoder decodeValueOfObjCType: @encode(uint32_t)
```
gcasa, line 1920: I believe this is the right thing to do.  Going from "unsigned int" to uint32_t is fine as long as we can still read current archives.

gcasa:
> On 8 Jan 2021, at 20:54, Fred Kiefer ***@***.***> wrote: The original bug in NSButtonCell isn't that easy to solve without copying the values. Or at least I don't see a way. There we have unsigned int stored in the Gorm file while the actual instance variable has to be of type NSInteger, according to the Apple documentation. We had a bug there even before I made the change in the header, which made it a lot worse.
> Sure ... that's a bug that requires a new archive version to fix, but its cause was simply archiving specifying a type different from the type of the variable pointed to.

I am, honestly, at a loss to explain why this occurred.  I am wondering if the variable was ever stored as unsigned int in the header.   If we need to bump the version this is not a bad thing, but we need to find a way to make sure that the older archives can still be read properly.

> So sometimes we need copying if we don't want to change the format in the archive. Which is of course something that we can do with versions, but it will break forward compatibility with old code unable to read newer archives. (No matter what strange code Greg and Riccardo are writing for that case)

I would prefer to know what "strange code" you are referring to.  I might be able to explain what the intent was.

> I think the existing code in that case creates broken archives on big-endian machines, so there's no way change things so that we can read those archives (though quite possibly nobody ever createsarchives on big endian machines, so it's not an issue). We should therefore create a new archive version that's not broken, so that in future the archives would contain the correct information irrespective of the architecture on which they were created.

Agreed.  This is possible (no one creating archives on big endian), but there is no reason it shouldn't work going forward.

> Also we already have a lot of copying going on for most of the boolean flags that we store in bit fields. And actually a few of the places I changed in the PR are also not fully correct in the size of the int values. I really do not like this new code but it is the safest I could come up with that is not too easy to break. Quite a few of the instance variables have the wrong type, so they may (and should) change in the future. One weaker argument is also that now we start what range of numbers we guarantee in the encoding. We always promise to deliver 32 bit and nothing more, even if the types used on a specific machine would provide more. I think this is the right way to go. This argument is weak as for most of the data encoded here, even less would do.

Agreed.

> I prefer to NOT limit to 32bit ... if we have a 64bit NSInteger/NSUInteger variable which actually has data in the top 32 bits, and we store that in an archive, that archive would be unreadable on a 32bit system (the unarchiving would raise an exception). While that could be seen as a bad thing, my feeling is that if the variable needed to be that big than its handling something that is beyond the capacity of a 32bit machine, so it doesn;t matter that we can't unmarchive on a 32bit system. On the other hand, allowing 64bit values in an archive means that the archive is future-proof for a world where we *do* use that kind of amount of data. Fundamentally though, it doesn't really matter; the choice is down to the encoding/decoding of individual classes.

Also agreed.

> Not sure how adding new helpers are a way out of this mess. They would be GNUstep specific and were would we put them?
> I think a private header.

Source/NSButtonCell.m
rfm: I will try to express my thoughts in a more concrete way.   Firstly, I think in general the proper way to do things is encode/decode the type of the actual instance variable (except in cases where we are encoding/decoding information that is not stored in a single instance variable of course).  Secondly, while the standard API supports that perfectly well, it does not provide for compile-time type checking, depending on the author of the code to supply correct arguments, and this is clearly a disadvantage.  To provide a mechanism to take advantage of compile-time type checking, I'm suggesting private stuff (eg added to GSPrivate.h in the base library) as follows:

/** Macro to create static wrapper functions to enable the compiler
 * to warn about incorrect variable types during decoding.
 */
#define GS_DECODER(name,type) \
static inline void name(NSCoder *coder, type *value) \
{ \
  [coder decodeValueOfObjCType: @encode(type) at: value]; \
} 

/** Macro to create static wrapper functions to enable the compiler
 * to warn about incorrect variable types during decoding.
 */
#define GS_ENCODER(name,type) \
static inline void name(NSCoder *coder, type *value) \
{ \
  [coder encodeValueOfObjCType: @encode(type) at: value]; \
} 

Using those macros we can then create inline function definitions which are local to the files where they are used (so no pollution of the public API/ABI) and which should have zero runtime cost.

eg.

Somewhere before we want to use it, we do
  GS_ENCODER(encodeNSInteger,NSInteger)
which generates the definition of a static function:
  static inline void encodeNSInteger(NSCoder *coder, NSInteger *value)
  {
    [coder encodeValueOfObjCType: @encode(NSInteger) at: value]; 
  }

Then if we are encoding a class which has an instance variable

  unsigned myIvar;

the code to encode it would look like this:

  encodeNSInteger(aCoder, &myIvar);

The compiler would convert that inline function to 
  [coder encodeValueOfObjCType: @encode(NSInteger) at: &myValue];
so there's no runtime overhead of doing this.
But ... the compiler would do type checking of the argument to the inline function, and would generate a warning that a pointer to an unsigned integer is being passed to a function which expects a pointer to an NSInteger

rfm: Actually, I just remembered we could use the ## to make a better (simpler) macro.
eg.
#define GS_ENCODER(type) \
static inline void encode_##type(NSCoder *coder, type *value) \
{ \
  [coder encodeValueOfObjCType: @encode(type) at: value]; \
}

so

GS_ENCODER(char)

would expand to

static inline void encode_char(NSCoder *coder, char *value)
{
  [coder encodeValueOfObjCType: @encode(char) at: value];
}

gcasa:
Upon further investigation I found this....

```
@interface NSButtonCell : NSActionCell
{
  // Attributes
  NSString *_altContents;
  NSImage *_altImage;
  NSString *_keyEquivalent;
  NSFont *_keyEquivalentFont;
  NSSound *_sound;
  unsigned int _keyEquivalentModifierMask;
  unsigned int _highlightsByMask;
  unsigned int _showAltStateMask;
  float _delayInterval;
  float _repeatInterval;
  NSBezelStyle _bezel_style;
  NSGradientType _gradient_type;
  BOOL _shows_border_only_while_mouse_inside;
  BOOL _mouse_inside;
  // Think of the following as a BOOL ivars
#define _buttoncell_is_transparent _cell.subclass_bool_one
#define _image_dims_when_disabled _cell.subclass_bool_two
}
```

This is the code as it was in version 303dd1baf951b72244967b8f0223d5f130d3ff46 as of 26-May-2005.

You may note that the types in question are in line with the types stored.  I believe the idea here was that if we are storing bitwise flags then there was no need for a signed integer.  Additionally, means that the change from "unsigned int" to NSInteger in the header should have been reflected in the encode/initWithCoder methods as well.

Source/NSButtonCell.m
fredkiefer: I like the idea with the macros. This enables us to have it in gui independent of base. That way we don't need a base release before being able to use it in gui, as Riccardo's solution would require.
I will check whether this results in my compiler (gcc 10) giving proper warnings.

gcasa:
> Upon further investigation I found this....
> 
> ```
> @interface NSButtonCell : NSActionCell
> {
>   // Attributes
>   NSString *_altContents;
>   NSImage *_altImage;
>   NSString *_keyEquivalent;
>   NSFont *_keyEquivalentFont;
>   NSSound *_sound;
>   unsigned int _keyEquivalentModifierMask;
>   unsigned int _highlightsByMask;
>   unsigned int _showAltStateMask;
>   float _delayInterval;
>   float _repeatInterval;
>   NSBezelStyle _bezel_style;
>   NSGradientType _gradient_type;
>   BOOL _shows_border_only_while_mouse_inside;
>   BOOL _mouse_inside;
>   // Think of the following as a BOOL ivars
> #define _buttoncell_is_transparent _cell.subclass_bool_one
> #define _image_dims_when_disabled _cell.subclass_bool_two
> }
> ```
> 
> This is the code as it was in version [303dd1b](https://github.com/gnustep/libs-gui/commit/303dd1baf951b72244967b8f0223d5f130d3ff46) as of 26-May-2005.
> 
> You may note that the types in question are in line with the types stored. I believe the idea here was that if we are storing bitwise flags then there was no need for a signed integer. Additionally, means that the change from "unsigned int" to NSInteger in the header should have been reflected in the encode/initWithCoder methods as well.

The original change from unsigned int to NSInteger was made in version 8ad4aa86c4f875d3b81dd67c78039b5c03b357cc.

Source/NSButtonCell.m
rfm: I wrote:
In an ideal world, we could minimise errors encoding a variable foo by doing
[aCoder encodeValueOfObjCType: @encode(typeof(foo)) at &foo];
However I'm pretty sure the compilers don't support applying @encode() to typeof().

However, in looking at existing code I see we are already using that technique in some places in the base library, so it presumably *does* work for all compilers we target.  That really makes the idea of having extra wrappers to get the compiler to do type checking semi-irrelevant because using typeof() guarantees that we encode/decode correctly as the compiler will supply the type information.
What it doesn't do is handle the situation where we change a variable from signed to unsigned or vice-versa, since the compiler would automatically change the encoding to match, but archives generated by the old code would not be compatible with those generated by the new code, and we would not notice that the archive version should be changed until/unless we try to decode an old archive generated before we made the code change.

rmottola:
I don't like macros, I'm quite spoiled by the new code you can write with new convenience methods, clear, effective, no temporary variables for reading/writing and full type checking.
What's the cost? some extensions to NSCoder, I myself see no issues there, we can do better than Apple, just mark these accordingly or put them in a header. What do you prefer?
I created two branches for base and gui - with the concept and to me they look as the best solution we have now.

In case please comment on the base branch. I see no issue either that we need to release base before gui, we usually do that anyway.

(how to handle a new version is a separate issue, but also quick to do).11

fredkiefer:
Riccardo, you code on NSCoder is quite nice and could be a possible solution. But please step back a bit and try to understand the points Richard is trying to make. Your new code is just a pretty wrap-up for the copying you and I did on our respective branches. Richard didn't like that and your code makes it even worse by adding a method call.

As for the required base release, I remember quite well how often it was you who complained when I relied in gui on new features of base. This shouldn't stop us, but also we should not ignore it as an issue. The same is true for the breaking of forward compatibility of the Gorm encoding. We only should do it if there is no other way.

And just to state this once: I think that we really should drop normal encoding and used keyed encoding instead. Not drop the support for it, instead not use it ourselves, for example in Gorm files.

Source/NSButtonCell.m
fredkiefer: I checkt with my compiler and the warnings I get from misusing your Marcos are great. It will get hard to use these incorrectly.
The second more compact version of the macros won't work for types like "unsigned int" or "long long". As that will result in a white space in the function name. But as long as we stick to simple type names we are good to go.
For gui I would put these macros into the file GSGuiPrivate.h.

In the decoding of NSButtonCell I would still use a temporary variable for four cases. Two where we need a signed integer, while an unsigned is encoded in Gorm files. And the other two where enums get decoded. This is something we could further discuss about.

gcasa:
> Riccardo, you code on NSCoder is quite nice and could be a possible solution. But please step back a bit and try to understand the points Richard is trying to make. Your new code is just a pretty wrap-up for the copying you and I did on our respective branches. Richard didn't like that and your code makes it even worse by adding a method call.
> 
> As for the required base release, I remember quite well how often it was you who complained when I relied in gui on new features of base. This shouldn't stop us, but also we should not ignore it as an issue. The same is true for the breaking of forward compatibility of the Gorm encoding. We only should do it if there is no other way.
> 
> And just to state this once: I think that we really should drop normal encoding and used keyed encoding instead. Not drop the support for it, instead not use it ourselves, for example in Gorm files.

I support this move as well.  Keyed encoding is more resilient to change and can be read by previous archive versions.  It might not be a bad idea to simply use .nib files since so much effort went into ensuring compatibility, but that is just a thought.

On a historical note, I had considered using gmodel files for a long time before we settled on .gorm files in their current incarnation, but gmodels used to be very slow when loading.   At any rate, this is a separate discussion and we should talk about this once this issue is addressed.

gcasa (approved):

Source/NSButtonCell.m
gcasa: @rfm, I concur with Fred on this.  I like the macros you've written.  If they provide is some protection against this type of thing occurring in the future then it's worth doing.

rmottola:
> Riccardo, you code on NSCoder is quite nice and could be a possible solution. But please step back a bit and try to understand the points Richard is trying to make. Your new code is just a pretty wrap-up for the copying you and I did on our respective branches. Richard didn't like that and your code makes it even worse by adding a method call.

True, my code (2nd version) changes exactly nothing, it only has a wrapper call which has several benefits:

- much easier to read code
- type safety
- allows easily explicit casting which can be useful (even works as a hack on the original issue)
This all at the expense of a new method call. I see this irrelevant, we decode maybe some couple of hundreds of objects with a lot of code already, I did not benchmarks, but I think it shouldn't matter and it only affects encoding/decoding of gorms. But perhaps I miss this performance hit and @rfm and @gcasa have more experience there.

I find it so much better... that I continued to convert other classes to see if it holds and continues to. I Think we actually should extend the wrapper methods to floats too.

 
> As for the required base release, I remember quite well how often it was you who complained when I relied in gui on new features of base. This shouldn't stop us, but also we should not ignore it as an issue. The same is true for the breaking of forward compatibility of the Gorm encoding. We only should do it if there is no other way.

About forward compatibility, in the past we often did that, I remember being bitten by that. It is even the reason why Gregory added the option to "encoding" older files. We could easily do that for this bug (although given that I have other controls broken, we might have other issues)

We should then agree on a strategy. We agree that the "32/64 fixed" enforcing we both tried is a dead end - I perhaps falsely interpreted Richard's suggestion and the coder is smarter than I thought.
I see two option with the current gorm format:

1. continue the original plan, with a lot of temporary variables and casts, which works but complicates the code even a little more
2. use wrappers and do almost the same thing but with the easier to read code
3. use macros, which is a mix between 1 and 2... of which I see only the advantage of using optimized functions instead of a real method call.

If we want to really break gorm compatibility or not, we can decide, Richard was in favor, me too. We can get away a cast for the ButtonCell issue, but we still have fields, labels PopUpButton menus to fix. Maybe we can postpone the decision?
 
> And just to state this once: I think that we really should drop normal encoding and used keyed encoding instead. Not drop the support for it, instead not use it ourselves, for example in Gorm files.

This is another big change, right? a totally new format. I like the current format, but of keyed format should have the big advantage of being able to properties which aren't read by an old version. I don't think it would fix type issues though, if there is a mistake done in the previous version or in the code. The advantage is that Apple already has many of the wrapper functions ready. I would miss the Unsigned versions of them though (which is also half of our problem) so again, should we add them extra?

rfm:
> On 9 Jan 2021, at 21:51, Fred Kiefer <notifications@github.com> wrote:
> 

> The second more compact version of the macros won't work for types like "unsigned int" or "long long". As that will result in a white space in the function name.

Good point.

> But as long as we stick to simple type names we are good to go.
> For gui I would put these macros into the file GSGuiPrivate.h.
> 
> In the decoding of NSButtonCell I would still use a temporary variable for four cases. Two where we need a signed integer, while an unsigned is encoded in Gorm files.

I think there is no reasonable fix possible without changing coding/decoding in base.  As I understand it the problem here is that we accidentally changed from coding a value as an unsigned to coding it as a signed.  In theory that means there will exist some archives where it is encoded one way and some archives where it is encoded another way and other archives where it is encoded the other way.  The only ways to handle both archives would be:
1. base is changed to permit a signed to be decoded as an unsigned or an unsigned to be decoded as a signed (as long as the stored value is actually positive, or the encoded and decoded values are the same size), so we can just decode both forms of archive.  If the stored value is negative and the size of the encoder is smaller than that of the decoder differ, I don't see how base could know whether sign extension of the smaller value was appropriate or not (it wouldn't know whether the signed version or the unsigned version was correct because it wouldn't know whether the value was being used as a number or a bit-mask) etc.
2. the gui is changed to try decoding the entire archive assuming the values are signed, and if that fails try the entire archive again ... horribly messy/inefficient because (iirc) there's no way to just retry decoding an individual value rather than the entire archive. 

But, maybe the number of broken archives is so small that it's just possible to ignore the issue of old archives, and move on by producing code that will work for future archives.

> And the other two where enums get decoded. This is something we could further discuss about.

Enumerated types are, I believe, constrained by C standards to fit inside an int ... but theoretically a compiler is free to use shorter storage if all the enumeration values will fit (though likely all compilers simply use int for all enumerations).

So if we have

enum { foo, bar } x = foo;

we can probably do encode_int(coder, &x);

Certainly, if we ensure that the enumeration contains -1 we should be guaranteed that the compiler stores it as an int.  So a declaration like:

enum {foo, bar, unused = -1} x;

should guarantee that encode_int(coder, &x); and decode_int(coder, &x); would work.

fredkiefer:
>  As I understand it the problem here is that we accidentally changed from coding a value as an unsigned to coding it as a signed. 

Not exactly. The encoding/decoding was always done as unsigned. But the actual instance variable was changed into signed, as it always should have been according to the spec. So we don't have a problem with mixed archives or all that stuff.

I try to rewrite my PR to use your macros and we all have a look at that.


@rmottola, no the macros are not a mix between 1&2. These two attempts try to hide the issue by using a local variable for (de-)coding. The macros are trying to make the issue visible to force the programmer to deal with it. In many cases a local copy will still be the best way to handle it, but that should then be a conscious decision.

fredkiefer:
I added the macros suggested by Richard and adjusted my pervious changes to use these. As you can see it really pays when we have enums that have a typedef to NSUInteger oder NSInteger, as this allows us to use the macros directly. In other cases we need the copying.

Source/GSGuiPrivate.h
```diff
@@ -103,5 +104,24 @@ static inline CGFloat GSRoundTowardsNegativeInfinity(CGFloat x)
   return ceil(x - 0.5);
 }
 
+#if !defined(GS_DECODER)
+#define GS_DECODER(type) \
+static inline void decode_##type(NSCoder *coder, type *value) \
+{ \
+[coder decodeValueOfObjCType: @encode(type) at: value]; \
+}
+
```
gcasa, line 113: While I don't believe it's as much of a risk for keyed archives, should there be an equivalent set of macros for them?

Source/NSActionCell.m
```diff
@@ -386,10 +387,7 @@ - (void) encodeWithCoder: (NSCoder*)aCoder
     }
   else
     {
-      int32_t tmp;
-
-      tmp = _tag;
-      [aCoder encodeValueOfObjCType: @encode(int32_t) at: &tmp];
+      encode_NSInteger(aCoder, &_tag);
```
gcasa, line 390: As you previously mentioned "long long" or any type with spaces in it will break this macro.  Is there any way to fix this?

rfm:
> On 11 Jan 2021, at 05:42, Gregory Casamento <notifications@github.com> wrote:
> 
> 
> @gcasa commented on this pull request.
> 
> In Source/GSGuiPrivate.h:
> 
> > @@ -103,5 +104,24 @@ static inline CGFloat GSRoundTowardsNegativeInfinity(CGFloat x)
>    return ceil(x - 0.5);
>  }
>  
> +#if !defined(GS_DECODER)
> +#define GS_DECODER(type) \
> +static inline void decode_##type(NSCoder *coder, type *value) \
> +{ \
> +[coder decodeValueOfObjCType: @encode(type) at: value]; \
> +}
> +
> 
> While I don't believe it's as much of a risk for keyed archives, should there be an equivalent set of macros for them?
> 
> In Source/NSActionCell.m:
> 
> > @@ -386,10 +387,7 @@ - (void) encodeWithCoder: (NSCoder*)aCoder
>      }
>    else
>      {
> -      int32_t tmp;
> -
> -      tmp = _tag;
> -      [aCoder encodeValueOfObjCType: @encode(int32_t) at: &tmp];
> +      encode_NSInteger(aCoder, &_tag);
> 
> As you previously mentioned "long long" or any type with spaces in it will break this macro. Is there any way to fix this?
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or unsubscribe.
> 


The previous version of the macro had another argument to supply the function name explicitly.
Without that macro, typedef could be used.
eg.
typedef long long longLong;
GS_DECODER(longLong)
would produce a longLong_decode() function that could be used to decode long long values.

rmottola:
I went on and converted also NSView in my branch, to integer/unsigned I also added BOOL and float as macros and I see it pays off, the code is very easy to read and I greatly prefer the method/return system. I compare it between the old code your macro code and my code.
I do not understand why explicit methods should be less safe than your macros in type-checking, input and return values are typed and there is no hassle with pass by reference!

I may add that Apple went on adding new helper methods for keyed encoding, why can't we too?

And last but not least, it could be a hint to cleanup internal ivars if we need too many different helper types.
My 2 cents.

fredkiefer:
Could we please reach a decision here? Should I go ahead and commit these changes and address the same issues on other gui classes? Or use the new helper methods that Riccardo suggested? Both are possible solutions. I tried to explain the difference in a previous post so all arguments are on the table.

gcasa:
> Could we please reach a decision here? Should I go ahead and commit these changes and address the same issues on other gui classes? Or use the new helper methods that Riccardo suggested? Both are possible solutions. I tried to explain the difference in a previous post so all arguments are on the table.

I trust your judgment, Fred.  I don't have strong feelings for either solution as both have advantages and disadvantages.      Please use whichever solution you feel the most comfortable with.  GC

gcasa:
> On 11 Jan 2021, at 05:42, Gregory Casamento ***@***.***> wrote: @gcasa commented on this pull request. In Source/GSGuiPrivate.h: > @@ -103,5 +104,24 @@ static inline CGFloat GSRoundTowardsNegativeInfinity(CGFloat x) return ceil(x - 0.5); } +#if !defined(GS_DECODER) +#define GS_DECODER(type) \ +static inline void decode_##type(NSCoder *coder, type *value) \ +{ \ +[coder decodeValueOfObjCType: @encode(type) at: value]; \ +} + While I don't believe it's as much of a risk for keyed archives, should there be an equivalent set of macros for them? In Source/NSActionCell.m: > @@ -386,10 +387,7 @@ - (void) encodeWithCoder: (NSCoder*)aCoder } else { - int32_t tmp; - - tmp = _tag; - [aCoder encodeValueOfObjCType: @encode(int32_t) at: &tmp]; + encode_NSInteger(aCoder, &_tag); As you previously mentioned "long long" or any type with spaces in it will break this macro. Is there any way to fix this? — You are receiving this because you were mentioned. Reply to this email directly, view it on GitHub, or unsubscribe.
> The previous version of the macro had another argument to supply the function name explicitly. Without that macro, typedef could be used. eg. typedef long long longLong; GS_DECODER(longLong) would produce a longLong_decode() function that could be used to decode long long values.

That, I believe, is a workable solution.

rfm:
> On 15 Jan 2021, at 14:07, Fred Kiefer <notifications@github.com> wrote:
> 
> 
> Could we please reach a decision here? Should I go ahead and commit these changes and address the same issues on other gui classes? Or use the new helper methods that Riccardo suggested? Both are possible solutions. I tried to explain the difference in a previous post so all arguments are on the table.

Rather like Greg ... I think simply having a bugfix is fine, however you like.
My suggestions were just intended to make that as lightweight as possible (particularly since we are talking about deprecated API, and expect everyone to be using keyed coding in future).

fredkiefer:
OK, I go ahead and merge this PR.
As for Riccardo's approach I suggest that he first cleans up the base changes by moving them into a category. We should then merge them and after the next base release We may discuss again about using that code in gui.

rmottola:
I confirm, your fix of the fix... unlocked all GUI initialization problems - I see labels, popupbuttons, check and radios working. I did a quick test on a couple of applications. On PPC64 so it looks fine! On SPARC64 the same problems are gone, only a color issue remains, but this is most probably back related? A separate thing for sure.

Please tell me more about the category, so I can prepare it, although no hurry for release.

Issue #91, opened 2021-02-06
-[NSCollectionView initWithCoder:] - Class lack non-keyed encoding...


gcasa:
The initWithCoder: method in this class lacks non-keyed coding implementation.

PR #92, opened 2021-02-06
Fix issue #91, add non-keyed encoding to initWithCoder:/encodeWithCoder

gcasa:
Adds initWithCoder:/encodeWithCoder implementation for NSCollectionView.

### DIFF
--- a/Source/NSCollectionView.m
+++ b/Source/NSCollectionView.m
@@ -602,11 +602,11 @@ static NSString *placeholderItem = nil;
   
   if (self)
     {
+      _itemSize = NSMakeSize(0, 0);
+      _tileWidth = -1.0;
+          
       if ([aCoder allowsKeyedCoding])
         {
-          _itemSize = NSMakeSize(0, 0);
-          _tileWidth = -1.0;
-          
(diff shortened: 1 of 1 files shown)


Source/NSCollectionView.m
```diff
@@ -622,6 +622,12 @@ - (id) initWithCoder: (NSCoder *)aCoder
         }
       else
         {
+          _minItemSize = [aCoder decodeSize];
+          _maxItemSize = [aCoder decodeSize];
+          [aCoder decodeValueOfObjCType: @encode(int64_t) at: &_maxNumberOfRows];
```
fredkiefer, line 627: As you may remember we just decided on not using explicit sized integer variants in coding. Could you please instead use NSUInteger and maybe even better use the new macros from GSGuiPrivate.h?

gcasa: Okay, I will change that.  Thx

fredkiefer (approved):
Looks fine to me now. Is this encoding compatible with Apple's or is it just something you made up? That wouldn't be too bad, as most of our other encodings in guy are that way.

PR #106, opened 2021-06-14
NSCollectionView branch -- Adding layouts

gcasa:
Adding layouts for collection views.

### DIFF
--- a/Headers/AppKit/NSCollectionView.h
+++ b/Headers/AppKit/NSCollectionView.h
@@ -431,7 +431,6 @@ APPKIT_EXPORT_CLASS
 
   // Managing items
   NSMutableArray *_visibleItems;
-  NSMutableSet *_indexPathsForVisibleItems;
   NSMutableDictionary *_visibleSupplementaryViews;
   NSMutableSet *_indexPathsForSupplementaryElementsOfKind;
 
--- a/Headers/AppKit/NSCollectionView.h
+++ b/Headers/AppKit/NSCollectionView.h
@@ -45,7 +45,6 @@
 @class NSPasteboard;
 @class NSNib;
 @class NSMapTable;
-@class NSMutableArray;
 @class NSMutableDictionary;
 @class NSMutableSet;
 
--- a/Headers/AppKit/NSCollectionView.h
+++ b/Headers/AppKit/NSCollectionView.h
@@ -414,6 +414,7 @@ APPKIT_EXPORT_CLASS
   IBOutlet id <NSCollectionViewPrefetching> _prefetchDataSource;
 
   BOOL _backgroundViewScrollsWithContent;
+  BOOL _allowReload;
 
   // Size
   NSSize _itemSize;
--- a/Headers/AppKit/NSCollectionView.h
+++ b/Headers/AppKit/NSCollectionView.h
@@ -414,45 +414,17 @@ APPKIT_EXPORT_CLASS
 {
   // Content
   NSArray *_content;
-  IBOutlet NSCollectionViewItem *itemPrototype;
   NSMutableArray *_items;
-
-  // Selection management...
-  BOOL _allowsMultipleSelection;
-  BOOL _isSelectable;
-  BOOL _allowsEmptySelection;
   NSIndexSet *_selectionIndexes;
-  NSSet *_selectionIndexPaths;
-  
+  NSSet *_selectionIndexPaths;  
   NSArray *_backgroundColors;
 
   // Delegate and datasource
+  IBOutlet NSCollectionViewItem *itemPrototype;
   IBOutlet id <NSCollectionViewDelegate> _delegate;
   IBOutlet id <NSCollectionViewDataSource> _dataSource;
   IBOutlet NSView *_backgroundView;
   IBOutlet id <NSCollectionViewPrefetching> _prefetchDataSource;
-
-  BOOL _backgroundViewScrollsWithContent;
-  BOOL _allowReload;
-
-  // Size
-  NSSize _itemSize;
-  NSSize _maxItemSize;
-  NSSize _minItemSize;
-  CGFloat _tileWidth;
-  CGFloat _verticalMargin;
-  CGFloat _horizontalMargin;
-
-  NSUInteger _maxNumberOfColumns;
-  NSUInteger _maxNumberOfRows;
-  NSUInteger _numberOfColumns;
-
-  // Drag & Drop
-  NSDragOperation _draggingSourceOperationMaskForLocal;
-  NSDragOperation _draggingSourceOperationMaskForRemote;
-  
-  NSUInteger _draggingOnRow;
-  NSUInteger _draggingOnIndex;
 
   // Layout
   NSCollectionViewLayout *_collectionViewLayout;
@@ -472,6 +444,31 @@ APPKIT_EXPORT_CLASS
   // Registered class/nib for item identifier
   NSMapTable *_registeredNibs;
   NSMapTable *_registeredClasses;
+
+  // Selection management...
+  BOOL _allowsMultipleSelection;
+  BOOL _isSelectable;
+  BOOL _allowsEmptySelection;
+  BOOL _backgroundViewScrollsWithContent;
+  BOOL _allowReload;
+
+  // Size
+  NSSize _itemSize;
+  NSSize _maxItemSize;
+  NSSize _minItemSize;
+  CGFloat _tileWidth;
+  CGFloat _verticalMargin;
+  CGFloat _horizontalMargin;
+
+  NSUInteger _maxNumberOfColumns;
+  NSUInteger _maxNumberOfRows;
+  NSUInteger _numberOfColumns;
+
+  // Drag & Drop
+  NSDragOperation _draggingSourceOperationMaskForLocal;
+  NSDragOperation _draggingSourceOperationMaskForRemote;  
+  NSUInteger _draggingOnRow;
+  NSUInteger _draggingOnIndex;
 }
 
 - (BOOL) allowsMultipleSelection;
--- a/Headers/AppKit/NSCollectionView.h
+++ b/Headers/AppKit/NSCollectionView.h
@@ -132,7 +132,7 @@ typedef NSString *NSUserInterfaceItemIdentifier;
 
 - (NSView *) collectionView: (NSCollectionView *)collectionView
              viewForSupplementaryElementOfKind: (NSCollectionViewSupplementaryElementKind)kind
-                atIndexPath:(NSIndexPath *)indexPath;
+                atIndexPath: (NSIndexPath *)indexPath;
 @end
 #endif
 
@@ -144,14 +144,14 @@ typedef NSString *NSUserInterfaceItemIdentifier;
 /**
  * Load the items listed in indexPaths in collectionView before they are displayed
  */
-- (void)collectionView:(NSCollectionView *)collectionView prefetchItemsAtIndexPaths:(NSArray *)indexPaths;
+- (void)collectionView: (NSCollectionView *)collectionView prefetchItemsAtIndexPaths: (NSArray *)indexPaths;
 #if GS_PROTOCOLS_HAVE_OPTIONAL
 @optional
 #endif
 /**
  * Cancel the prefetch request for the listed indexPaths.
  */ 
-- (void)collectionView:(NSCollectionView *)collectionView cancelPrefetchingForItemsAtIndexPaths:(NSArray *)indexPaths;
+- (void)collectionView: (NSCollectionView *)collectionView cancelPrefetchingForItemsAtIndexPaths: (NSArray *)indexPaths;
 @end
 #endif
 
@@ -312,7 +312,7 @@ typedef NSString *NSUserInterfaceItemIdentifier;
       forItemsAtIndexes: (NSIndexSet *)indexes;
 
 /**
- * Allows application to specify the screenPoint at which a dragging session will end
+ * Allows application to specify the screenPoint at which a dragging session did end
  */
 - (void) collectionView: (NSCollectionView *)collectionView
         draggingSession: (NSDraggingSession *)session
@@ -565,7 +565,7 @@ APPKIT_EXPORT_CLASS
 
 - (void)registerClass: (Class)viewClass 
         forSupplementaryViewOfKind: (NSCollectionViewSupplementaryElementKind)kind 
-       withIdentifier:(NSUserInterfaceItemIdentifier)identifier;
+       withIdentifier: (NSUserInterfaceItemIdentifier)identifier;
 
 - (void) registerNib: (NSNib *)nib 
          forSupplementaryViewOfKind: (NSCollectionViewSupplementaryElementKind)kind 
--- a/Headers/AppKit/NSCollectionView.h
+++ b/Headers/AppKit/NSCollectionView.h
@@ -443,12 +443,12 @@ APPKIT_EXPORT_CLASS
   NSMutableSet *_indexPathsForVisibleItems;
   NSMutableDictionary *_visibleSupplementaryViews;
   NSMutableSet *_indexPathsForSupplementaryElementsOfKind;
-  NSMutableDictionary *_itemsToAttributes;
 
   // Private
   // Map items -> indexPath
   NSMapTable *_itemsToIndexPaths;
   NSMapTable *_indexPathsToItems;
+  NSMapTable *_itemsToAttributes;
   
   // Registered class/nib for item identifier
   NSMapTable *_registeredNibs;
--- a/Headers/AppKit/NSCollectionView.h
+++ b/Headers/AppKit/NSCollectionView.h
@@ -444,14 +444,15 @@ APPKIT_EXPORT_CLASS
   NSMutableDictionary *_visibleSupplementaryViews;
   NSMutableSet *_indexPathsForSupplementaryElementsOfKind;
   NSMutableDictionary *_itemsToAttributes;
+
+  // Private
+  // Map items -> indexPath
+  NSMapTable *_itemsToIndexPaths;
+  NSMapTable *_indexPathsToItems;
   
   // Registered class/nib for item identifier
   NSMapTable *_registeredNibs;
   NSMapTable *_registeredClasses;
-
-  // Registered class/nib for item kind & identifier
-  NSMapTable *_registeredNibForItemWithIdentifier;
-  NSMapTable *_registeredClassForItemWithIdentifier;
 }
 
 - (BOOL) allowsMultipleSelection;
--- a/Headers/AppKit/NSCollectionView.h
+++ b/Headers/AppKit/NSCollectionView.h
@@ -394,10 +394,12 @@ shouldSelectItemsAtIndexPaths: (NSSet *)indexPaths;
 APPKIT_EXPORT_CLASS
 @interface NSCollectionView : NSView //<NSDraggingDestination, NSDraggingSource>
 {
+  // Content
   NSArray *_content;
   IBOutlet NSCollectionViewItem *itemPrototype;
   NSMutableArray *_items;
-  
+
+  // Selection management...
   BOOL _allowsMultipleSelection;
   BOOL _isSelectable;
   BOOL _allowsEmptySelection;
@@ -406,6 +408,7 @@ APPKIT_EXPORT_CLASS
   
   NSArray *_backgroundColors;
 
+  // Delegate and datasource
   IBOutlet id <NSCollectionViewDelegate> _delegate;
   IBOutlet id <NSCollectionViewDataSource> _dataSource;
   IBOutlet NSView *_backgroundView;
@@ -413,6 +416,7 @@ APPKIT_EXPORT_CLASS
 
   BOOL _backgroundViewScrollsWithContent;
 
+  // Size
   NSSize _itemSize;
   NSSize _maxItemSize;
   NSSize _minItemSize;
@@ -423,13 +427,15 @@ APPKIT_EXPORT_CLASS
   NSUInteger _maxNumberOfColumns;
   NSUInteger _maxNumberOfRows;
   NSUInteger _numberOfColumns;
-  
+
+  // Drag & Drop
   NSDragOperation _draggingSourceOperationMaskForLocal;
   NSDragOperation _draggingSourceOperationMaskForRemote;
   
   NSUInteger _draggingOnRow;
   NSUInteger _draggingOnIndex;
 
+  // Layout
   NSCollectionViewLayout *_collectionViewLayout;
 
   // Managing items
--- a/Headers/AppKit/NSCollectionView.h
+++ b/Headers/AppKit/NSCollectionView.h
@@ -36,6 +36,7 @@
 #import <AppKit/NSDragging.h>
 #import <AppKit/NSNibDeclarations.h>
 #import <AppKit/NSView.h>
+#import <AppKit/NSUserInterfaceItemIdentification.h>
 
 @class NSCollectionViewItem;
 @class NSCollectionView;
@@ -95,6 +96,24 @@ typedef NSUInteger NSCollectionViewScrollPosition;
 
 typedef NSString *NSCollectionViewSupplementaryElementKind;
 typedef NSString *NSUserInterfaceItemIdentifier;
+
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_11, GS_API_LATEST)
+@protocol NSCollectionViewElement <NSObject, NSUserInterfaceItemIdentification>
+#if GS_PROTOCOLS_HAVE_OPTIONAL
+@optional
+#endif
+
+- (void) prepareForReuse;
+
+- (void) applyLayoutAttributes: (NSCollectionViewLayoutAttributes *)layoutAttributes;
+
+- (void) willTransitionFromLayout: (NSCollectionViewLayout *)oldLayout toLayout: (NSCollectionViewLayout *)newLayout;
+
+- (void) didTransitionFromLayout: (NSCollectionViewLayout *)oldLayout toLayout: (NSCollectionViewLayout *)newLayout;
+
+- (NSCollectionViewLayoutAttributes *) preferredLayoutAttributesFittingAttributes: (NSCollectionViewLayoutAttributes *)layoutAttributes;
+@end
+#endif
 
 #if OS_API_VERSION(MAC_OS_X_VERSION_10_11, GS_API_LATEST)
 @protocol NSCollectionViewDataSource <NSObject>
--- a/Headers/AppKit/NSCollectionViewLayout.h
+++ b/Headers/AppKit/NSCollectionViewLayout.h
@@ -29,6 +29,7 @@
 #import <Foundation/NSGeometry.h>
 
 #import <AppKit/NSCollectionView.h>
+#import <AppKit/AppKitDefines.h>
 
 #if OS_API_VERSION(MAC_OS_X_VERSION_10_11, GS_API_LATEST)
 
@@ -57,6 +58,7 @@ NSCollectionViewSupplementaryElementKind NSCollectionElementKindInterItemGapIndi
 @class NSNib;
 @class NSSet;
 
+APPKIT_EXPORT_CLASS
 @interface NSCollectionViewLayoutAttributes : NSObject <NSCopying>
 {
   NSRect _frame;
@@ -111,6 +113,7 @@ enum
 };
 typedef NSInteger NSCollectionUpdateAction;
 
+APPKIT_EXPORT_CLASS  
 @interface NSCollectionViewUpdateItem : NSObject
 {
   NSIndexPath *_indexPathBeforeUpdate;
@@ -124,7 +127,7 @@ typedef NSInteger NSCollectionUpdateAction;
 
 @end
 
-
+APPKIT_EXPORT_CLASS
 @interface NSCollectionViewLayoutInvalidationContext : NSObject
 {
   NSPoint _contentOffsetAdjustment;
@@ -158,6 +161,7 @@ typedef NSInteger NSCollectionUpdateAction;
 
 @end
 
+APPKIT_EXPORT_CLASS
 @interface NSCollectionViewLayout : NSObject <NSCoding>
 {
   NSCollectionView *_collectionView; // weak
@@ -183,6 +187,7 @@ typedef NSInteger NSCollectionUpdateAction;
   
 @end
 
+APPKIT_EXPORT_CLASS
 @interface NSCollectionViewLayout (NSSubclassingHooks)
   
 // Methods to override for specific layouts...
@@ -218,6 +223,7 @@ typedef NSInteger NSCollectionUpdateAction;
 
 @end
 
+APPKIT_EXPORT_CLASS
 @interface NSCollectionViewLayout (NSUpdateSupportHooks)
 
 // Update support
--- a/Headers/AppKit/NSCollectionViewLayout.h
+++ b/Headers/AppKit/NSCollectionViewLayout.h
@@ -48,7 +48,7 @@ typedef NSInteger NSCollectionElementCategory;
   
 
 typedef NSString *NSCollectionViewDecorationElementKind;
-const static NSCollectionViewSupplementaryElementKind NSCollectionElementKindInterItemGapIndicator;
+APPKIT_EXPORT NSCollectionViewSupplementaryElementKind NSCollectionElementKindInterItemGapIndicator;
 
 @class NSCollectionViewLayoutAttributes;
 @class NSCollectionView;
--- a/Headers/AppKit/NSCollectionViewLayout.h
+++ b/Headers/AppKit/NSCollectionViewLayout.h
@@ -162,7 +162,6 @@ typedef NSInteger NSCollectionUpdateAction;
 {
   NSCollectionView *_collectionView; // weak
   BOOL _valid;
-  NSMutableDictionary *_itemsToAttributes;
 
   Class _layoutAttributesClass;
   Class _invalidationContextClass;
--- a/Headers/AppKit/NSCollectionViewLayout.h
+++ b/Headers/AppKit/NSCollectionViewLayout.h
@@ -250,10 +250,10 @@ APPKIT_EXPORT_CLASS
   finalLayoutAttributesForDisappearingDecorationElementOfKind: (NSCollectionViewDecorationElementKind)elementKind
                                                    atIndexPath: (NSIndexPath *)decorationIndexPath;
 
-- (NSSet *)indexPathsToDeleteForSupplementaryViewOfKind: (NSCollectionViewSupplementaryElementKind)elementKind;
-- (NSSet *)indexPathsToDeleteForDecorationViewOfKind: (NSCollectionViewDecorationElementKind)elementKind;
-- (NSSet *)indexPathsToInsertForSupplementaryViewOfKind: (NSCollectionViewSupplementaryElementKind)elementKind;
-- (NSSet *)indexPathsToInsertForDecorationViewOfKind: (NSCollectionViewDecorationElementKind)elementKind;
+- (NSSet *) indexPathsToDeleteForSupplementaryViewOfKind: (NSCollectionViewSupplementaryElementKind)elementKind;
+- (NSSet *) indexPathsToDeleteForDecorationViewOfKind: (NSCollectionViewDecorationElementKind)elementKind;
+- (NSSet *) indexPathsToInsertForSupplementaryViewOfKind: (NSCollectionViewSupplementaryElementKind)elementKind;
+- (NSSet *) indexPathsToInsertForDecorationViewOfKind: (NSCollectionViewDecorationElementKind)elementKind;
 
 @end
 
--- a/Headers/AppKit/NSCollectionViewLayout.h
+++ b/Headers/AppKit/NSCollectionViewLayout.h
@@ -48,7 +48,7 @@ typedef NSInteger NSCollectionElementCategory;
   
 
 typedef NSString *NSCollectionViewDecorationElementKind;
-NSCollectionViewSupplementaryElementKind NSCollectionElementKindInterItemGapIndicator;
+const NSCollectionViewSupplementaryElementKind NSCollectionElementKindInterItemGapIndicator;
 
 @class NSCollectionViewLayoutAttributes;
 @class NSCollectionView;
--- a/Headers/AppKit/NSCollectionViewLayout.h
+++ b/Headers/AppKit/NSCollectionViewLayout.h
@@ -62,7 +62,6 @@ APPKIT_EXPORT_CLASS
 @interface NSCollectionViewLayoutAttributes : NSObject <NSCopying>
 {
   NSRect _frame;
-  NSSize _size;
   CGFloat _alpha;
   NSInteger _zIndex;
   BOOL _hidden; 
--- a/Headers/AppKit/NSCollectionViewLayout.h
+++ b/Headers/AppKit/NSCollectionViewLayout.h
@@ -48,7 +48,7 @@ typedef NSInteger NSCollectionElementCategory;
   
 
 typedef NSString *NSCollectionViewDecorationElementKind;
-const NSCollectionViewSupplementaryElementKind NSCollectionElementKindInterItemGapIndicator;
+const static NSCollectionViewSupplementaryElementKind NSCollectionElementKindInterItemGapIndicator;
 
 @class NSCollectionViewLayoutAttributes;
 @class NSCollectionView;
--- a/Headers/AppKit/NSFont.h
+++ b/Headers/AppKit/NSFont.h
@@ -74,7 +74,7 @@ typedef enum _NSFontRenderingMode
 } NSFontRenderingMode;
 #endif
 
-APPKIT_EXPORT const CGFloat *NSFontIdentityMatrix;
+APPKIT_EXPORT const CGFloat NSFontIdentityMatrix[6];
 
 APPKIT_EXPORT_CLASS
 @interface NSFont : NSObject <NSCoding, NSCopying>
--- a/Source/GSXib5KeyedUnarchiver.m
+++ b/Source/GSXib5KeyedUnarchiver.m
@@ -3562,7 +3562,7 @@ didStartElement: (NSString*)elementName
                ([@"NSIntercellSpacingWidth" isEqualToString: key]))
         {
           hasValue = [currentElement elementForKey: @"intercellSpacing"] != nil;
-        } 
+        }
       else if ([@"NSContents" isEqualToString: key])
         {
           hasValue  = [currentElement attributeForKey: @"title"] != nil;
--- a/Source/GSXib5KeyedUnarchiver.m
+++ b/Source/GSXib5KeyedUnarchiver.m
@@ -326,7 +326,7 @@ static NSArray      *XmlBoolDefaultYes  = nil;
                                            @"beginningViews", @"NSStackViewBeginningContainer",  // NSStackView
                                            @"middleViews", @"NSStackViewMiddleContainer",
                                            @"endViews", @"NSStackViewEndContainer",
-                                           @"primaryBackgroundColor", @"NSBackgroundColors",
+                                         // @"primaryBackgroundColor", @"NSBackgroundColors",
                                            nil];
           RETAIN(XmlKeyMapTable);
 
@@ -3549,6 +3549,10 @@ didStartElement: (NSString*)elementName
       if ([XmlKeyMapTable objectForKey: key])
         {
           hasValue = [self containsValueForKey: [XmlKeyMapTable objectForKey: key]];
+        }
+      else if ([@"NSBackgroundColors" isEqualToString: key])
+        {
+          hasValue = [currentElement elementForKey: @"primaryBackgroundColor"] != nil;
         }
       else if ([@"NSCollectionViewLayout" isEqualToString: key])
         {
--- a/Source/GSXib5KeyedUnarchiver.m
+++ b/Source/GSXib5KeyedUnarchiver.m
@@ -43,8 +43,6 @@
 #import "AppKit/NSBrowserCell.h"
 #import "AppKit/NSButtonCell.h"
 #import "AppKit/NSCell.h"
-#import "AppKit/NSCollectionView.h"
-#import "AppKit/NSCollectionViewLayout.h"
 #import "AppKit/NSClipView.h"
 #import "AppKit/NSFormCell.h"
 #import "AppKit/NSGridView.h"
@@ -327,7 +325,6 @@ static NSArray      *XmlBoolDefaultYes  = nil;
                                            @"middleViews", @"NSStackViewMiddleContainer",
                                            @"endViews", @"NSStackViewEndContainer",
 					   @"collectionViewLayout", @"NSCollectionViewLayout",
-                                         // @"primaryBackgroundColor", @"NSBackgroundColors",
                                            nil];
           RETAIN(XmlKeyMapTable);
 
--- a/Source/GSXib5KeyedUnarchiver.m
+++ b/Source/GSXib5KeyedUnarchiver.m
@@ -326,6 +326,7 @@ static NSArray      *XmlBoolDefaultYes  = nil;
                                            @"beginningViews", @"NSStackViewBeginningContainer",  // NSStackView
                                            @"middleViews", @"NSStackViewMiddleContainer",
                                            @"endViews", @"NSStackViewEndContainer",
+					   @"collectionViewLayout", @"NSCollectionViewLayout",
                                          // @"primaryBackgroundColor", @"NSBackgroundColors",
                                            nil];
           RETAIN(XmlKeyMapTable);
@@ -3325,10 +3326,6 @@ didStartElement: (NSString*)elementName
         {
           object = [self decodeObjectForKey: @"primaryBackgroundColor"];
         }
-      else if ([@"NSCollectionViewLayout" isEqualToString: key])
-        {
-          object = [self decodeObjectForKey: @"collectionViewLayout"];
-        }
       else if (([@"NSSearchButtonCell" isEqualToString: key]) ||
                ([@"NSCancelButtonCell" isEqualToString: key]))
         {
@@ -3554,10 +3551,6 @@ didStartElement: (NSString*)elementName
         {
           hasValue = [currentElement elementForKey: @"primaryBackgroundColor"] != nil;
         }
-      else if ([@"NSCollectionViewLayout" isEqualToString: key])
-        {
-          hasValue = [currentElement elementForKey: @"collectionViewLayout"] != nil;
-        } 
       else if (([@"NSIntercellSpacingHeight" isEqualToString: key]) ||
                ([@"NSIntercellSpacingWidth" isEqualToString: key]))
         {
--- a/Source/GSXib5KeyedUnarchiver.m
+++ b/Source/GSXib5KeyedUnarchiver.m
@@ -326,6 +326,7 @@ static NSArray      *XmlBoolDefaultYes  = nil;
                                            @"beginningViews", @"NSStackViewBeginningContainer",  // NSStackView
                                            @"middleViews", @"NSStackViewMiddleContainer",
                                            @"endViews", @"NSStackViewEndContainer",
+                                           @"primaryBackgroundColor", @"NSBackgroundColors",
                                            nil];
           RETAIN(XmlKeyMapTable);
 
@@ -3549,10 +3550,6 @@ didStartElement: (NSString*)elementName
         {
           hasValue = [self containsValueForKey: [XmlKeyMapTable objectForKey: key]];
         }
-      else if ([@"NSBackgroundColors" isEqualToString: key])
-        {
-          hasValue = [currentElement elementForKey: @"primaryBackgroundColor"] != nil;
-        } 
       else if ([@"NSCollectionViewLayout" isEqualToString: key])
         {
           hasValue = [currentElement elementForKey: @"collectionViewLayout"] != nil;
--- a/Source/NSCollectionView.m
+++ b/Source/NSCollectionView.m
@@ -1511,7 +1511,7 @@ static NSString *_placeholderItem = nil;
 
 - (NSIndexPath *) indexPathForItem: (NSCollectionViewItem *)item
 {
-  return [_itemsToIndexPaths objectForKey: item]; // h];
+  return [_itemsToIndexPaths objectForKey: item];
 }
 
 - (NSIndexPath *) indexPathForItemAtPoint: (NSPoint)point
--- a/Source/NSCollectionView.m
+++ b/Source/NSCollectionView.m
@@ -1136,24 +1136,7 @@ static NSString *_placeholderItem = nil;
             }
           else if ([theEvent modifierFlags] & NSShiftKeyMask)
             {
-              /*
-              id firstSelected = [currentSet firstObject];
-              NSRange selectedRange;
-              
-              if (firstSelected == nil)
-                {
-                  selectedRange = NSMakeRange(index, index);
-                }
-              else if (index < firstSelectedIndex)
-                {
-                  selectedRange = NSMakeRange(index, (firstSelectedIndex - index + 1));
-                }
-              else
-                {
-                  selectedRange = NSMakeRange(firstSelectedIndex, (index - firstSelectedIndex + 1));
-                }
-              [currentIndexSet addIndexesInRange: selectedRange];
-              */
+	      // TODO: Implement multiple selection logic
               [self setSelectionIndexPaths: currentSet];
             }
         }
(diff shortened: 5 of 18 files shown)


gcasa:
Please find the attached xib and enumerated nib.   The test for this is at git@github.com/gcasa/NSCollectionView_test.git

[MainMenu.xib.txt](https://github.com/gnustep/libs-gui/files/6652943/MainMenu.xib.txt)
[MainMenu-enumerated.nib.txt](https://github.com/gnustep/libs-gui/files/6652944/MainMenu-enumerated.nib.txt)

.travis.yml
```diff
@@ -70,4 +70,5 @@ before_script: >
     . $HOME/staging/share/GNUstep/Makefiles/GNUstep.sh;
 script: >
     ./configure $BASE_ABI || (cat config.log && false);
-    make && make install && make check || (cat Tests/tests.log && false);
+    make && make install
+    # make check || (cat Tests/tests.log && false);
```
fredkiefer, line 74: Why is this included in this pull request? Did you have issues with the base tests?
At least we should do this separately from this PR.

gcasa: Arghhh... I used git commit -am ... instead of -m.   I'll correct it.

Headers/AppKit/NSCollectionView.h
```diff
@@ -34,9 +37,12 @@
 #import <AppKit/NSDragging.h>
 #import <AppKit/NSNibDeclarations.h>
 #import <AppKit/NSView.h>
+#import <AppKit/NSPasteboard.h>
```
fredkiefer, line 40: Wouldn't it be sufficient to declare the class here?

.travis.yml
gcasa: Off topic: I don't think this change originated on my branch.  I had done a rebase from master that may have picked this up.   Also, I don't think we use the travis.yml stuff any more since github made a change moving away from it.

gcasa: `3627b22481 (fredkiefer             2019-12-06 17:32:25 +0100 72)     ./configure $BASE_ABI || (cat config.log && false);
94cc250d2d (Gregory John Casamento 2021-07-15 11:36:08 -0400 73)     make && make install
94cc250d2d (Gregory John Casamento 2021-07-15 11:36:08 -0400 74)     # make check || (cat Tests/tests.log && false);`

Strange, I don't recall making this change.  Nevertheless, I am obliterating it.

fredkiefer: At least it is not on the master branch.

gcasa: I have removed the changes.

Headers/AppKit/NSCollectionView.h
gcasa: I added forward declarations and removd the import.

gcasa:
[MainMenu-enumerated.txt](https://github.com/gnustep/libs-gui/files/8237467/MainMenu-enumerated.txt)

In this nib, it appears as though Apple has created a class named NSCollectionViewCore to handle some NSCollectionView responsibilities.     Also, it seems that they are using UICollectionView as well.   You will see what I am talking about near index 68-71.   This nib is "annotated" I used my ruby script to do this, so it is not loadable, just easier to read. :) 

This comes from my test for this in the repo [here](https://github.com/gcasa/NSCollectionViewLayout_test.git)

One of the good things about keyed nibs is that we don't necessarily need to follow everything they are doing.

@fredkiefer

Headers/AppKit/NSCollectionView.h
```diff
@@ -37,6 +40,22 @@
...
+@class NSCollectionViewLayoutAttributes;
+@class NSCollectionViewTransitionLayout;
+@class NSPasteboard;
+@class NSNib;
+@class NSMapTable;
+@class NSMutableArray;
+@class NSMutableDictionary;
+@class NSMutableSet;
+
+@class NSMapTable;
```
fredkiefer, line 53: This whole block is duplicated

Headers/AppKit/NSCollectionView.h
```diff
@@ -37,6 +40,22 @@
 
 @class NSCollectionViewItem;
 @class NSCollectionView;
+@class NSCollectionViewLayout;
+@class NSCollectionViewLayoutAttributes;
+@class NSCollectionViewTransitionLayout;
+@class NSPasteboard;
+@class NSNib;
+@class NSMapTable;
+@class NSMutableArray;
```
fredkiefer, line 49: You don't need any of these mutable classes, why do you define these?

Headers/AppKit/NSCollectionViewLayout.h
```diff
@@ -0,0 +1,258 @@
...
+@class NSCollectionViewLayoutAttributes;
+@class NSCollectionView;
+@class NSDictionary;
+@class NSIndexPath;
+@class NSNib;
+@class NSSet;
+
+@interface NSCollectionViewLayoutAttributes : NSObject <NSCopying>
+{
+  NSRect _frame;
```
fredkiefer, line 61: Are you sure that these two values frame and size are completely independent of each other? I would have expected size to be the size of the frame.

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -3563,11 +3569,19 @@ - (BOOL) containsValueForKey: (NSString *)key
         {
           hasValue = [self containsValueForKey: [XmlKeyMapTable objectForKey: key]];
         }
+      else if ([@"NSBackgroundColors" isEqualToString: key])
```
fredkiefer, line 3572: Which value is it? "NSCollectionViewBackgroundColor" or "NSBackgroundColors"? It is highly unlikely that you will have to use a different value for these two methods.
Anyway, why didn't you just add this mapping to `XmlKeyMapTable`?

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -3563,11 +3569,19 @@ - (BOOL) containsValueForKey: (NSString *)key
...
+        {
+          hasValue = [currentElement elementForKey: @"primaryBackgroundColor"] != nil;
+        } 
+      else if ([@"NSCollectionViewLayout" isEqualToString: key])
```
fredkiefer, line 3576: This case should work out of the box. Are you sure you need this code?

Source/NSCollectionView.m
```diff
@@ -629,6 +709,7 @@ - (id) initWithCoder: (NSCoder *)aCoder
           decode_NSUInteger(aCoder, &_maxNumberOfColumns);
           [aCoder decodeValueOfObjCType: @encode(BOOL) at: &_isSelectable];
           [self setBackgroundColors: [aCoder decodeObject]]; // decode color...
+          [self setCollectionViewLayout: [aCoder decodeObject]];
```
fredkiefer, line 712: Most likely there are no gorm files for this class, but normally we would change the version of the class and make the decoding version dependent.

Source/NSCollectionView.m
```diff
@@ -1114,4 +1199,382 @@ - (BOOL) wantsPeriodicDraggingUpdates
...
+        forSupplementaryViewOfKind: GSNoSupplementaryElement
+       withIdentifier: identifier];
+}
+
+- (void) registerNib: (NSNib *)nib 
+         forItemWithIdentifier: (NSUserInterfaceItemIdentifier)identifier
+{
+  [self registerNib: nib
+        forSupplementaryViewOfKind: GSNoSupplementaryElement
+     withIdentifier: identifier];
```
fredkiefer, line 1275: Strange indentation

Source/NSCollectionView.m
```diff
@@ -1114,4 +1199,382 @@ - (BOOL) wantsPeriodicDraggingUpdates
...
+/* Providing the collection view's data */
+
+- (id<NSCollectionViewDataSource>) dataSource
+{
+  return _dataSource;
+}
+
+- (void) setDataSource: (id<NSCollectionViewDataSource>)dataSource
+{
+  NSLog(@"Adding data source %@", dataSource);
```
fredkiefer, line 1328: Is this log statement on purpose?

renjwjx:
Hi, I am interested in this implementation about layout,  what's the plan for the implementation of NSCollectionViewFlowLayout. Can I suppose it is the same design as Apple AppKit?

Source/NSCollectionView.m
gcasa: It seems to be how emacs wanted to place it, but I have aligned it.

Source/NSCollectionView.m
gcasa: Removed

Headers/AppKit/NSCollectionView.h
gcasa: These are used in a number of places now in the code, so I'm not sure why you're saying they are unecessary.  I think this comment is outdated.

Source/NSCollectionView.m
gcasa: Keysight has been using this class for a while, not from Gorm, though.  I just updated it since it is harmless to do so.

Source/GSXib5KeyedUnarchiver.m
gcasa:
```
    <key>NSBackgroundColors</key>
	<dict>
		<key>CF$UID</key>
		<integer>29</integer>
	</dict>
```
I believe it's NSBackgroundColors because that is what is in the nib representation of this.  Moved to the XmlKeyMapTable.

gcasa:
```
 NO, representedObject = (null)
2022-09-24 07:54:11.108 NSCollectionViewLayout_test[120916:120916] Exception occurred while loading model: Can not determine type information for -[GSNamedColor count]
2022-09-24 07:54:11.108 NSCollectionViewLayout_test[120916:120916] Failed to load Xib
2022-09-24 07:54:11.108 NSCollectionViewLayout_test[120916:120916] Cannot load the main model file 'MainMenu'
```
Mapping it gets the above error, tracing down why, but I may re-instate the previous change temporarily.

Source/GSXib5KeyedUnarchiver.m
gcasa: When I remove this the layout is no longer rendered.  I'm not sure why.   I'm leaving this in for the time being.

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -324,6 +326,7 @@ + (void) initialize
                                            @"beginningViews", @"NSStackViewBeginningContainer",  // NSStackView
                                            @"middleViews", @"NSStackViewMiddleContainer",
                                            @"endViews", @"NSStackViewEndContainer",
+                                           @"primaryBackgroundColor", @"NSBackgroundColors",
```
advcsi, line 329: The XmlKeyMapTable was really for defining a direct key mapping for names that were not easily translated.

NSBackgroundColors was originally an array that included the primary color (primaryBackgroundColor), and if defined, the secondary color (secondaryBackgroundColor).  Therefore, it was mapped to a decoder function (decodeBackgroundColorsForElement) in XmlKeyToDecoderSelectorMap in order to process the necessary attributes.  Adding this definition will break that processing.  Has something changed with the Apple NIBs?

Source/GSXib5KeyedUnarchiver.m
advcsi: This is technically redundant (as is line 3328), as the name translation SHOULD take care of these.  

Of course, unless the processing sequence has been broken somewhere.

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -3318,6 +3321,14 @@ - (id) decodeObjectForKey: (NSString *)key
           // New xib stores values as attributes...
           object = [currentElement attributeForKey: key];
         }
+      else if ([@"NSCollectionViewBackgroundColor" isEqualToString: key])
```
advcsi, line 3324: NSCollectionViewBackgroundColor (line 3324), if it is TRULY mapped to one attribute named primaryBackgroundColor, could be added to the XmlKeyMapTable to be translated.

Of course, unless the processing sequence has been broken somewhere.

Source/GSXib5KeyedUnarchiver.m
gcasa: No. Nothing has changed.  I noticed this breakage and reverted that change.

Source/GSXib5KeyedUnarchiver.m
gcasa: I don't think that's the problem in this case.

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -3539,11 +3550,19 @@ - (BOOL) containsValueForKey: (NSString *)key
...
+      else if ([@"NSCollectionViewLayout" isEqualToString: key])
+        {
+          hasValue = [currentElement elementForKey: @"collectionViewLayout"] != nil;
+        } 
       else if (([@"NSIntercellSpacingHeight" isEqualToString: key]) ||
                ([@"NSIntercellSpacingWidth" isEqualToString: key]))
         {
           hasValue = [currentElement elementForKey: @"intercellSpacing"] != nil;
-        }
+        }
```
fredkiefer, line 3565: Trailing space

Headers/AppKit/NSCollectionViewLayout.h
gcasa: I changed this to use size in the _frame var.

Source/GSXib5KeyedUnarchiver.m
gcasa: Fixed

Source/GSXib5KeyedUnarchiver.m
gcasa: The mapping method does the following:

screenshot

So it is not a direct mapping.   I believe that, by default, it filters out the entire class name instead of just NS.  In this case we just need NS filtered out, thus the code.   Of course I think there is one other way to do this using the dictionary that is hard coded above.  I've been thinking about moving that into a file that is read as a resource of the bundle (a thought for another time and, likely, not a good idea).

gcasa: I'm closing this... reopen, if there is an objection to the above reasoning.

gcasa:
@fredkiefer Please review and comment when you can. :). I am going to extend my Test at 
[
](https://github.com/gcasa/NSCollectionViewLayout_test)

Source/GSXib5KeyedUnarchiver.m
advcsi: I see the problem, as I have the original GNUstep code that I implemented here, and just compared it to the community version.

It turns out that when the community accepted the code it made some changes, particular to lines 

      else if ([key hasPrefix: @"NS"])
        {
          // Try a key minus a (potential) NS prefix...
          NSString *newKey = [self alternateName: key];
          hasValue = [self containsValueForKey: newKey];
        }

which originally was:

      else if ([key hasPrefix:@"NS"])
        {
          // Try a key minus a (potential) NS prefix...
          NSString *newKey = [key stringByDeletingPrefix:@"NS"];
          newKey = [[[newKey substringToIndex:1] lowercaseString] stringByAppendingString:[newKey substringFromIndex:1]];
          hasValue = [self containsValueForKey:newKey];
        }

All the original code was doing was remove the “NS” prefix, and accept the rest of the word as a key with the first letter lowercased.

The new code will not only remove the NS prefix but also anything associated with the potential key, the key in this case probably "collectionViewLayout".  The “alternateName” method changes the NSCollectionViewLayout to "", so I see what you’re saying.  That is, if I follow the code correctly.  Thus, your need to add an explicit entry translation.

I am not sure why that code section was changed, but I suspect it was either due to an attempt at optimization OR a misunderstanding of the original intent was for it.   I guess there could have been some valid reason but since I wasn't involved I can only guess.  It suspect it will be a source of confusion and contention in there going forward as I am not sure of even how many things it could have potentially have broken based on what I was doing, and thus all the new entries in that section.

Anyway, sorry for my original comment(s) on the code section as I hadn't followed through what code the community had actually decided to accept, thinking it was pretty much mostly what I had put together.  I just went through the community code and compared it against what I had originally done, thus finding this discrepancy.  I will try to avoid such confusion going forward if I decide to comment.

gcasa: No need for apologies... your input is always appreciated, of course.  I don't think it was due to a misunderstanding of the original intent.  A lot of the time (majority of cases) the key ends up being mapped from NSClassNameKeyName => keyName, so it was a sensible change.   As we all know, however, Apple sometimes breaks its own conventions.

fredkiefer: I understand why this isn't working, but we really should come up with a better solution. Either add this to the XmlKeyMapTable, which is always consulted first. Or we could change the method alternateName: to return a list of names that should be tried one after another.
And remember that all the special cases present here need also to be handled in containsValueForKey:.

Source/NSCollectionView.m
```diff
@@ -26,37 +29,81 @@ Frank LeGrand (frank.legrand@testplant.com)
...
+#import "AppKit/NSNib.h"
 #import "AppKit/NSPasteboard.h"
 #import "AppKit/NSWindow.h"
+
 #import "GSGuiPrivate.h"
+#import "GSFastEnumeration.h"
 
 #include <math.h>
 
 static NSString* NSCollectionViewMinItemSizeKey              = @"NSMinGridSize";
```
fredkiefer, line 59: Don't all these values now require APPKIT_DECLARE?

Source/NSCollectionView.m
```diff
@@ -26,37 +29,81 @@ Frank LeGrand (frank.legrand@testplant.com)
...
+@end
+
+@implementation NSObject (CollectionViewInternalPrivate)
+
+- (BOOL) overridesSelector: (SEL)s
+{
+  Class osc = [self superclass];
+  BOOL overriden = NO;
+
+  while(osc != nil)
```
fredkiefer, line 92: Here a space is missing.

Source/NSCollectionView.m
```diff
@@ -26,37 +29,81 @@ Frank LeGrand (frank.legrand@testplant.com)
...
+ * This method is used to determine if the receiver overrides the
+ * given selector.
+ */
+- (BOOL) overridesSelector: (SEL)s;
+
+@end
+
+@implementation NSObject (CollectionViewInternalPrivate)
+
+- (BOOL) overridesSelector: (SEL)s
```
fredkiefer, line 87: Are you sure you need the full complexity of this method here? It seems to get used only once. You could have a look at the way I solved a similar issue in NSDocument.m line 88.

Source/NSCollectionView.m
```diff
@@ -351,7 +549,69 @@ - (void) setSelectable: (BOOL)flag
               [item setSelected:NO];
```
fredkiefer, line 549: Here a space is missing after the colon

Source/NSCollectionView.m
```diff
@@ -525,12 +807,12 @@ - (void) tile
   NSSize itemSize = NSMakeSize(_minItemSize.width, _minItemSize.height);
   
   _numberOfColumns = MAX(1.0, floor(width / itemSize.width));
-
+
```
fredkiefer, line 810: There seem to be a few lines with trailing spaces.

Source/NSCollectionView.m
```diff
@@ -603,35 +885,68 @@ - (id) initWithCoder: (NSCoder *)aCoder
...
+            {
+              _isSelectable = [aCoder decodeBoolForKey: NSCollectionViewSelectableKey];
+            }
+          
+          if ([aCoder containsValueForKey: NSCollectionViewAllowsMultipleSelectionKey])
+            {
+              _allowsMultipleSelection = [aCoder decodeBoolForKey: NSCollectionViewAllowsMultipleSelectionKey];
+            }
+          
+          if ([aCoder containsValueForKey: NSCollectionViewBackgroundColorsKey])
```
fredkiefer, line 923: Not sure why this is working as you did not override the containsValueForKey: method for this key.

Source/NSCollectionView.m
```diff
@@ -664,6 +979,9 @@ - (void) encodeWithCoder: (NSCoder *)aCoder
       //[aCoder encodeCGFloat: _verticalMargin forKey: NSCollectionViewVerticalMarginKey];
       [aCoder encodeObject: _backgroundColors 
                     forKey: NSCollectionViewBackgroundColorsKey];
+
```
fredkiefer, line 982: Looks like you are not encoding most of the new values you decode.

Source/NSCollectionView.m
```diff
@@ -785,39 +1112,98 @@ - (NSInteger) _indexAtPoint: (NSPoint)point
...
+                }
+              else
+                {
+                  [currentSet addObject: index];
+                }
+              [self setSelectionIndexPaths: currentSet];
+            }
+          else if ([theEvent modifierFlags] & NSShiftKeyMask)
+            {
+              /*
```
fredkiefer, line 1139: In general left over commented out code is a bad thing.

Source/NSCollectionView.m
```diff
@@ -1114,4 +1500,556 @@ - (BOOL) wantsPeriodicDraggingUpdates
...
+}
+
+- (NSSet *) indexPathsForVisibleSupplementaryElementsOfKind: (NSCollectionViewSupplementaryElementKind)elementKind
+{
+  return nil;
+}
+
+- (NSIndexPath *) indexPathForItem: (NSCollectionViewItem *)item
+{
+  return [_itemsToIndexPaths objectForKey: item]; // h];
```
fredkiefer, line 1531: A rather strange comment

Source/NSCollectionView.m
```diff
@@ -1114,4 +1500,556 @@ - (BOOL) wantsPeriodicDraggingUpdates
...
+  _dataSource = dataSource;
+  [self reloadData];
+}
+
+/* Configuring the Collection view */
+
+- (NSNib *) _nibForClass: (Class)cls
+{
+  NSString *clsName = NSStringFromClass(cls);
+  NSNib *nib = [[NSNib alloc] initWithNibNamed: clsName
```
fredkiefer, line 1675: Looks like this leaks the NSNib object

Source/NSCollectionView.m
```diff
@@ -1114,4 +1500,556 @@ - (BOOL) wantsPeriodicDraggingUpdates
...
+          frame.size = sz;
+          [v setFrame: frame];
+          [v setHidden: hidden];
+          [v setAlphaValue: alpha];
+
+          [_itemsToAttributes setObject: attrs
+                                 forKey: item];
+          [_itemsToIndexPaths setObject: path
+                                 forKey: item];
+          [_indexPathsToItems setObject: item
```
fredkiefer, line 1744: Why here and not in makeItemWithIdentifier: forIndexPath:?

Source/NSCollectionView.m
```diff
@@ -1114,4 +1500,556 @@ - (BOOL) wantsPeriodicDraggingUpdates
...
+        }
+    }
+}
+
+- (void) _loadSectionAtIndex: (NSUInteger)cs
+{
+  NSInteger ni = [self numberOfItemsInSection: cs];
+  NSInteger ci = 0;
+  
+  NSDebugLog(@"current section = %ld", cs);
```
fredkiefer, line 1761: This should be removed or turned into a class specific log.

Source/NSCollectionView.m
```diff
@@ -1114,4 +1500,556 @@ - (BOOL) wantsPeriodicDraggingUpdates
...
+  NSUInteger c = 0;
+  NSUInteger i = 0;
+  
+  c = [sections getIndexes: buffer
+                  maxCount: [sections count]
+              inIndexRange: NULL];
+
+  if (buffer != NULL)
+    {
+      for(i = 0; i < c; i++)
```
fredkiefer, line 1874: Here a space is missing.

Source/NSCollectionView.m
```diff
@@ -1114,4 +1500,556 @@ - (BOOL) wantsPeriodicDraggingUpdates
...
+{
+  _allowsEmptySelection = flag;
+}
+
+- (IBAction) selectAll: (id)sender
+{
+  NSEnumerator *en = [_itemsToIndexPaths keyEnumerator];
+  id obj = nil;
+
+  while((obj = [en nextObject]) != nil)
```
fredkiefer, line 1978: A space is missing

Source/NSCollectionView.m
```diff
@@ -1114,4 +1500,556 @@ - (BOOL) wantsPeriodicDraggingUpdates
...
+        }
+    }
+}
+
+- (IBAction) deselectAll: (id)sender
+{
+  NSEnumerator *en = [_itemsToIndexPaths keyEnumerator];
+  id obj = nil;
+
+  while((obj = [en nextObject]) != nil)
```
fredkiefer, line 1992: Again a missing space.

Source/NSCollectionView.m
```diff
@@ -1114,4 +1500,556 @@ - (BOOL) wantsPeriodicDraggingUpdates
...
+        {
+          [obj setSelected: NO];
+        }
+    }  
+}
+
+- (void) selectItemsAtIndexPaths: (NSSet *)indexPaths 
+                  scrollPosition: (NSCollectionViewScrollPosition)scrollPosition
+{
+  FOR_IN(NSIndexPath*,p,indexPaths)
```
fredkiefer, line 2004: The next few FOR_INs would benefit from a few more spaces

fredkiefer (changes requested):
This PR is way too big to review. I tried to look at a few methods and comment on these. As soon as this gets corrected and I find more time I will try to go through the rest

gcasa:
It took almost 1.5 years to do (on and off) so an overnight review is not expected.  I just wanted to let you know it's in a state where it can be reviewed. Take your time it's important to get it right.

In retrospect it could likely have been done in stages but the change was so big to begin with (they basically rewrote the whole class and left the legacy code in place) that it was difficult to say where the lines should have been drawn. 

As it is I am leaving the NSCollectionViewCompositionalLayout class as a stub in this PR since it would only add tons more and is a very complex class. GC

gcasa:
@fredkiefer as always, you have a keen eye. Thanks for your comments thus far.  I will get to them asap.

Source/GSXib5KeyedUnarchiver.m
gcasa: @fredkiefer Agreed.   I believe it should try removing "NS" first and then "NS{ClassName}" second.  I am not sure which is the most common case, but that does seem to be what's implied.  Let me know what you think.   I will implement this in this branch.  If you would like me to move it to a separate branch, I will do so.

Headers/AppKit/NSCollectionView.h
```diff
@@ -44,47 +58,384 @@ enum
...
   IBOutlet NSCollectionViewItem *itemPrototype;
   NSMutableArray *_items;
-  
+
+  // Selection management...
   BOOL _allowsMultipleSelection;
   BOOL _isSelectable;
+  BOOL _allowsEmptySelection;
   NSIndexSet *_selectionIndexes;
+  NSSet *_selectionIndexPaths;
```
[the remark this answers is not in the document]
gcasa: No, one is for NSIndexPaths.   I believe that they are fundamentally different.   If there is a way to store both in the same place, that would be better, but I did it this way because I wasn't sure if I could store the indexPaths in the selectionIndexes ivar.

Headers/AppKit/NSCollectionView.h
```diff
@@ -44,47 +58,384 @@ enum
...
-  IBOutlet id <NSCollectionViewDelegate> delegate;
-  
+
+  // Delegate and datasource
+  IBOutlet id <NSCollectionViewDelegate> _delegate;
+  IBOutlet id <NSCollectionViewDataSource> _dataSource;
+  IBOutlet NSView *_backgroundView;
+  IBOutlet id <NSCollectionViewPrefetching> _prefetchDataSource;
+
+  BOOL _backgroundViewScrollsWithContent;
```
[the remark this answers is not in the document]
gcasa: I have done some rearrangement towards this.

Headers/AppKit/NSCollectionViewLayout.h
```diff
@@ -0,0 +1,267 @@
...
+
+APPKIT_EXPORT_CLASS
+@interface NSCollectionViewLayoutAttributes : NSObject <NSCopying>
+{
+  NSRect _frame;
+  CGFloat _alpha;
+  NSInteger _zIndex;
+  BOOL _hidden; 
+  NSIndexPath *_indexPath;
+  NSCollectionElementCategory _representedElementCategory;
```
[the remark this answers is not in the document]
gcasa: I believe you're referring to _representedElementCategory and _representedElementKind.   These are currently not used by the existing layouts.  I believe they are there so that they can provide extra information for custom layouts.

Source/NSCollectionView.m
gcasa: I am wondering if it should not be moved to GNUstepBase/NSObject+GNUstepBase.*.   Since it is generally useful and, since you solved a similar issue (I haven't looked at NSDocument yet...) would it be better to move it to someplace central?

Source/NSCollectionView.m
gcasa: Removed

Source/NSCollectionView.m
gcasa: Actually, I am going to move them to externs.m.  I think they belong there.

Source/GSXib5KeyedUnarchiver.m
gcasa: I opted to do this the simple way by adding it to the XmlKeyMapTable.  I will add the more "generic" logic in a separate branch.

Source/NSCollectionView.m
gcasa: I should be encoding all of them.   The only new ones are the BackgroundColors and collectionViewLayout keys.

gcasa:
@fredkiefer Please take another look and respond to some of the questions.

Headers/AppKit/NSCollectionView.h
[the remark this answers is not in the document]
gcasa: I'm going to resolve this one.

gcasa:
@fredkiefer I am going to attempt to address the remainder of your concerns on this PR.  Please let's try to do a review sometime before the meeting this weekend.    Thanks, GC

Headers/AppKit/NSFont.h
```diff
@@ -74,7 +74,7 @@ typedef enum _NSFontRenderingMode
 } NSFontRenderingMode;
 #endif
 
-APPKIT_EXPORT const CGFloat *NSFontIdentityMatrix;
+APPKIT_EXPORT const CGFloat NSFontIdentityMatrix[6];
```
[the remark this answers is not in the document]
gcasa: I found this issue when I was building on my i386 distro on my older black MacBook.   I don't believe it causes any functional difference.   It's not strictly related to the PR, but was only encountered while building.   I'm going to go ahead and resolve this unless you believe there is a compelling reason not to include it.

Source/NSCollectionView.m
gcasa: Added analogous macro OVERRIDDEN per suggestion. Resolving.

Source/NSCollectionView.m
gcasa: Could you please clarify which ones you believe are not being encoded?  It looks to me like the same keys are being decoded in initWithCoder: as those encoded in encodeWithCoder: @fredkiefer

Source/NSCollectionView.m
gcasa: I don't believe the current test code exercises this.

Source/NSCollectionView.m
gcasa: Same with above.

Source/NSCollectionView.m
```diff
@@ -1114,4 +1456,558 @@ - (BOOL) wantsPeriodicDraggingUpdates
...
+  DESTROY(_itemsToIndexPaths);
+  DESTROY(_itemsToAttributes);
+  
+  // reallocate the maps...
+  _itemsToAttributes = RETAIN([NSMapTable strongToStrongObjectsMapTable]);
+  _itemsToIndexPaths = RETAIN([NSMapTable strongToStrongObjectsMapTable]);
+  _indexPathsToItems = RETAIN([NSMapTable strongToStrongObjectsMapTable]);
+}
+
+- (void) _setFrameWithoutTile: (NSRect)f
```
fredkiefer, line 1751: Is this needed as a separate method? I only did see one call and there it could be inlined.

Source/NSCollectionView.m
```diff
@@ -1114,4 +1456,558 @@ - (BOOL) wantsPeriodicDraggingUpdates
...
+  if (loaded == NO)
+    {
+      item = nil;
+      NSLog(@"Could not load model %@", nib);
+    }
+  
+  RELEASE(nib);
+  
+  // Add to maps...
+  [_itemsToIndexPaths setObject: indexPath
```
fredkiefer, line 1556: You should not add these if the loading failed.

Source/NSCollectionView.m
```diff
@@ -1114,4 +1456,558 @@ - (BOOL) wantsPeriodicDraggingUpdates
...
+/* Locating Items and Views */
+
+- (NSArray *) visibleItems
+{
+  return _visibleItems;
+}
+
+- (NSSet *) indexPathsForVisibleItems
+{
+  return _indexPathsForVisibleItems;
```
fredkiefer, line 1472: Again you seem to be keeping two variables with connected content. It would be better to just store one of these and compute the other on the fly. The methods `indexPathForItem:` and `itemAtIndexPath` would be helpful here.

Source/NSCollectionView.m
fredkiefer: Why did you resolve this? The NSNib object still does not get autoreleased here.

Source/NSCollectionView.m
```diff
@@ -1114,4 +1456,558 @@ - (BOOL) wantsPeriodicDraggingUpdates
...
+{
+  return _allowsEmptySelection;
+}
+
+- (void) setAllowsEmptySelection: (BOOL)flag;
+{
+  _allowsEmptySelection = flag;
+}
+
+- (IBAction) selectAll: (id)sender
```
fredkiefer, line 1931: I don't see how any of the following methods would affect `_selectionIndexPaths`

fredkiefer (changes requested):

Source/NSCollectionView.m
gcasa: You're right they don't... I will fix this and resolve.

gcasa: Fixed and pushed.

Source/NSCollectionView.m
gcasa: fixed, also removed release because the nib is being AUTORELEASED per the previous change.

Headers/AppKit/NSCollectionView.h
```diff
@@ -44,47 +58,401 @@ enum
...
 @interface NSCollectionView : NSView //<NSDraggingDestination, NSDraggingSource>
 {
+  // Content
   NSArray *_content;
-  IBOutlet NSCollectionViewItem *itemPrototype;
   NSMutableArray *_items;
-  
-  BOOL _allowsMultipleSelection;
-  BOOL _isSelectable;
   NSIndexSet *_selectionIndexes;
```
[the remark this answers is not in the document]
gcasa: The don't need to be.  I thought we had covered this.   Basically with these changes we, essentially,  have two separate versions of this class in ONE file.   One handles things using an array controller and indexes and the other uses indexPaths.   So I don't think they need to be kept in sync...  I will test this, but I doubt there is any relationship between the two.

Headers/AppKit/NSCollectionView.h
```diff
@@ -44,47 +58,401 @@ enum
...
+  IBOutlet id <NSCollectionViewDelegate> _delegate;
+  IBOutlet id <NSCollectionViewDataSource> _dataSource;
+  IBOutlet NSView *_backgroundView;
+  IBOutlet id <NSCollectionViewPrefetching> _prefetchDataSource;
+
+  // Layout
+  NSCollectionViewLayout *_collectionViewLayout;
+
+  // Managing items
+  NSMutableArray *_visibleItems;
```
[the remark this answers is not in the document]
gcasa: I am now only keeping _visibleItems.   The other is generated.

Headers/AppKit/NSCollectionView.h
```diff
@@ -44,47 +58,401 @@ enum
...
+ * Allows application to specify the screenPoint at which a dragging session will begin
+ * for the given indexes
+ */
+- (void) collectionView: (NSCollectionView *)collectionView
+        draggingSession: (NSDraggingSession *)session
+       willBeginAtPoint: (NSPoint)screenPoint
+      forItemsAtIndexes: (NSIndexSet *)indexes;
+
+/**
+ * Allows application to specify the screenPoint at which a dragging session will end
```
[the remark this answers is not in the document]
gcasa: changed

Headers/AppKit/NSCollectionView.h
```diff
@@ -44,47 +58,401 @@ enum
...
+
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_11, GS_API_LATEST)
+@protocol NSCollectionViewPrefetching <NSObject>
+#if GS_PROTOCOLS_HAVE_OPTIONAL
+@required
+#endif
+/**
+ * Load the items listed in indexPaths in collectionView before they are displayed
+ */
+- (void)collectionView:(NSCollectionView *)collectionView prefetchItemsAtIndexPaths:(NSArray *)indexPaths;
```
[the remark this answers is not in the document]
gcasa: Should be corrected.

Headers/AppKit/NSCollectionViewLayout.h
```diff
@@ -0,0 +1,267 @@
...
+  finalLayoutAttributesForDisappearingSupplementaryElementOfKind: (NSCollectionViewSupplementaryElementKind)elementKind
+                                                     atIndexPath: (NSIndexPath *)elementIndexPath;
+- (NSCollectionViewLayoutAttributes *)
+  initialLayoutAttributesForAppearingDecorationElementOfKind: (NSCollectionViewDecorationElementKind)elementKind
+                                                 atIndexPath: (NSIndexPath *)decorationIndexPath;
+- (NSCollectionViewLayoutAttributes *)
+  finalLayoutAttributesForDisappearingDecorationElementOfKind: (NSCollectionViewDecorationElementKind)elementKind
+                                                   atIndexPath: (NSIndexPath *)decorationIndexPath;
+
+- (NSSet *)indexPathsToDeleteForSupplementaryViewOfKind: (NSCollectionViewSupplementaryElementKind)elementKind;
```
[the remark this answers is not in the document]
gcasa: Fixed

Headers/AppKit/NSCollectionView.h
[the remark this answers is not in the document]
gcasa:
```
Enabled
2023-02-04 10:20:49.810566-0500 NSCollectionViewLayout_test[41075:3382853] indexPaths = _NSCollectionViewMutableIndexPathSet {1:(4)}
2023-02-04 10:20:49.810717-0500 NSCollectionViewLayout_test[41075:3382853] *** Assertion failure in NSIndexSet *_NSIndexSetFromCollectionViewIndexPaths(NSSet<NSIndexPath *> *)_block_invoke(), NSCollectionView.m:9000
2023-02-04 10:20:49.811025-0500 NSCollectionViewLayout_test[41075:3382853] [General] NSIndexPath(s) with section != 0
2023-02-04 10:20:49.816050-0500 NSCollectionViewLayout_test[41075:3382853] [General] (
	0   CoreFoundation                      0x00007ff80b29640b __exceptionPreprocess + 242
	1   libobjc.A.dylib                     0x00007ff80ade2e25 objc_exception_throw + 48
	2   Foundation                          0x00007ff80c5b9067 -[NSMutableDictionary(NSMutableDictionary) initWithContentsOfFile:] + 0
	3   AppKit                              0x00007ff80edc5284
```

When NSIndexPaths are used, it apparently throws an exception when you even try to access the value of "selectionIndexes"... here is the code...

```

- (void)collectionView:(NSCollectionView *)collectionView didSelectItemsAtIndexPaths:(NSSet<NSIndexPath *> *)indexPaths
{
    NSLog(@"selectionIndexes: %@", [self.collectionView selectionIndexes]);
    NSLog(@"selectionIndexPaths: %@", [self.collectionView selectionIndexPaths]);
}
@end
```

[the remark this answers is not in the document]
gcasa: I just implemented the call for the "collectionView:didSelectItemsAtIndexPaths:" it just prints the values of both.   As I had mentioned, I don't think they need to be kept in sync.   If you want me to implement the exception, I will, but I think that is something we don't really need to do for GS.

Source/NSCollectionView.m
gcasa: Inlined.

gcasa:
@fredkiefer I believe I have adequately answered all concerns.

gcasa:
@fredkiefer Please take a look.

Source/NSCollectionView.m
fredkiefer: Most likely you closed this in error, mistaking it for the comment on line 526.

Headers/AppKit/NSCollectionView.h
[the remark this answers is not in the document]
fredkiefer: I don't see how the stack trace proves what you think. Why would Apple have a function for this conversion, when it is always bound to fail? But you were able to see your whole program and if you are sure these two variables are unrelated go ahead and implement it that way.

fredkiefer:
> @fredkiefer I believe I have adequately answered all concerns.

I see plenty of comment that where not addressed. If you want to ignore them go ahead and merge. Just don't state that they were answered. You might have to display the hidden items. Github is trying to compress the history.

Headers/AppKit/NSCollectionView.h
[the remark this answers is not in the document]
gcasa: I'm not sure if you have a copy of any version of macOS that contains the new enhancements. 

The original implementation of NSCollectionView only implemented one layout and kept track of things using NSIndexSet which merely tracks the the selected indices of the items not the indexPaths (which are different).   Additionally, the original implementation uses an NSArrayController and a content array. 

The enhancement which implements layouts and can support custom layouts tracks selections using a set of index paths and uses the delegate / data source approach.  

Can you figure out a way to reconcile index sets and index paths??  If so I would very much like to hear what your opinion is. Is there a compelling reason to keep them in sync?  I know this from looking at the header and playing with the implementation in the test which I wrong.  

Additionally, when using the array controller that exception that I showed you doesn't occur.

gcasa:
> > @fredkiefer I believe I have adequately answered all concerns.
> 
> 
> 
> I see plenty of comment that where not addressed. If you want to ignore them go ahead and merge. Just don't state that they were answered. You might have to display the hidden items. Github is trying to compress the history.

Instead of, yet again, being falsely accused of merging prematurely, I will review the changes again even though this PR has been pending for over a year and will attempt to make sure that your concerns have all been addressed.

Headers/AppKit/NSCollectionView.h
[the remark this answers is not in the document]
gcasa: > I don't see how the stack trace proves what you think. Why would Apple have a function for this conversion, when it is always bound to fail? But you were able to see your whole program and if you are sure these two variables are unrelated go ahead and implement it that way.

YOU are able to see my whole program too.  I have repeatedly stated where I put every single test I make every time I submit a PR.  Please review the history of this PR to find it as it has been stated at least one time.  

:/

Source/NSCollectionView.m
```diff
@@ -348,10 +519,72 @@ - (void) setSelectable: (BOOL)flag
           id item = [_items objectAtIndex: index];
           if ([item respondsToSelector: @selector(setSelected:)])
             {
-              [item setSelected:NO];
+              [item setSelected: NO];
             }
         }
+
+      FOR_IN(NSIndexPath*, i, _selectionIndexPaths)
```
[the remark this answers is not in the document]
gcasa: Disprove my assertion.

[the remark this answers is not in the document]
gcasa: I'm wondering why you feel, incorrectly, they are linked in some inane way.  Could you elaborate on their tenuous and imaginary connection in your mind?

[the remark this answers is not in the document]
gcasa: Also.  I if they are related then why have separate variables for them?!

gcasa:
> [MainMenu-enumerated.txt](https://github.com/gnustep/libs-gui/files/8237467/MainMenu-enumerated.txt)
> 
> In this nib, it appears as though Apple has created a class named NSCollectionViewCore to handle some NSCollectionView responsibilities. Also, it seems that they are using UICollectionView as well. You will see what I am talking about near index 68-71. This nib is "annotated" I used my ruby script to do this, so it is not loadable, just easier to read. :)
> 
> This comes from my test for this in the repo [here](https://github.com/gcasa/NSCollectionViewLayout_test.git)
> 
> One of the good things about keyed nibs is that we don't necessarily need to follow everything they are doing.
> 
> @fredkiefer

Just wanted to make sure you see this.  Again the test is here https://github.com/gcasa/NSCollectionViewLayout_test

Headers/AppKit/NSCollectionView.h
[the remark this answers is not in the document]
gcasa: https://github.com/gcasa/NSCollectionViewLayout_test

Source/NSCollectionView.m
gcasa:
```
- (NSSet *) indexPathsForVisibleItems
{
  NSMutableSet *result = [NSMutableSet setWithCapacity: [_visibleItems count]];

  FOR_IN(NSCollectionViewItem*, item, _visibleItems)
    {
      NSIndexPath *p = [_itemsToIndexPaths objectForKey: item];
      [result addObject: p];
    }
  END_FOR_IN(_visibleItems);
  
  return [NSSet setWithSet: result]; // return immutable copy
}
```
This is what is currently at line 1473-1485 in NSCollectionView.m

gcasa: I'm resolving this...  it has been addressed.

Source/NSCollectionView.m
```diff
@@ -1114,4 +1456,558 @@ - (BOOL) wantsPeriodicDraggingUpdates
...
+}
+
+- (NSArray *) visibleSupplementaryViewsOfKind: (NSCollectionViewSupplementaryElementKind)elementKind
+{
+  return [_visibleSupplementaryViews objectForKey: elementKind];
+}
+
+- (NSSet *) indexPathsForVisibleSupplementaryElementsOfKind: (NSCollectionViewSupplementaryElementKind)elementKind
+{
+  return nil;
```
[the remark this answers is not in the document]
gcasa: Implemented as such...
```
- (NSSet *) indexPathsForVisibleSupplementaryElementsOfKind: (NSCollectionViewSupplementaryElementKind)elementKind
{
  NSArray *views = [self visibleSupplementaryViewsOfKind: elementKind];
  NSMutableSet *indxs = [NSMutableSet setWithCapacity: [views count]];
  
  FOR_IN(id, item, views)
    {
      NSIndexPath *p = [self indexPathForItem: item];
      [indxs addObject: p];
    }
  END_FOR_IN(views);

  return [NSSet setWithSet: indxs];
}
```

Source/NSCollectionView.m
```diff
@@ -603,35 +858,68 @@ - (id) initWithCoder: (NSCoder *)aCoder
...
+            }
+          
+          if ([aCoder containsValueForKey: NSCollectionViewMaxItemSizeKey])
+            {
+              _maxItemSize = [aCoder decodeSizeForKey: NSCollectionViewMaxItemSizeKey];
+            }
+          
+          if ([aCoder containsValueForKey: NSCollectionViewMaxNumberOfRowsKey])
+            {
+              _maxNumberOfRows = [aCoder decodeInt64ForKey: NSCollectionViewMaxNumberOfRowsKey];
```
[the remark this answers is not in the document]
gcasa: Since this wasn't implemented by me, I am not sure why that is.   I think that decodeIntForKey should be sufficient.  I'll change it.

Source/NSCollectionView.m
```diff
@@ -26,37 +29,54 @@ Frank LeGrand (frank.legrand@testplant.com)
...
+APPKIT_DECLARE NSString* NSCollectionViewBackgroundColorsKey         = @"NSBackgroundColors";
+APPKIT_DECLARE NSString* NSCollectionViewLayoutKey                   = @"NSCollectionViewLayout";
+
+APPKIT_DECLARE NSCollectionViewSupplementaryElementKind GSNoSupplementaryElement  = @"GSNoSupplementaryElement"; // private
+
+/* 
+ * Private helper macro to check, if the method given via the selector sel 
+ * has been overridden in the current subclass.
+ */
+#define OVERRIDDEN(sel) ([self methodForSelector: @selector(sel)] != [[_collectionViewLayout class] instanceMethodForSelector: @selector(sel)])
```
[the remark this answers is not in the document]
gcasa: Corrected.

Source/NSCollectionView.m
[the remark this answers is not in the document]
gcasa: I am wondering why you consistently, like every single PR, miss the mention of the test I always put in the beginning of the PR... here it is again...

https://github.com/gcasa/NSCollectionViewLayout_test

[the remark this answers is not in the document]
gcasa: My code, as I tried to tell you in a previous comment, tries to print out the values of both when one is modified.  In the case of a collectionview that uses indexPaths and custom layouts, attempting to print the value for _selectionIndexes produces an exception.   I am not sure why it generates an exception, that seems overkill, but it does.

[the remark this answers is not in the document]
gcasa: FYI, for every class I have added there is a similar test in my personal repos.  Usually in the form ${CLASSNAME}_test or just ${CLASSNAME}.

Source/NSTextField.m
```diff
@@ -776,7 +776,6 @@ - (id) initWithCoder: (NSCoder*)aDecoder
   self = [super initWithCoder: aDecoder];
   if ([aDecoder allowsKeyedCoding])
     {
-      // do nothing for now...
```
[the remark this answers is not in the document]
gcasa: Should be gone.

Source/NSCollectionViewFlowLayout.m
```diff
@@ -0,0 +1,518 @@
...
+
+  // Calculations...
+  h = sz.height;
+  w = sz.width;
+  x = (r * w) + si.left + mis;
+  y = (s * h) + si.top + mls;
+  f = NSMakeRect(x, y, w, h);
+
+  // Determine if it is needed to reflow the given element...
+  if ((x + w) > (vf.size.width - w))
```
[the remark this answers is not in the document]
gcasa: Corrected.

Source/NSCollectionViewFlowLayout.m
```diff
@@ -0,0 +1,518 @@
...
+  [super prepareLayout];
+  _ds = 0;
+  _dr = 0;
+}
+
+- (NSCollectionViewLayoutAttributes *) layoutAttributesForItemAtIndexPath: (NSIndexPath *)indexPath
+{
+  NSCollectionViewLayoutAttributes *attrs = AUTORELEASE([[NSCollectionViewLayoutAttributes alloc] init]);
+  NSSize sz = NSZeroSize;
+  id <NSCollectionViewDelegateFlowLayout> d = (id <NSCollectionViewDelegateFlowLayout>)[_collectionView delegate];
```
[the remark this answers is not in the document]
gcasa: Done.  I have changed most, not, all to longer names.

Source/NSCollectionView.m
gcasa: This has been resolved.  I have eliminated as much whitespace as I could find.  I am tempted to write a script to do this automatically when needed.

gcasa:
@fredkiefer All I see left is our disagreement about _selectionIndexPaths and _selectionIndexes.   I will do further research to see if there is, indeed, a relationship between these two variables.   Please refer to the test I have referenced a number of times.

gcasa:
I just added this test... https://github.com/gcasa/NSCollectionViewArray_test it seems we have some deeper concerns.   This test works on macOS but does NOT work correctly on GNUstep. This means that Testplant (now Keysight) wrote this class so that it worked for Eggplant/EPF and didn't test it in the general case. SMDH... I am continuing to do research to see if those two variables are related, but this is an interesting detail to add.   I am not sure if I will take care of all of the issues with their implementation of the legacy NSCollectionView support in this PR.  We will save that for a later discussion.

gcasa:
I am wondering if I couldn't do something like this...
```
NSMutableIndexSet *indexSet = [NSMutableIndexSet indexSet];
for (NSIndexPath *indexPath in indexPathsSet) {
    [indexSet addIndex:indexPath.row];
}
NSIndexSet *convertedIndexSet = [indexSet copy];
```
I have noticed, in the exception, I sent you, that there is a suspiciously named function...
```
2023-02-11 11:23:46.959590-0500 NSCollectionViewLayout_test[35647:19643369] *** Assertion failure in NSIndexSet *_NSIndexSetFromCollectionViewIndexPaths(NSSet<NSIndexPath *> *)_block_invoke(), NSCollectionView.m:9000
```
It seems there might be a way to keep them in sync.  It's not obvious from the documentation that this is the case, but I don't think it hurts to implement it.  I have been unable to prove a relationship either way.

Headers/AppKit/NSCollectionView.h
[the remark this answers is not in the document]
gcasa: As discussed during the meeting and commented on below, Apple is, apparently, keeping these in sync.  The exception is from when the section is != 0, but does not appear otherwise.  I had assumed, incorrectly, that since it wasn't possible to keep them in sync in the general case that it wouldn't be done.  The Apple implementation issues an exception whereas the GNUstep code issues a log warning.   This difference is acceptable, I believe because the exception does not cause a crash and also because this functionality is entirely undocumented. GC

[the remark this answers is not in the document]
gcasa: I am resolving this for now, please resurrect it if you find any other issues.   I am hoping there aren't that many more things to review.   I am going to leave the fixes for the legacy functionality and the composite layout for separate PRs given that this PR is massive, and old, already.  GC

gcasa:
As additional proof

```
2023-02-12 11:31:58.746793-0500 NSCollectionViewLayout_test[71987:20745783] selectionIndexes: <NSMutableIndexSet: 0x600003527840>[number of indexes: 1 (in 1 ranges), indexes: (1)]
2023-02-12 11:31:58.746848-0500 NSCollectionViewLayout_test[71987:20745783] selectionIndexPaths: _NSCollectionViewMutableIndexPathSet {0:(1)}
``` 
when the datasource method numberOfSectionsInCollectionView: returns 1, the exception does not occur and, as you can see, they are synchronized.   The GS code does the same thing now with the methods I added yesterday.

I have modified the layout test [here ](https://github.com/gcasa/NSCollectionViewLayout_test) to illustrate this.  I have left a comment in it explaining.

Source/NSCollectionView.m
fredkiefer: Most other trailing whitespace are gone, but this specific is still there. Doesn't your editor have a command for removing these?

Source/NSCollectionView.m
```diff
@@ -1114,4 +1507,591 @@ - (BOOL) wantsPeriodicDraggingUpdates
...
+{
+  NSMutableSet *paths = [NSMutableSet setWithCapacity: [_itemsToIndexPaths count]];
+
+  FOR_IN(NSCollectionViewItem*, obj, _itemsToIndexPaths)
+    {
+      if ([obj respondsToSelector: @selector(setSelected:)])
+        {
+	  NSIndexPath *p = [_itemsToIndexPaths objectForKey: obj];
+
+          [obj setSelected: YES];
```
fredkiefer, line 2005: The actual selection already gets done in-setSelectionIndexPaths:, you don't need to do it here. But this is no big issue it would just make the code as a whole easier to understand if everything builds up on minimal methods.

Source/NSCollectionView.m
gcasa: I'm using emacs.  I just now ran cleanup-whitespace (I didn't know it had that feature) so, hopefully, that resolves everything.

gcasa: @fredkiefer Please let me know.  I just pushed.

Source/NSCollectionView.m
gcasa: I removed these.  This should be okay now.

Source/NSCollectionView.m
gcasa: I am resolving this as I believe all of the whitespace should be gone.

fredkiefer (approved):

Issue #108, opened 2021-09-24
--enable-imagemagick no longer works


eukara:
ImageMagick's BlobToImage can't seem to decode whatever blob it is receiving.

In order to compile with --enable-imagemagick, you also need to apply the following diff:

```diff --git a/Source/GSImageMagickImageRep.m b/Source/GSImageMagickImageRep.m
index fa98add60..f54898cc3 100644
--- a/Source/GSImageMagickImageRep.m
+++ b/Source/GSImageMagickImageRep.m
@@ -82,11 +82,11 @@
       NSSize res;
       if (image->units == PixelsPerCentimeterResolution)
 	{
-	  res = NSMakeSize(image->x_resolution * 2.54, image->y_resolution * 2.54);
+	  res = NSMakeSize(image->resolution.x * 2.54, image->resolution.y * 2.54);
 	}
       else
 	{
-	  res = NSMakeSize(image->x_resolution, image->y_resolution);
+	  res = NSMakeSize(image->resolution.x, image->resolution.y);
 	}
 
       if (res.width > 0 && res.height > 0)
@@ -121,7 +121,7 @@
   
   // Set the background color to transparent
   // (otherwise SVG's are rendered against a white background by default)
-  QueryColorDatabase("none", &imageinfo->background_color, exception);
+  QueryColorCompliance("none", AllCompliance, &imageinfo->background_color, exception);
 
   images = BlobToImage(imageinfo, [data bytes], [data length], exception);
 
```
The backend is important, as the internally supported formats alone are severely limiting.

-- Marco

eukara:
It would seem that it's only .tga files being unsupported.
.pcx files seem to decode just fine.

Oddly enough my system imagemagick tools handle .tga files flawlessly.
This is really bizarre.

PR #109, opened 2021-09-24
GSImageMagickImageRep: Update to work with ImageMagick 7

eukara:
QueryColorDatabase() has been deprecated it appears. (you'll get missing symbol errors...)
So this should address that.

Issue #108 still lists that some formats, like .TGA cannot be decoded. However because other format blobs get sent properly I'm not sure if that is just an issue with imageMagick itself.

### DIFF
--- a/Source/GSImageMagickImageRep.m
+++ b/Source/GSImageMagickImageRep.m
@@ -82,11 +82,19 @@
       NSSize res;
       if (image->units == PixelsPerCentimeterResolution)
 	{
+#if (MagickLibVersion >= 0x700)
 	  res = NSMakeSize(image->resolution.x * 2.54, image->resolution.y * 2.54);
+#else
+	  res = NSMakeSize(image->x_resolution * 2.54, image->y_resolution * 2.54);
+#endif
 	}
       else
 	{
+#if (MagickLibVersion >= 0x700)
 	  res = NSMakeSize(image->resolution.x, image->resolution.y);
+#else
+	  res = NSMakeSize(image->x_resolution, image->y_resolution);
+#endif
 	}
 
       if (res.width > 0 && res.height > 0)
@@ -121,7 +129,11 @@
   
   // Set the background color to transparent
   // (otherwise SVG's are rendered against a white background by default)
+#if (MagickLibVersion >= 0x700)
   QueryColorCompliance("none", AllCompliance, &imageinfo->background_color, exception);
+#else
+  QueryColorDatabase("none", &imageinfo->background_color, exception);
+#endif
 
   images = BlobToImage(imageinfo, [data bytes], [data length], exception);
 
--- a/Source/GSImageMagickImageRep.m
+++ b/Source/GSImageMagickImageRep.m
@@ -82,11 +82,11 @@
       NSSize res;
       if (image->units == PixelsPerCentimeterResolution)
 	{
-	  res = NSMakeSize(image->x_resolution * 2.54, image->y_resolution * 2.54);
+	  res = NSMakeSize(image->resolution.x * 2.54, image->resolution.y * 2.54);
 	}
       else
 	{
-	  res = NSMakeSize(image->x_resolution, image->y_resolution);
+	  res = NSMakeSize(image->resolution.x, image->resolution.y);
 	}
 
       if (res.width > 0 && res.height > 0)
@@ -121,7 +121,7 @@
   
   // Set the background color to transparent
   // (otherwise SVG's are rendered against a white background by default)
-  QueryColorDatabase("none", &imageinfo->background_color, exception);
+  QueryColorCompliance("none", AllCompliance, &imageinfo->background_color, exception);
 
   images = BlobToImage(imageinfo, [data bytes], [data length], exception);
 


fredkiefer:
I think that this change in itself fix correct, but it should only be applied for specific versions of ImageMagick. That means we need to detect the version of ImageMagick during configure and use either the old code or the new one depending on that version.

eukara:
Had some free time today, so I set out to test GNUstep GUI without my modifications against an older version of ImageMagick.
I started with the previous version 6 which comes installed with Slackware64 14.2. This was a fresh install. It compiles but doesn't link at all against the right libraries.

```
Making all for tool make_services...                                                                                                           
 Linking tool make_services ...                                                                                                                
/usr/lib64/libMagickCore-6.Q16.so: undefined reference to `DestroyPixelWand'                                                                   
/usr/lib64/libMagickCore-6.Q16.so: undefined reference to `PixelSetColor'                                                                      
/usr/lib64/libMagickCore-6.Q16.so: undefined reference to `DrawSetStrokeWidth'                                                                 
/usr/lib64/libMagickCore-6.Q16.so: undefined reference to `DrawRectangle'                                                                      
/usr/lib64/libMagickCore-6.Q16.so: undefined reference to `DrawPushPattern'                                                                    
/usr/lib64/libMagickCore-6.Q16.so: undefined reference to `DrawSetStrokeColor'                                                                 
/usr/lib64/libMagickCore-6.Q16.so: undefined reference to `DrawPathLineToAbsolute'                                                                                      
/usr/lib64/libMagickCore-6.Q16.so: undefined reference to `DrawRoundRectangle'                                                                 
/usr/lib64/libMagickCore-6.Q16.so: undefined reference to `DrawAnnotation'                                                                     
/usr/lib64/libMagickCore-6.Q16.so: undefined reference to `DrawPathStart'                                                                      
/usr/lib64/libMagickCore-6.Q16.so: undefined reference to `DrawSetClipPath'                                                                    
/usr/lib64/libMagickCore-6.Q16.so: undefined reference to `DrawPathClose'                                                                      
/usr/lib64/libMagickCore-6.Q16.so: undefined reference to `DrawSetViewbox'                                                                     
/usr/lib64/libMagickCore-6.Q16.so: undefined reference to `PixelSetRedQuantum'                                                                 
/usr/lib64/libMagickCore-6.Q16.so: undefined reference to `DrawPopClipPath'                                                                    
/usr/lib64/libMagickCore-6.Q16.so: undefined reference to `DrawSetStrokeAntialias'                                                                                      
/usr/lib64/libMagickCore-6.Q16.so: undefined reference to `DestroyMagickWand'                                                                  
/usr/lib64/libMagickCore-6.Q16.so: undefined reference to `DestroyDrawingWand'                                                                 
/usr/lib64/libMagickCore-6.Q16.so: undefined reference to `DrawColor'                                                                          
/usr/lib64/libMagickCore-6.Q16.so: undefined reference to `NewMagickWandFromImage'                                                                                      
/usr/lib64/libMagickCore-6.Q16.so: undefined reference to `DrawScale'                                                                          
/usr/lib64/libMagickCore-6.Q16.so: undefined reference to `DrawLine'                                                                           
/usr/lib64/libMagickCore-6.Q16.so: undefined reference to `PixelSetOpacityQuantum'                                                                                      
/usr/lib64/libMagickCore-6.Q16.so: undefined reference to `DrawRender'                                                                         
/usr/lib64/libMagickCore-6.Q16.so: undefined reference to `DrawSetStrokePatternURL'                                                                                         
/usr/lib64/libMagickCore-6.Q16.so: undefined reference to `DrawSetFillPatternURL'                                                                                   
/usr/lib64/libMagickCore-6.Q16.so: undefined reference to `DrawPushDefs'                                                                       
/usr/lib64/libMagickCore-6.Q16.so: undefined reference to `DrawComment'                                                                        
/usr/lib64/libMagickCore-6.Q16.so: undefined reference to `PixelSetBlueQuantum'                                                                
/usr/lib64/libMagickCore-6.Q16.so: undefined reference to `DrawSetFontSize'                                                                    
/usr/lib64/libMagickCore-6.Q16.so: undefined reference to `DrawSetStrokeDashArray'                                                                                      
/usr/lib64/libMagickCore-6.Q16.so: undefined reference to `DrawPathMoveToAbsolute'                                                                                      
/usr/lib64/libMagickCore-6.Q16.so: undefined reference to `DrawPushClipPath'                                                                   
/usr/lib64/libMagickCore-6.Q16.so: undefined reference to `DrawPopPattern' 
[continues]
```
I don't think the feature has been tested for a while if 2016's ImageMagick isn't properly supported by the build system either.
Which range of ImageMagick versions would you want to see supported?

fredkiefer:
This isn't actually a GNUstep issue. None of the mentioned functions gets called from GNUstep. It looks like ImageMagick 6 is not properly installed on that machine.

I still think using some code like this

#if  MAGICKCORE_CHECK_VERSION(7, 0, 0)
new code
#else
old code
#endif

would be nice.

eukara:
No problem, I just wanted to see when exactly this broke. I tried looking at the MagickCore headers for any reliable way to get the version (turns out their MagickLibInterface resets every major version increase or something) and I wanted to see it on an system that shipped with the appropriate version anyway. I guess the ImageMagick compile on Slackware 14.2 specifically has got something wonky going on then. It could have also been the build-system just not finding MagickWand and only MagickCore.

eukara:
Alright, I've pushed up the most reliable way I could find to check for the ImageMagick version, which is checking for MagickLibVersion.

fredkiefer (approved):

PR #110, opened 2021-09-26
Add protocols to NSComboBox.  Leave informal protocols in place.

gcasa:
Adding protocols for NSComboBoxDelegate and NSComboBoxDataSource.  I have left the informal protocols in place for compatibility.

### DIFF
--- a/Headers/AppKit/NSComboBox.h
+++ b/Headers/AppKit/NSComboBox.h
@@ -94,7 +94,11 @@
 #endif
 @end
 
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_6, GS_API_LATEST)
+@protocol NSComboBoxDataSource <NSObject>
+#else
 @interface NSObject (NSComboBoxDataSource)
+#endif
 - (NSInteger) numberOfItemsInComboBox: (NSComboBox *)aComboBox;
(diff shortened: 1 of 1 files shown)


Headers/AppKit/NSComboBox.h
```diff
@@ -106,13 +106,19 @@
 #endif
 @end
 
+@protocol NSComboBoxDataSource <NSObject>
```
fredkiefer, line 109: I don't see the point of having an empty protocol along a defined interface. Either replace the interface with the protocol completely or do something like 
#if OS_API_VERSION(MAC_OS_X_VERSION_10_6, GS_API_LATEST)
@protocol NSComboBoxDataSource <NSObject>
#else
@interface NSObject (NSComboBoxDataSource)
#endif

And the same is true for NSComboBoxDelegate

gcasa: The point would be to allow code which uses those protocols to compile, at least that was my initial thought.  The effect is basically the same, as all methods are @optional in both protocols.   I will change it in the way that you suggest (your second suggestion).  Thx... GC

gcasa: Fix committed.  I'm resolving this.

fredkiefer (approved):

PR #112, opened 2021-10-02
Improve NSPopover implementation

gcasa:
The NSPopover implementation needed improvement.  I have added code which makes the popover work more like the implementation in Cocoa.  I also added some cosmetic improvements.  The code also uses a subview to make the popover look better.

### DIFF
--- a/Source/NSPopover.m
+++ b/Source/NSPopover.m
@@ -42,9 +42,9 @@
 // Popover private classes
 
 @interface GSPopoverView : NSView
-{
-  NSView *_contentView;
-}
+
+- (void) setContentView: (NSView *)view;
+
 @end
 
 @implementation GSPopoverView
@@ -63,13 +63,8 @@
 {
   if (view != nil)
     {
-      ASSIGN(_contentView, view);
-      [_contentView setAutoresizingMask: NSViewWidthSizable | NSViewHeightSizable];
-      [self addSubview: _contentView];
-    }
-  else
-    {
-      [_contentView removeFromSuperview];
+      [view setAutoresizingMask: NSViewWidthSizable | NSViewHeightSizable];
+      [self addSubview: view];
     }
 }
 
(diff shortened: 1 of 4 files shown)


Source/NSPopover.m
```diff
@@ -34,101 +34,290 @@
...
+      [self setAutoresizingMask: NSViewWidthSizable | NSViewHeightSizable];
+    }
+  return self;
+}
+
+- (void) setContentView: (NSView *)view
+{
+  if (view != nil)
+    {
+      ASSIGN(_contentView, view);
```
fredkiefer, line 66: You should decide whether to store this in _contentView or to have it as a sub view. Doing both is inconsistent and will leak memory.

fredkiefer: Looking at the code further down in GSPopoverPanel you don't need any of this at all.

Source/NSPopover.m
```diff
@@ -34,101 +34,290 @@
...
+}
+
+@end
 
 /* Class */
 @implementation NSPopover
 
 /* Properties */
+
+/**
```
fredkiefer, line 151: Normally we would add such comments to the header file.

Source/NSPopover.m
```diff
@@ -34,101 +34,290 @@
...
+ */
 - (void) setContentViewController: (NSViewController *)controller
 {
+  if ([NSStoryboard mainStoryboard] == nil)
+    {
+      NSString *controllerClassName = NSStringFromClass([controller class]);
+      BOOL loaded = [NSBundle loadNibNamed: controllerClassName
+				     owner: controller];
+      if (!loaded)
+	{
```
fredkiefer, line 234: Indentations seems wrong here.

fredkiefer (changes requested):

Source/NSPopover.m
gcasa: I need the subview to draw the box around the view (as in GSToolTip.m).  Otherwise the popover looks a bit flat.   Which part do I not nee
popover
d?

Source/NSPopover.m
gcasa: Hmm... weird.  I will try to do untabify on it as it looks right in emacs.

gcasa:
Here is the test case I am using: https://github.com/gcasa/NSPopover_test

Source/NSPopover.m
```diff
@@ -34,13 +34,115 @@
...
+  return [subviews objectAtIndex: 0];
+}
+
+@end
 
 /* Class */
 @implementation NSPopover
 
 /* Properties */
+
```
fredkiefer, line 145: Looks like there are a few leftover newlines here.

Source/NSPopover.m
gcasa: Comments moved to header.

Source/NSPopover.m
fredkiefer: Still looks wrong, but this might be an artefact of GitHub.

gcasa: Fixed indention by using spaces.

Source/NSPopover.m
```diff
@@ -34,13 +34,115 @@
...
+  if (self)
+    {
+      [self setAutoresizingMask: NSViewWidthSizable | NSViewHeightSizable];
+    }
+  return self;
+}
+
+- (void) setContentView: (NSView *)view
+{
+  if (view != nil)
```
fredkiefer, line 64: I still don't think this method ever gets Calle. Could you please add an NSLog statement here and see if it gets printed out?

fredkiefer (approved):
There are a few minor remaining things otherwise this looks good to me.

Source/NSPopover.m
gcasa: You are correct.  I removed the method after testing it, as you suggested.

Source/NSPopover.m
gcasa: Removed.

PR #111, opened 2021-09-29
Merge test plant changes into GNUstep main GUI repo on the gnustep_testplant_branch

gcasa:
Like with base, go through all of these changes to see what can be done to merge them.

PR #113, opened 2021-10-05
All relevant changes from testplant/eggplant/keysight -- PLEASE REVIEW

gcasa:
*NOTE: We are in the process of extracting changes from this branch into other, smaller and more manageable pieces.  This should simplify the review process as smaller changes will not be blocked by some of the more complex/questionable changes.  These branches will be prefixed by TP_.  GC*

All relevant changes from testplant/eggplant/keysight since the divergence from gnustep_testplant_branch some comments are in https://github.com/gnustep/libs-gui/pull/111

I will need help updating this since it is such a large merge.  There are some good changes, but also some that I don't fully understand. 

In some places the community has COMPLETELY refactored the way something works (for instance some code in certain classes in GUI were moved to the GSTheme code and completely refactored) and in other cases the code has just been refactored in place and the changes on the testplant branch are rendered either obsolete or no longer mergeable by the algorithm used by git, so a manual merge has been done.   I am leaving some comments in the curated code to explain, they will likely be removed once the merge moves to the main/master branch.


Source/GSLayoutManager.m
```diff
@@ -3218,8 +3223,10 @@ - (void) insertGlyphs: (const NSGlyph*)glyph_list
   glyph_run_t *run;
   int i;
   unsigned int gpos, cpos;
-  NSSize advances[length];
-
+  // TESTPLANT-MAL-07312017: Would be nice to optimize this...
+  //     but we've been seeing stack overwrite crashes due to this...
+  NSSize *advances = NSZoneMalloc(NSDefaultMallocZone(), sizeof(NSSize)*length);
+
```
iamleeg, line 3229: This looks like it's working around a stack smashing vulnerability, by allocating `advances` onto its own page so you quietly write off the end of it. I'd suggest that there's a problem here in how `advances` is used, probably in `-insertGlyphs:withAdvancements:length:forStartingGlyphAtIndex:characterIndex:` that this change papers over.

Source/NSButtonCell.m
```diff
@@ -1099,7 +1099,7 @@ - (void) drawInteriorWithFrame: (NSRect)cellFrame inView: (NSView*)controlView
    * if highlighting is set (when a button is pushed it's
    * content is changed to the face of reversed state).
    */
```
iamleeg, line 1101: this comment needs updating :)

Source/NSGlyphGenerator.m
```diff
@@ -93,13 +93,15 @@ - (void) generateGlyphsForGlyphStorage: (id <NSGlyphStorage>)storage
                         characterIndex: (NSUInteger*)index
 {
   // Try to get enough space for all glyphs
-  NSGlyph glyphs[2 * num];
+  // TESTPLANT-MAL-09132018: Would be nice to optimize this...
+  //     but we've been seeing stack overwrite crashes due to this...
+  NSGlyph *glyphs = NSZoneMalloc(NSDefaultMallocZone(), 2 * num * sizeof(NSGlyph));
```
iamleeg, line 98: This is the same workaround I complained about up there ^, that if you're smashing the stack the solution is not to smash the heap instead.

Source/NSPasteboard.m
```diff
@@ -1939,154 +1939,182 @@ + (id) _lostServer: (NSNotification*)notification
...
+      //
+      // NOTE: This is a bit strange and bears review, I believe what TP is doing here is
+      //    using an external pasteboard server instead of gpbs.
+      //
+      // TESTPLANT-MAL-12112017: If the pasteboard server class it available
+      // the use it directly to avoid windows issue with the GPBS communications...
+      if (NSClassFromString(@"PasteboardServer") != nil)
+        {
+          the_server = [[NSClassFromString(@"PasteboardServer") alloc] init];
+        }
```
iamleeg, line 1951: This probably shouldn't go in? It looks like choosing to use their own class from a bundle. Perhaps they should swizzle `_lostServer:` in an injected lib.

Source/NSPopUpButton.m
```diff
@@ -232,6 +232,14 @@ - (void) removeItemAtIndex: (NSInteger)index
...
 
   [self synchronizeTitleAndSelectedItem];
+
+  // TESTPLANT-MAL-09182017: removing causes a multiple sequence eventually
+  //   deselecting any item - after multiple attempts to try to avoid the sequence
+  //   we'll just override the selection at this point...
+  if ([_cell usesItemFromMenu] && ([self selectedItem] == nil) && ([self numberOfItems] > 0))
+    {
+      [self selectItemAtIndex: [self numberOfItems]-1];
+    }
```
iamleeg, line 242: This is apparently working around a bug in `-synchronizeTitleAndSelectedItem`, as the [docs there](http://gnustep.org/resources/OpenStepSpec/ApplicationKit/Classes/NSPopUpButton.html) say "If there's no selected item, this method selects the first item in the item list and sets the receiver's title to match." but on line 239 the selected item is nil.

Source/NSSecureTextField.m
```diff
@@ -263,7 +278,7 @@ - (void) generateGlyphsForGlyphStorage: (id <NSGlyphStorage>)storage
                             glyphIndex: (NSUInteger*)glyph
                         characterIndex: (NSUInteger*)index
 {
-  NSGlyph glyphs[num];
+  NSGlyph *glyphs = NSZoneMalloc(NSDefaultMallocZone(), (sizeof(NSGlyph) * num));
```
iamleeg, line 281: Also this one here.

Source/GSLayoutManager.m
iamleeg: I looked at that other method today; it only reads the advances and doesn't write them. So there isn't a stack smasher after all, which is good :). I guess what the comment about "stack overwrite crashes" means is that if `length` is too big, then the array is too big to allocate on the stack. Then heap allocation is the right approach.

Source/NSCollectionView.m
```diff
@@ -442,6 +442,10 @@ - (NSRect) frameForItemAtIndex: (NSUInteger)theIndex
       
       x += _itemSize.width + _horizontalMargin;
     }
+  
+  if (_maxNumberOfColumns == 1)
+    itemFrame.size.width = self.frame.size.width;
```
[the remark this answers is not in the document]
fredkiefer: Why was this marked as resolved? If you don't like my input just remove me from the reviewer list.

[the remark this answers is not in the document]
gcasa: I had thought I commented that this was rejected.   Apparently that didn't post when I hit resolve.   Never ever think that I don't like your input.  I believe you interpret mistakes too harshly.

Source/NSCollectionView.m
```diff
@@ -474,6 +478,19 @@ - (NSCollectionViewItem *) itemAtIndex: (NSUInteger)index
         {
           [item setSelected: YES];
         }
+
+      // GJC: I realize this isn't being used, but it seems informational... we can
```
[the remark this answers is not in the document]
gcasa: This one was also marked resolved as it was rejected.   I will simply leave them in this state so that it is clear they were not accepted.

gcasa:
All relevant merges from TP have been made.  These changes are on hold.  Closing for now.

PR #115, opened 2021-10-13
add a preference to disable spellchecker

rmottola:
A different, lighter implementation of the option (same name) to disable the spell-checker.
I like it, it can be useful for specific application and environment and also because often standard install will die with "unable to contact the spell server".


rmottola:
> If this shows correctly here, the last commit messes up the closing `</desc>` tag.

O that was a strange mistake! Put it back!

PR #117, opened 2021-10-14
Add code to break out of the loop if char_last has exceeded the end of the existing set of characters

gcasa:


gcasa:
This PR will not be merged.  Closing

PR #123, opened 2021-11-07
16bit

fredkiefer:
This is a first attempt to sort out byte order for 16 bit images. I added all the new format constants that Apple has defined and use that in the 16 bit image code. As for now I handle the absence of these constants as host byte order, I am not sure whether this is correct.
The swap statements could be removed by cleverly merging the constant check with the byte order check, but I am to lazy to thing about that in the moment.
We still don't handle 32 bit images and images with floating point data.

I also fixed a few compiler warnings. This caused an issue with clang as the new integer types seem to come from different include files. Hopefully I got that correct now.

### DIFF
--- a/ColorPickers/GSNamedColorPicker.m
+++ b/ColorPickers/GSNamedColorPicker.m
@@ -27,7 +27,8 @@
 #import <Foundation/Foundation.h>
 #import <AppKit/AppKit.h>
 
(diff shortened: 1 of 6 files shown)


rmottola (approved):
Tested on Linux x86-64 and SPARC64

PR #124, opened 2021-12-17
Add helper methods in GSWindowDecorator protocol

nongio:
This PR extends the GSWindowDecorator protocol with new methods for checking if a point is inside the titlebar rectangle or resize bar rectangle.
This is useful for the wayland backend implementation where the backend needs to handle window move and resize.
As well this change would allow themes to customize more easily window behavior on mouse interactions on the titlebar or resize bar.

for more context check the cursor implementation:
https://github.com/nongio/libs-back/blob/wayland-pointer/Source/wayland/WaylandServer%2BCursor.m#L274

### DIFF
--- a/Source/GSStandardWindowDecorationView.m
+++ b/Source/GSStandardWindowDecorationView.m
@@ -381,9 +381,7 @@ calc_new_frame(NSRect frame, NSPoint point, NSPoint firstPoint,
 
 - (BOOL) pointInTitleBarRect:(NSPoint)point
 {
-  return !NSPointInRect(point, closeButtonRect) &&
-         !NSPointInRect(point, miniaturizeButtonRect) &&
-         NSPointInRect(point, titleBarRect);
+  return NSPointInRect(point, titleBarRect);
 }
 
 - (BOOL) pointInResizeBarRect:(NSPoint)point
--- a/Source/GSStandardWindowDecorationView.m
+++ b/Source/GSStandardWindowDecorationView.m
@@ -114,6 +114,10 @@
 				   [theme titlebarButtonSize], [theme titlebarButtonSize]);
       [closeButton setFrame: closeButtonRect];
     }
+  else
+    {
+        closeButtonRect = NSZeroRect;
+    }
 
   if (hasMiniaturizeButton)
     {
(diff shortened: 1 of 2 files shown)


Source/GSStandardWindowDecorationView.m
```diff
@@ -379,23 +374,58 @@ - (BOOL) acceptsFirstMouse: (NSEvent*)theEvent
...
 }
 
+- (BOOL) pointInContentRect:(NSPoint)point
+{
+    return NSPointInRect(point, contentRect);
+}
+
+- (BOOL) pointInTitleBarRect:(NSPoint)point
+{
+  return !NSPointInRect(point, closeButtonRect) &&
```
fredkiefer, line 384: Why are you excluding the buttons here?

nongio: the function should return YES when the cursor is inside the titlebar area that if clicked activates the window dragging. In the previous implementation this check is implicit given the view hierarchy, the close button is catching the event before the window decoration view

fredkiefer: I would prefer if we leave the buttons out here. That way we have the old code for non-Wayland. And even with Wayland it should not change the behaviour as you already pointed out as the view hierarchy will take care of the correct delivery of the mouse event.

nongio: sounds good, i'll update it

nongio: it is now updated

fredkiefer (approved):

rmottola:
This broke compilation for me:
```GSStandardWindowDecorationView.m: In function ‘-[GSStandardWindowDecorationView resizeWindowStartingWithEvent:]’:
GSStandardWindowDecorationView.m:329:3: warning: ‘GSStandardWindowDecorationVie’ may not respond to ‘-resizeModeForPoint:’
   GSResizeEdgeMode mode = [self resizeModeForPoint: firstPoint];
   ^~~~~~~~~~~~~~~~
GSStandardWindowDecorationView.m:329:3: warning: (Messages without a matching method signature
GSStandardWindowDecorationView.m:329:3: warning: will be assumed to return ‘id’ and accept
GSStandardWindowDecorationView.m:329:3: warning: ‘...’ as arguments.)
GSStandardWindowDecorationView.m:329:27: error: incompatible types when initializing type ‘GSResizeEdgeMode’ {aka ‘enum _GSResizeEdgeMode’} using type ‘id’
   GSResizeEdgeMode mode = [self resizeModeForPoint: firstPoint];
          
```

PR #125, opened 2021-12-23
* check the drag source' delegate on NSDraggingSource

svgol:
It eliminates requirement for subclassing of a drag source. One can only implement NSDraggingSource in the drag source' delegate like it has already been  done for NSDraggingDestination in NSWindow (see macros GSPerformDragSelector, GSPerformVoidDragSelector).


svgol:
I made the pull request prematurely

Issue #129, opened 2022-01-10
GSScaleFactor setting and titlebar window decorations


anthonyc-r:
I've noticed that the titlebar height is defined as a constant in GSThemeDrawing.m. This results in the titlebar buttons being clipped when scaling is set to, for example, 1.5. It would be possible to modify [GSTheme titlebarHeight] to return the scaled height, either calculated every time, or stored as a instance variable which is set up on init. Would one of these solutions be acceptable?
Cheers.

anthonyc-r:
Actually, it seems to make sense to incorporate [NSScreen userSpaceContentScale] into the calculations in NSStandardWindowDecoration.m rather than changing the theme constants.

PR #130, opened 2022-01-11
GSStandardWindowDecorationView: Apply userSpaceScaleFactor to calcula…

anthonyc-r:
…tions involving the title bar height

This resolves this issue #129 that I noticed when using client side decorations. I've tested this with a number of apps TalkSoup, GWorkspace, Terminal. With the default theme and also with Rik.theme on arch linux, kernel 5.15.12-arch1-1, x86_64.

### DIFF
--- a/Source/GSStandardWindowDecorationView.m
+++ b/Source/GSStandardWindowDecorationView.m
@@ -67,7 +67,7 @@
   if (style
     & (NSTitledWindowMask | NSClosableWindowMask | NSMiniaturizableWindowMask))
     {
-      *t = [theme titlebarHeight];
+      *t = [[NSScreen mainScreen] userSpaceScaleFactor] * [theme titlebarHeight];
     }
   if (style & NSResizableWindowMask)
     {


Source/GSStandardWindowDecorationView.m
```diff
@@ -97,7 +97,7 @@ - (void) updateRects
 
   if (hasTitleBar)
     {
-      CGFloat titleHeight = [theme titlebarHeight];
+      CGFloat titleHeight = [[NSScreen mainScreen] userSpaceScaleFactor] * [theme titlebarHeight];
```
fredkiefer, line 100: Are you sure that this second usage is also needed? According to my view we would only need the manual scaling for coordinates in window coordinate systems. The values here are in the bounds coordinate system and should be automatically scaled. If this does not happen, I will have to rethink a few things.

anthonyc-r: Ah, I’m not actually - I’ll remove this and give it another test.

anthonyc-r: It does work fine with just the first change, I've amended my commit - thanks for the help.

fredkiefer (approved):

PR #139, opened 2022-02-11
Fix minor memory leaks

williameveretteggplant:
A leak of NSColor objects in NSBox, and GSToolTip objects in NSView

### DIFF
--- a/Source/NSView.m
+++ b/Source/NSView.m
@@ -750,10 +750,9 @@ GSSetDragTypes(NSView* obj, NSArray *types)
   TEST_RELEASE(_frameMatrix);
   TEST_RELEASE(_boundsMatrix);
   TEST_RELEASE(_sub_views);
-  if (_rFlags.has_tooltips != 0)
-    {
-      [GSToolTips removeTipsForView: self];
-    }
+
+  [GSToolTips removeTipsForView: self];
+
   if (_rFlags.has_currects != 0)
     {
       [self discardCursorRects];	// Handle release of cursors
--- a/Source/NSView.m
+++ b/Source/NSView.m
@@ -752,9 +752,9 @@ GSSetDragTypes(NSView* obj, NSArray *types)
   TEST_RELEASE(_sub_views);
 
   if (_rFlags.has_tooltips != 0)
-  {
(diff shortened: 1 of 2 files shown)


fredkiefer:
I fully understand the NSBox change. For the tool tips change in NSView I am sure that it won't do any harm but at the moment I don't see why this change would be needed? A view start off without tool tips and when we add one we also set the flag. We even don't clear the flag when we remove the tool tips. That way it looks to me as the flag should always be set when there is a tool tip to remove. Why do you think this isn't the case?

williameveretteggplant:
There appears to be a line number mixup. From what I can see that line number is part of the writePDFInsideRect:toPasteboard: method and doesn't have anything to do with Tool Tips. Could you clarify what method you are talking about?

fredkiefer:
No, it is worse. I did not mix up the line number but the file name. I wanted to refer to NSTextView.m not NSView.m. Sorry for the confusion.

Source/NSView.m
```diff
@@ -750,10 +750,9 @@ - (void) dealloc
   TEST_RELEASE(_frameMatrix);
   TEST_RELEASE(_boundsMatrix);
   TEST_RELEASE(_sub_views);
-  if (_rFlags.has_tooltips != 0)
-    {
-      [GSToolTips removeTipsForView: self];
-    }
+
+  [GSToolTips removeTipsForView: self];
```
fredkiefer, line 754: Did you test, whether you still need this change? It won't do any harm, maybe slow things a bit down. But what we really should know is whether we addressed the underlying issue.

Source/NSView.m
```diff
@@ -750,10 +750,11 @@ - (void) dealloc
   TEST_RELEASE(_frameMatrix);
   TEST_RELEASE(_boundsMatrix);
   TEST_RELEASE(_sub_views);
+
   if (_rFlags.has_tooltips != 0)
-    {
-      [GSToolTips removeTipsForView: self];
-    }
+  {
```
fredkiefer, line 755: Looks like the indentation got messed up. Could you please fix this?

fredkiefer (approved):

PR #143, opened 2022-03-18
only for ImageCell set refuses first responder, as behavior on Mac.

rmottola:
With this change I made GS match Mac behavior.
It makes my test case work and match Mac on both GORM instantiated NSImageView and its NSImageCell, as code-instantiated NSImageCell, NSActionCell, NSCell.

However PRICE still doesn't respond properly when copy&paste into the image-view and that still puzzles me.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2022-03-18  Riccardo Mottola <rm@gnu.org>
+
+	* Source/NSCell.m: only for ImageCell set refuses first responder,
+	as behavior on Mac.
+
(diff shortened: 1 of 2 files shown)


rmottola:
you are right, I made more precise tests to check for this differentiated behavior, I updated the Issue with the new findings.

rmottola:
@fredkiefer however NSImageCell has no initImageCell, this is why I changed NSCell. Should I make an explicit subclass to override this value? How, since in init actually [self initImageCell] is used... give me a hint.

PR #144, opened 2022-03-20
Create files to support NSMovie and NSMovieView

gcasa:
implementation for NSMovie and NSMovieView.   The tests for this are here:
https://github.com/gcasa/NSMovie_test
https://github.com/gcasa/apps-videoplayer

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,23 +1,26 @@
 2025-05-29 Gregory John Casamento <greg.casamento@gmail.com>
 
-	* config.make.in
+	* config.make.in: Add HAVE_* variables.
 	* configure
-	* configure.ac
+	* configure.ac: Look for relevant libraries
 	* Headers/Additions/GNUstepGUI/config.h.in
 	* Headers/AppKit/NSMovie.h
 	* Headers/AppKit/NSMovieView.h
 	* Source/GNUmakefile
 	* Source/GNUmakefile.preamble
+	* Source/GSAudioPlayer.h
+	* Source/GSAudioPlayer.m: Added to play audio from avcodec
 	* Source/GSMovieView.h
-	* Source/GSMovieView.m
+	* Source/GSMovieView.m: Subclass to handle video playback
 	* Source/NSMovie.m
-	* Source/NSMovieView.m
-	* Source/NSSound.m: Cleanup NSSound, add support for playing ffmpeg
-	movies using NSMovie, NSMovieView.
+	* Source/NSMovieView.m: Cleanup and add textfield.
+	* Source/NSSound.m: Cleanup NSSound, add support for playing
+	ffmpeg movies using NSMovie, NSMovieView.
 
 2025-04-30 Gregory John Casamento <greg.casamento@gmail.com>
 
-	* Headers/AppKit/NSColorSpace.h: Add NSColorSpaceModel* enumerated values.
+	* Headers/AppKit/NSColorSpace.h: Add NSColorSpaceModel*
+	enumerated values.
 
 2025-02-11 Richard Frith-Macdonald <rfm@gnu.org>
 
@@ -26,8 +29,7 @@
 	* INSTALL:
 	* NEWS:
 	* Source/DocMakefile:
-	* Version:
-	Updates for 0.32.0 release
+	* Version: Updates for 0.32.0 release
 
 2024-11-08 Gregory John Casamento <greg.casamento@gmail.com>
 
@@ -44,7 +46,8 @@
 
 2024-10-29 Fred Kiefer <FredKiefer@gmx.de>
 
-	* Headers/AppKit/NSLayoutAnchor.h: Add missing include that is required after a change in base.
+	* Headers/AppKit/NSLayoutAnchor.h: Add missing include that is
+	required after a change in base.
 	* Source/NSTableView.m: Forward declare new methods.
 
 2024-08-29 Gregory John Casamento <greg.casamento@gmail.com>
@@ -62,7 +65,7 @@
 	for NSTreeController, a subclass of NSTreeNode.
 	* Source/GSThemeDrawing.m: Optimize code a bit and
 	add changes for bindings.
-	* Source/GSXib5KeyedUnarchiver.m: Add support for new keys
+	* Source/GSib5KeyedUnarchiver.m: Add support for new keys
 	on NSTreeController.
 	* Source/NSKeyValueBinding.m: Add new methods to
 	GSKeyValueBinding
--- a/ChangeLog
+++ b/ChangeLog
@@ -65,7 +65,7 @@
 	for NSTreeController, a subclass of NSTreeNode.
 	* Source/GSThemeDrawing.m: Optimize code a bit and
 	add changes for bindings.
-	* Source/GSib5KeyedUnarchiver.m: Add support for new keys
+	* Source/GSXib5KeyedUnarchiver.m: Add support for new keys
 	on NSTreeController.
 	* Source/NSKeyValueBinding.m: Add new methods to
 	GSKeyValueBinding
--- a/Headers/AppKit/NSMovie.h
+++ b/Headers/AppKit/NSMovie.h
@@ -1,4 +1,4 @@
-/** <title>NSMovie</title>
+/* <title>NSMovie</title>
 
    <abstract>Encapsulate a Quicktime movie</abstract>
 
@@ -30,9 +30,25 @@
 
 #ifndef _GNUstep_H_NSMovie
 #define _GNUstep_H_NSMovie
-#import <AppKit/AppKitDefines.h>
-
 #import <Foundation/NSObject.h>
+#import "AppKit/AppKitDefines.h"
+
+/**
+ * NSMovie
+ *
+ * Encapsulates a QuickTime movie object, providing methods to initialize and
+ * retrieve movie data from various sources, such as NSData, NSURL, or NSPasteboard.
+ *
+ * Instance Variables:
+ *   _movie:
+ *     The raw NSData representing the contents of the QuickTime movie.
+ *
+ *   _url:
+ *     The URL from which the movie was loaded. May be nil if not URL-based.
+ *
+ *   _tmp:
+ *     A boolean indicating whether the movie data is temporary (not persisted).
+ */
 
 @class NSArray;
 @class NSData;
--- a/Headers/AppKit/NSMovieView.h
+++ b/Headers/AppKit/NSMovieView.h
@@ -5,7 +5,7 @@
    Copyright <copy>(C) 2003 Free Software Foundation, Inc.</copy>
 
    Author: Gregory John Casamento <greg.casamento@gmail.com>
-   Date: May 2025
+   Date: March 2003
    Author: Fred Kiefer <FredKiefer@gmx.de>
    Date: March 2003
 
--- a/Headers/AppKit/NSMovieView.h
+++ b/Headers/AppKit/NSMovieView.h
@@ -1,4 +1,4 @@
-/** <title>NSMovieView</title>
+/* <title>NSMovieView</title>
 
    <abstract>Encapsulate a view for Quicktime movies</abstract>
 
@@ -34,6 +34,37 @@
 #import <AppKit/AppKitDefines.h>
 #import <AppKit/NSNibDeclarations.h>
 #import <AppKit/NSView.h>
+
+/**
+ * NSMovieView
+ *
+ * Encapsulates a view component for playing QuickTime movies using the NSMovie class.
+ * Provides playback controls, status monitoring, and display customization.
+ *
+ * Instance Variables:
+ *   _statusField:
+ *     An optional NSTextField outlet that displays the playback status of the movie.
+ *
+ *   _movie:
+ *     The NSMovie object currently loaded in the view.
+ *
+ *   _rate:
+ *     The current playback rate of the movie.
+ *
+ *   _volume:
+ *     The current playback volume level.
+ *
+ *   _flags:
+ *     A bitfield structure representing various movie playback options and UI states:
+ *       - playing: Whether the movie is currently playing.
+ *       - muted: Whether playback is muted.
+ *       - loopMode: Loop playback mode (normal, looping, back-and-forth).
+ *       - plays_selection_only: Whether only the selected portion of the movie is played.
+ *       - plays_every_frame: Whether all frames are rendered during playback.
+ *       - is_controller_visible: Whether the movie controller is currently visible.
+ *       - editable: Whether the movie view is editable.
+ *       - reserved: Reserved bits for future use.
+ */
 
 @class NSTextField;
 @class NSMovie;
--- a/Headers/AppKit/NSMovieView.h
+++ b/Headers/AppKit/NSMovieView.h
@@ -79,6 +79,8 @@ APPKIT_EXPORT_CLASS
 @interface NSMovieView : NSView
 {
   IBOutlet NSTextField *_statusField;
+  IBOutlet id _positionField;
+  
   NSMovie* _movie;
   
   float _rate;
@@ -289,7 +291,7 @@ APPKIT_EXPORT_CLASS
 /**
  * GNUstep extension, get current position in the movie.
  */
-- (CGFloat) currentPosition;
+- (double) currentPosition;
 
 @end
 
--- a/Headers/AppKit/NSMovieView.h
+++ b/Headers/AppKit/NSMovieView.h
@@ -52,8 +52,8 @@ APPKIT_EXPORT_CLASS
   
   float _rate;
   float _volume;
-  BOOL _playing;
   struct NSMovieViewFlags {
+    unsigned int playing: 1;
     unsigned int muted: 1;
     unsigned int loopMode: 3;
     unsigned int plays_selection_only: 1;
--- a/Headers/AppKit/NSMovieView.h
+++ b/Headers/AppKit/NSMovieView.h
@@ -247,10 +247,13 @@ APPKIT_EXPORT_CLASS
 /**
  * GNUstep extension, this allows the user to set a
  * text field to monitor the status of the movie view.
- *
+ */
 - (void) setStatusField: (id)field;
+
+/**
+ * GNUstep extension, get the status field.
+ */
 - (id) statusField;
-*/
 
 @end
 
--- a/Headers/AppKit/NSMovieView.h
+++ b/Headers/AppKit/NSMovieView.h
@@ -286,6 +286,11 @@ APPKIT_EXPORT_CLASS
  */
 - (id) statusField;
 
+/**
+ * GNUstep extension, get current position in the movie.
+ */
+- (CGFloat) currentPosition;
+
 @end
 
 #endif /* _GNUstep_H_NSMovieView */
--- a/Source/GSAVUtils.m
+++ b/Source/GSAVUtils.m
@@ -54,12 +54,13 @@ AVPacket AVPacketFromNSDictionary(NSDictionary *dict)
   NSNumber *duration = [dict objectForKey: @"duration"];
   NSNumber *flags = [dict objectForKey: @"flags"];
 
-  AVPacket packet;
-  packet.data = (uint8_t *)[data bytes];
-  packet.size = (int)[data length];
-  packet.pts = [pts longLongValue];
-  packet.duration = [duration intValue];
-  packet.flags = [flags intValue];
+  AVPacket *packet = av_packet_alloc();
 
-  return packet;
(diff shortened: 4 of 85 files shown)


ChangeLog
```diff
@@ -29,6 +49,7 @@
         from here.
 	* Source/GSXibKeyedUnarchiver.m: Implement logic here to build
         _customClasses array when loading a XIB of version 3 or 4.
+>>>>>>> e0012349f6df6895f736de1f158c52116fce7fc1
```
fredkiefer, line 52: This line seems to be a leftover from a failed merge.

Source/NSMovie.m
```diff
@@ -54,24 +128,51 @@ + (BOOL) canInitWithPasteboard: (NSPasteboard*)pasteboard
...
+                {
+                  _source = [[pluginClass alloc] initWithData: movieData];
+                }
+            }
+          END_FOR_IN(__videoSourcePlugIns);
+        }
+      else
+        {
+          RELEASE(self);
+        }
```
fredkiefer, line 163: you need to return `nil` here and not `self`.

gcasa: Obsolete.

ChangeLog
gcasa: Obsolete.

gcasa:
For some reason the sound is a little choppy, I am trying to address this.  In the meantime I am going to work on implementing the other features of this class.

gcasa:
Sound seems to be working now, apparently the sample rate was misconfigured.   I am working on implementing the video looping settings as well as other settings.  The methods for editing are going to be left unimplemented as I am not certain how to do those yet.

gcasa:
I am making this ready for review so that, perhaps, people who are more familiar with avcodec et. al. might be able to help me improve this.

gcasa:
I think this is the best I can do on this.  @fredkiefer could you take a look and see what else you might be able to see.  This is certainly better than we have at the moment (nothing) and I think it would be useful for others to take a look and try to improve this.

gcasa:
screen_grab

The sound is now working and I have added a set of (default) controls.

Headers/AppKit/NSMovieView.h
```diff
@@ -43,66 +47,213 @@ typedef enum {
...
-      unsigned int is_controller_visible: 1;
-      unsigned int editable: 1;
-      unsigned int reserved: 24;
-    } _flags;
+  IBOutlet NSTextField *_statusField;
+  NSMovie* _movie;
+  
+  float _rate;
+  float _volume;
+  BOOL _playing;
```
fredkiefer: Have you considered putting this value into one of the reserved bit on the following flag array?

Source/GSAVUtils.m
```diff
@@ -0,0 +1,65 @@
...
+
+#import <Foundation/NSData.h>
+#import <Foundation/NSDictionary.h>
+#import <Foundation/NSValue.h>
+
+#import "GSAVUtils.h"
+
+NSDictionary *NSDictionaryFromAVPacket(AVPacket *packet)
+{
+  NSData *data = [NSData dataWithBytes:packet->data length:packet->size];
```
fredkiefer: Could you please leave a space after the colon? I think this is more readable and still the more common GNUstep coding style.

Source/GSAudioPlayer.m
```diff
@@ -0,0 +1,422 @@
...
+    }
+
+  if (_aoDev)
+    {
+      ao_close(_aoDev);
+    }
+
+  RELEASE(_audioPackets);
+
+  ao_shutdown();
```
fredkiefer, line 89: This only works if this class is used as a singleton. I am not sure whether this is the case or not.

Source/GSAudioPlayer.m
```diff
@@ -0,0 +1,422 @@
...
+		      AV_SAMPLE_FMT_S16,
+		      _audioCodecCtx->sample_rate,
+		      &_audioCodecCtx->ch_layout,
+		      _audioCodecCtx->sample_fmt,
+		      _audioCodecCtx->sample_rate,
+		      0, NULL);
+  int r = swr_init(_swrCtx);
+  if (r == 0)
+    {
+      NSLog(@"[GSAudioPlayer] WARNING: swr_init returned 0");
```
fredkiefer: Why is there now `return` here?

Source/GSAudioPlayer.m
```diff
@@ -0,0 +1,422 @@
...
+  AVCodecParameters *audioPar = formatCtx->streams[audioStreamIndex]->codecpar;
+  const AVCodec *audioCodec = avcodec_find_decoder(audioPar->codec_id);
+
+  if (!audioCodec)
+    {
+      NSLog(@"Audio codec not found.");
+      return;
+    }
+
+  _audioCodecCtx = avcodec_alloc_context3(audioCodec);
```
fredkiefer, line 166: if this method gets called a second time we will leak memory.

Source/GSAudioPlayer.m
```diff
@@ -0,0 +1,422 @@
...
+
+  // Set buffer size to reduce skipping (in bytes)
+  // Calculate buffer for about 100ms of audio
+  int bufferSamples = _audioCodecCtx->sample_rate / 10; // 100ms
+  int bufferSize = bufferSamples * out_channels * 2; // 16-bit stereo
+  char bufferSizeStr[32];
+  snprintf(bufferSizeStr, sizeof(bufferSizeStr), "%d", bufferSize);
+  ao_append_option(&options, "buffer_time", "100000"); // 100ms in microseconds
+
+  NSLog(@"[GSAudioPlayer] Initializing audio: %d Hz, %d channels, buffer size: %d bytes",
```
fredkiefer: This probably should be a debug log.

Source/GSAudioPlayer.m
```diff
@@ -0,0 +1,422 @@
...
+  int64_t totalSamplesPlayed = 0;
+
+  while (_running)
+    {
+      // create pool...
+      CREATE_AUTORELEASE_POOL(pool);
+      {
+	NSDictionary *dict = nil;
+
+	@synchronized (_audioPackets)
```
fredkiefer, line 257: I think this is only supported by clang. It would be nice to get this code running with gcc is well. Otherwise we need to protect this code in the Makefile so it won't be used when compiling with gcc.

Source/GSAudioPlayer.m
```diff
@@ -0,0 +1,422 @@
...
+		[_audioPackets removeObjectAtIndex:0];
+	      }
+	  }
+
+	if (!_started && dict)
+	  {
+	    audioStartTime = av_gettime();
+	    totalSamplesPlayed = 0;
+	    _started = YES;
+	    NSLog(@"[GSAudioPlayer] Audio playback started | Timestamp: %ld", audioStartTime);
```
fredkiefer: Again this should be a debug log.

Source/GSAudioPlayer.m
```diff
@@ -0,0 +1,422 @@
...
+	      }
+
+	    // Update audio clock for video synchronization
+	    // The audio clock represents the actual time of the audio currently being played
+	    _audioClock = audioStartTime + (totalSamplesPlayed * 1000000LL / _audioCodecCtx->sample_rate);
+
+	    // Debug logging for audio clock updates (reduced frequency)
+	    if (totalSamplesPlayed % (_audioCodecCtx->sample_rate / 4) == 0) // Log 4 times per second
+	      {
+		fprintf(stderr, "[GSAudioPlayer] Audio clock: %ld | PTS: %ld | Samples: %ld | Timing error: %ld us\n",
```
fredkiefer: Why `fprintf` and not NSLog, or rather NSDebugLog?

Source/GSAudioPlayer.m
```diff
@@ -0,0 +1,422 @@
...
+    {
+      volume = 0.0;
+    }
+
+  if (volume > 1.0)
+    {
+      volume = 1.0;
+    }
+
+  _volume = volume;
```
fredkiefer, line 601: Why doesn't this method set _muted?

Source/GSAudioPlayer.m
```diff
@@ -0,0 +1,422 @@
...
+    {
+      NSLog(@"Skipping corrupt audio packet");
+      return 0;
+    }
+
+  while (avcodec_receive_frame(_audioCodecCtx, _audioFrame) == 0)
+    {
+      int outSamples = _audioFrame->nb_samples;
+      int outBytes = av_samples_get_buffer_size(NULL, 2, outSamples, AV_SAMPLE_FMT_S16, 1);
+      uint8_t *outBuf = (uint8_t *) malloc(outBytes);
```
fredkiefer: We should think about keeping a buffer for a longer time. Most samples will require a buffer with the same size and we could reuse that.

Source/GSMovieView.h
```diff
@@ -0,0 +1,120 @@
...
+  GSAudioPlayer *_audioPlayer;
+
+  AVCodecContext *_videoCodecCtx;
+  AVFrame *_videoFrame;
+  AVFormatContext *_formatCtx;
+  AVStream *_stream;
+  struct SwsContext *_swsCtx;
+  AVRational _timeBase;
+
+  BOOL _running; // is the loop currently running...
```
fredkiefer, line 74: Doesn't this duplicate the _playing variable from the super class?

Source/NSMovie.m
```diff
@@ -54,55 +74,61 @@ + (BOOL) canInitWithPasteboard: (NSPasteboard*)pasteboard
...
+  self = [super init];
+  if (self != nil)
+    {
+      NSString *filepath = _writeDataToTempFile(_movie);
 
-- (id) initWithMovie: (void*)movie
-{
-  //FIXME
+      _url = [NSURL fileURLWithPath: filepath];
+      _tmp = YES;
```
fredkiefer, line 91: It seems like this variable is only ever set but never read.

fredkiefer (changes requested):
I only got halfway through in reviewing this code but I think this already gives you enough to further work on your code.

Headers/AppKit/NSMovieView.h
gcasa: Hey, Fred!  I hope all is well.  I hadn't considered that, but it is certainly a good idea.

Source/GSAVUtils.m
gcasa: yes, absolutely.

Source/GSAudioPlayer.m
gcasa: The documentation indicates that this is not a critical issue.   There are instances where this works just fine.

Source/GSAudioPlayer.m
gcasa: I believe this is okay as ao_shutdown() is process isolated.   So since I am calling ao_initialize() earlier in the process it should be okay.

Source/GSAudioPlayer.m
gcasa: This is printing out every frame, so it's probably a bit more than is needed.  I likely need to turn this logging into a DebugLog instead of doing it like this.  It was helping me to monitor frames and synchronization.

Source/GSAudioPlayer.m
gcasa: Done.

Source/NSMovie.m
gcasa: I now have it deleting the temporary file if it exists.

Source/GSMovieView.h
gcasa: Fixed.

Source/GSAudioPlayer.m
gcasa: Added an if to prevent this.

Source/GSAudioPlayer.m
gcasa: I'm using NSDebugLog.  I have less debugging to do on this at the moment, and the frame numbers are also output on the statusField.

Source/GSAudioPlayer.m
gcasa: Agreed.

gcasa: Done.

gcasa:
@fredkiefer Take a look at the updates.  Thanks for reviewing the code.

ChangeLog
```diff
@@ -45,7 +65,7 @@
 	for NSTreeController, a subclass of NSTreeNode.
 	* Source/GSThemeDrawing.m: Optimize code a bit and
 	add changes for bindings.
-	* Source/GSXib5KeyedUnarchiver.m: Add support for new keys
+	* Source/GSib5KeyedUnarchiver.m: Add support for new keys
```
fredkiefer: Something went wrong here.

Source/GSAudioPlayer.h
```diff
@@ -0,0 +1,139 @@
...
+  struct GSAudioPlayerFlags {
+    unsigned int playing: 1;
+    unsigned int started: 1;
+    unsigned int muted: 1;
+    unsigned int needsRestart: 1;
+    unsigned int reserved: 28;
+  } _flags;
+
+  // New variables for proper pause/resume handling
+  BOOL _reachedEOF;
```
fredkiefer: Why not as part of the bit field?
And "new" doesn't make much sense in the comment, as this will be the first version of the file for all others.

Source/GSAudioPlayer.m
fredkiefer: But will it work with a process that produces two sounds at once?

Source/GSAudioPlayer.m
```diff
@@ -0,0 +1,732 @@
...
+      NSDebugLog(@"Audio codec already initialized");
+      return;
+    }
+
+  _audioCodecCtx = avcodec_alloc_context3(audioCodec);
+  avcodec_parameters_to_context(_audioCodecCtx, audioPar);
+
+  if (avcodec_open2(_audioCodecCtx, audioCodec, NULL) < 0)
+    {
+      NSDebugLog(@"Failed to open audio codec.");
```
fredkiefer, line 176: Should we release _audioCodecCtx here?

Source/GSAudioPlayer.m
```diff
@@ -0,0 +1,732 @@
...
+      return;
+    }
+
+  _audioFrame = av_frame_alloc();
```
fredkiefer, line 185: All these calls to allocate may return NULL, we should try to handle that.

Source/GSAudioPlayer.m
```diff
@@ -0,0 +1,732 @@
...
+
+  // Configure audio device options for better buffering to reduce skipping
+  ao_option *options = NULL;
+
+  // Set buffer size to reduce skipping (in bytes)
+  // Calculate buffer for about 100ms of audio
+  int bufferSamples = _audioCodecCtx->sample_rate / 10; // 100ms
+  int bufferSize = bufferSamples * out_channels * 2; // 16-bit stereo
+  char bufferSizeStr[32];
+  snprintf(bufferSizeStr, sizeof(bufferSizeStr), "%d", bufferSize);
```
fredkiefer: To me it looks like bufferSizeStr is never used.

Source/GSAudioPlayer.m
```diff
@@ -0,0 +1,732 @@
...
+      // create pool...
+      CREATE_AUTORELEASE_POOL(pool);
+      {
+	NSDictionary *dict = nil;
+
+	@synchronized (_audioPackets)
+	  {
+	    if ([_audioPackets count] > 0)
+	      {
+		dict = RETAIN([_audioPackets objectAtIndex: 0]);
```
fredkiefer, line 261: Something seems wrong with the indentation.

Source/NSMovie.m
```diff
@@ -21,88 +23,112 @@
...
   return self;
 }
 
-- (id) initWithURL: (NSURL*)url byReference: (BOOL)byRef
+- (instancetype) initWithMovie: (void *)movie
 {
-  NSData* data = [url resourceDataUsingCache: YES];
-
-  self = [self initWithData: data];
+  return [self initWithData: movie];
```
fredkiefer, line 100: This seems strange. Shouldn't you  package up the void pointer in an NSData object before passing it on?

ChangeLog
gcasa: Not sure how that happened.  Corrected.

Source/GSAudioPlayer.h
gcasa: Fixed.

Source/GSAudioPlayer.m
gcasa: Done

Source/GSAudioPlayer.m
gcasa: Added checks.

Source/GSAudioPlayer.m
gcasa: Removed

Source/GSAudioPlayer.m
gcasa: Done.

Source/NSMovie.m
gcasa: I wasn't sure how to handle this one.  Given that we really don't know what the pointer being passed in is.  I felt as though it was safer to just treat it like an NSData.

gcasa:
I think this is as close to done as I can get it.  It works.  I would like others to have the opportunity to look at and fix the code on this.  Please review and approve if you can @rfm or @fredkiefer

gcasa:

gcasa:
@fredkiefer can you take one final look at this.  I really have taken this as far as I can.  It's not perfect, but it works.  I would like to merge this so that others can inspect and improve the code.

gcasa:
Unless there is an objection I am going to merge this.  I have taken it as far as I can.  I want more people to be able to test it and improve it.

fredkiefer:
Just give me a few more hours. I am planing of doing a final review of this code tonight.

Headers/AppKit/NSMovie.h
```diff
@@ -1,9 +1,11 @@
-/** <title>NSMovie</title>
+/* <title>NSMovie</title>
```
fredkiefer, line 1: Why did you remove this star here? Is this something we should change in other files as well?

gcasa:
Okay. I didn't mean to rush you. I just wanted others to be able to take advantage of it.

Source/NSMovie.m
fredkiefer: I think you are wrong here. The pointer will most likely point to raw movie data, in Apples case a QTMovie. I am unsure how we should handle that. Maybe leave it unimplemented as it was?

Source/GSMovieView.m
```diff
@@ -0,0 +1,1338 @@
...
+@end
+
+@implementation NSMovieView (AVCodec)
+
+// Make sure we have all of the video and audio components, to build this...
++ (id) allocWithZone: (NSZone *)zone
+{
+  if (self == [NSMovieView class])
+    {
+      return [GSMovieView allocWithZone: zone];
```
fredkiefer, line 118: I would prefer an implementation  where we don't subclass NSMovieView. That way we could have different decoder alongside each other.

Source/GSMovieView.m
```diff
@@ -0,0 +1,1338 @@
...
+  sws_scale(_swsCtx,
+	    (const uint8_t * const *)videoFrame->data,
+	    videoFrame->linesize,
+	    0,
+	    height,
+	    rgbData,
+	    rgbLineSize);
+
+  // Create bitmap with NULL planes so it allocates its own memory
+  NSBitmapImageRep *bitmap = [[NSBitmapImageRep alloc]
```
fredkiefer, line 1085: Here we should try to avoid generating a new NSImage and NSBitmapImageRep for each frame. Or at least avoid the `memcpy`by providing `rgbData` as the first parameter to this method call.

fredkiefer (commented):
I don't quite like the concepts behind the current implementation, but I don't see much harm in it. If you really want to, you may go ahead and merge this.

Source/NSMovie.m
gcasa: Actually, I think you're right on this one.

Headers/AppKit/NSMovie.h
gcasa: No, this was a mistake.

Source/GSMovieView.m
gcasa: Well, the way it is implemented it uses avcodec which supports a ton of decoders.   I have considered moving support for the movie playing either to the backend or into a separate bundle, but I am not sure.

gcasa:
Merging this with fred's okay.  Everyone please test this and let me know what I have done wrong with decoding, if anything or how I can improve my approach.

PR #145, opened 2022-03-23
Cell responder

rmottola:
Second attempt.
This makes test pass and the same test are proven to fail, do you like it @fredkiefer ?


rmottola:
@gcasa all GORM files created int he past have this property wrongly set. At a first test, creating a new GORM makes a correct file. This property is not currently inspectable in GORM, so hard to fix current file. What do you think of creating a new coding version and fixing that on load?

fredkiefer (approved):

PR #146, opened 2022-03-24
Another windows fix, per Greg C recommendation.

DavidHVernon:


fredkiefer:
I see the change you made but what is the issue you are trying to resolve?

gcasa:
> I see the change you made but what is the issue you are trying to resolve?

I believe the issue is that in some cases the block isn't defined but the compiler still tries to build the code.  So it is getting an error.  Moving the block declaration outside means it is always defined.

fredkiefer:
I would suggest we first find out what the actual root cause of the problem is and try to address that. And only if wa fail move the code around. This change won't do any harm. But understanding the real issue is more important.

fredkiefer (approved):

PR #147, opened 2022-03-26
Update GSXIBKeyedUnarchiver to build customClasses array when loading in Gorm

gcasa:
This change allows XIB 3/4 files to load properly in Gorm.  Moved the "customClass" method and ivar to GSXIBKeyedUnarchiver instead of in its subclass.

### DIFF
--- a/Source/GSXibKeyedUnarchiver.m
+++ b/Source/GSXibKeyedUnarchiver.m
@@ -198,9 +198,10 @@
                                         {
                                           NSXMLElement *refNode = [refNodes objectAtIndex: 0];
                                           NSString *refId = [refNode stringValue];
-                                          NSDictionary *dict = [NSDictionary dictionaryWithObjectsAndKeys:
-                                                                               refId, @"id", sc, @"parentClassName", cn, @"customClassName", nil];
-                                          [_customClasses addObject: dict];
+
+                                          [self createCustomClassRecordForId: refId
+                                                             withParentClass: sc
+                                                              forCustomClass: cn];
                                         }
                                     }
                                 }
@@ -1109,6 +1110,20 @@ didStartElement: (NSString*)elementName
 - (NSArray *) customClasses
 {
   return _customClasses;
+}
+
+- (void) createCustomClassRecordForId: (NSString *)theId
+                      withParentClass: (NSString *)parentClassName
+                       forCustomClass: (NSString *)customClassName
+{
+  if (theId == nil || customClassName == nil)
+    return;
+
+  NSMutableDictionary *dict = [NSMutableDictionary dictionaryWithObjectsAndKeys:
+                                                     theId, @"id",
+                                                   parentClassName, @"parentClassName",
+                                                   customClassName, @"customClassName",nil];
+  [_customClasses addObject: dict];
 }
 
 @end
--- a/Source/GSXibKeyedUnarchiver.m
+++ b/Source/GSXibKeyedUnarchiver.m
@@ -86,22 +86,13 @@
 - (NSString *) _substituteClassForClassName: (NSString *)className
 {
   NSString *result = className;
-  NSEnumerator *en = [_customClasses objectEnumerator];
-  NSDictionary *dict = nil;
+  NSDictionary *dict = [_customClasses objectForKey: className];  
 
-  //  NSLog(@"_customClasses = %@", _customClasses);
-  while ((dict = [en nextObject]) != nil)
+  if (dict != nil)
     {
-      NSString *customClassName = [dict objectForKey: @"customClassName"];
-
-      if ([customClassName isEqualToString: className])
-        {
-          result = [dict objectForKey: @"parentClassName"];
-          break;
-        }
+      result = [dict objectForKey: @"parentClassName"];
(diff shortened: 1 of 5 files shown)


Source/GSXibKeyedUnarchiver.m
```diff
@@ -264,6 +326,9 @@ - (id) initForReadingWithData: (NSData*)data
   NSXMLParser *theParser;
   NSData *theData = data;
 
+  // Dictionary which contains custom class information for Gorm/IB.
+  _customClasses = [[NSMutableArray alloc] initWithCapacity: 10];
```
fredkiefer, line 330: As the comment already suggests this should be a dictionary keyed on the custom class name. That way we don't need a loop when getting a loop to find the matching definitions.

Source/GSXibKeyedUnarchiver.m
```diff
@@ -143,8 +165,46 @@ - (NSData *) _preProcessXib: (NSData *)data
...
+                                                                       @"//object[@class=\"IBMutableOrderedSet\"][@key=\"objectRecords\"]"
+                                                                     @"/object/object[@class=\"IBObjectRecord\"]/int[@key=\"objectID\"]"
+                                                                     @"[text()=\"%@\"]/../reference[@key=\"object\"]/@ref", num];
+                                      
+                                      NSArray *refNodes = [document nodesForXPath: refXPath error: NULL];
+                                      if ([refNodes count] > 0)
+                                        {
+                                          NSXMLElement *refNode = [refNodes objectAtIndex: 0];
+                                          NSString *refId = [refNode stringValue];
+                                          NSDictionary *dict = [NSDictionary dictionaryWithObjectsAndKeys:
```
fredkiefer, line 201: You could use the method -createCustomClassRecordForId: withParentClass: forCustomClass: (By moving that method up one class). That way it is easier to change the storage structure we use.

Source/GSXibKeyedUnarchiver.m
```diff
@@ -143,8 +165,46 @@ - (NSData *) _preProcessXib: (NSData *)data
...
+                          if ([NSClassSwapper isInInterfaceBuilder] == YES)
+                            {
+                              NSString *xpath = [NSString stringWithFormat: @"//object[@class=\"IBClassDescriber\"]"
+                                                          @"//string[@key=\"className\"][text()=\"%@\"]"
+                                                          @"/../string[@key=\"superclassName\"]", cn];
+                              NSArray *descriptionObjs = [document nodesForXPath: xpath error: NULL];
+                              if ([descriptionObjs count] > 0)
+                                {
+                                  NSUInteger idx = [key rangeOfString: @"."].location;
+                                  if (idx != NSNotFound) // unlikely to be NSNotFound...
```
fredkiefer, line 186: Why don't you evaluate this before doing the xpath request?

Source/GSXibKeyedUnarchiver.m
gcasa: This would mean changing it here and in Gorm.  I would like to leave this how it is for the time being.

Source/GSXibKeyedUnarchiver.m
gcasa: Done.

Source/GSXibKeyedUnarchiver.m
gcasa: UGH... No time like the present.  It needs to be keyed on the customClass name and ID.   This way when the unarchiver is working in Gorm it can pick up the record based on the ID of the object being unarchived.

fredkiefer: I understand the first argument, although I don't share it, that this would require changes to Gorm too. But what do you mean by the second argument? Gorm just loops over all the elements in the collection. For that it is completely irrelevant whether the collection is an array or a dictionary.

gcasa: This is done.  Pushed this code.

Source/GSXibKeyedUnarchiver.m
gcasa: Done

gcasa:
@fredkiefer Please take another look. All issues addressed.

Source/GSXibKeyedUnarchiver.m
```diff
@@ -1033,4 +1097,26 @@ - (NSDictionary *) decoded
...
+                      withParentClass: (NSString *)parentClassName
+                       forCustomClass: (NSString *)customClassName
+{
+  if (theId == nil || customClassName == nil)
+    return;
+
+  NSMutableDictionary *dict =
+    [NSMutableDictionary dictionaryWithObjectsAndKeys:
+                           parentClassName, @"parentClassName",
+                         theId, @"id",nil];
```
fredkiefer, line 1116: if you need the class name separately in Gorm you need to add it here.

fredkiefer (approved):

Source/GSXibKeyedUnarchiver.m
gcasa: Nope, it is not needed separately.

PR #150, opened 2022-08-19
Don't misuse libgnustep-gui_INTERFACE_VERSION to define Resource Inst…

buzzdeee:
…all paths

$(GNUSTEP_INSTANCE)_INTERFACE_VERSION is used in gnustep-make to define
the version of the shared object.

As a packager, there might be various reasons to
override the $(GNUSTEP_INSTANCE)_INTERFACE_VERSION
in order to control the library version.

So I do for gnustep-gui on OpenBSD packages. The problem here is, that
the libgnustep-gui_INTERFACE_VERSION is additionally used to specify the path
where resources like Panels, Localizations etc. get installed, however,
in Source/NSApplication.m, the use of
GNUSTEP_GUI_MAJOR_VERSION.GNUSTEP_GUI_MINOR_VERSION

As long as the libgnustep-gui_INTERFACE_VERSION is not overridden with
some other value, everything is fine. However, as in my case, Applications
aren't able to find the Panels/Translations provided by -gui anymore.

Decided to just remove the definition of libgnustep-gui_INTERFACE_VERSION
in these three Makefiles, as it was only used once, and used the
GNUSTEP_GUI_MAJOR_VERSION.GNUSTEP_GUI_MINOR_VERSION directly.

Alternatively, libgnustep-gui_INTERFACE_VERSION in these Makefiles
could could have been renamed to avoid this conflict.

### DIFF
--- a/Panels/GNUmakefile.postamble
+++ b/Panels/GNUmakefile.postamble
@@ -28,8 +28,7 @@
 # it.
 
 include ../Version
-libgnustep-gui_INTERFACE_VERSION=$(GNUSTEP_GUI_MAJOR_VERSION).$(GNUSTEP_GUI_MINOR_VERSION)
-POSTAMBLE_INSTALL_DIR = $(GNUSTEP_LIBRARY)/Libraries/gnustep-gui/Versions/$(libgnustep-gui_INTERFACE_VERSION)/Resources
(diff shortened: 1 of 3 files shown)


fredkiefer (approved):
I am a bit unsure whether tis change might have any unexpected results but looking through the code in GNUstep make I could not find anything that relies on this value being set here.

buzzdeee:
thanks @fredkiefer

Issue #151, opened 2022-08-25
NSRadio button 10.7 behavior is not implemented in GNUstep


gcasa:
The behavior described here is not implemented in GNUtstep.

https://stackoverflow.com/questions/61243478/assign-radio-group-to-radio-buttons

I have created the following test: 

https://github.com/gcasa/NSRadio_test

Steps to recreate:

1. in Xcode on a mac create a new application
2. Add two radio buttons in the same view
3. Connect them to the same actions
4. Build it on gs using buildtool
5. test

Expected: Radio buttons should behave properly
Actual: Radio buttons do not behave correctly

gcasa:
This issue has been addressed and merged into master.

PR #152, opened 2022-08-27
Changes to address issue #151.  This should correct behavior of NSRadio buttons in GS

gcasa:
This should implement the behavior seen in 10.7+.  I am not 100% sure if this is the right approach, but it does work for now.   Please review @fredkiefer

### DIFF
--- a/Source/NSButton.m
+++ b/Source/NSButton.m
@@ -547,42 +547,42 @@ static id buttonCellClass = nil;
   return [_cell sound];
 }
 
+- (BOOL) _isRadioButton
+{
+  return ([[self cell] image] ==
+          [NSImage imageNamed: @"NSRadioButton"]);
+}
+
 - (void) mouseDown: (NSEvent *)event
 {
   [super mouseDown: event];
 
+  if ([self _isRadioButton] == NO)
+    return;
+  
   if ([[self superview] isKindOfClass: [NSMatrix class]] == NO)
   {
-    NSImage *i = [[self cell] image];
-    BOOL isRadio = (i == [NSImage imageNamed: @"NSRadioButton"]);
-  
-    if (isRadio)
+    NSView *sv = [self superview];
+    NSArray *subviews = [sv subviews];
+    
+    FOR_IN(NSView*, v, subviews)
       {
-        NSView *sv = [self superview];
-        NSArray *subviews = [sv subviews];
-        
-        FOR_IN(NSView*, v, subviews)
+        if ([v isKindOfClass: [NSButton class]])
           {
-          if ([v isKindOfClass: [NSButton class]])
-            {
-              NSButton *b = (NSButton *)v;
-              NSImage *image = [[b cell] image];
-              BOOL flag = (image == [NSImage imageNamed: @"NSRadioButton"]);
-              
-              if (flag)
-                {
-                  if ([self action] == [b action] && b != self)
-                    {
-                      if ([self state] == NSOnState)
-                        {
-                          [b setState: NSOffState];
-                        }
-                    }
-                }
-            }
+            NSButton *b = (NSButton *)v;
+            if ([b _isRadioButton])
+              {
+                if ([self action] == [b action] && b != self)
+                  {
+                    if ([self state] == NSOnState)
+                      {
+                        [b setState: NSOffState];
+                      }
+                  }
+              }
           }
-        END_FOR_IN(subviews);
       }
+    END_FOR_IN(subviews);
   }
 }
 
--- a/Source/NSButton.m
+++ b/Source/NSButton.m
@@ -36,8 +36,6 @@
 #import "AppKit/NSButtonCell.h"
 #import "AppKit/NSEvent.h"
 #import "AppKit/NSWindow.h"
-#import "AppKit/NSImage.h"
-#import "AppKit/NSMatrix.h"
 
 #import "GSFastEnumeration.h"
 
--- a/Source/NSButton.m
+++ b/Source/NSButton.m
@@ -547,16 +547,29 @@ static id buttonCellClass = nil;
   return [_cell sound];
 }
 
+// Implement 10.7+ radio button behavior
 - (BOOL) _isRadioButton
 {
   return ([[self cell] image] ==
           [NSImage imageNamed: @"NSRadioButton"]);
 }
 
-- (void) mouseDown: (NSEvent *)event
+- (void) _flipState: (NSButton *)b
 {
-  [super mouseDown: event];
+  if ([b _isRadioButton])
+    {
+      if ([self action] == [b action] && b != self)
+        {
+          if ([self state] == NSOnState)
+            {
+              [b setState: NSOffState];
+            }
+        }
+    }
+}
 
+- (void) _handleRadioStates
+{
   if ([self _isRadioButton] == NO)
     return;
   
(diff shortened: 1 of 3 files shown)


Source/NSButton.m
```diff
@@ -543,4 +547,43 @@ - (NSSound *) sound
   return [_cell sound];
 }
 
+- (void) mouseDown: (NSEvent *)event
```
fredkiefer, line 550: Are you sure this behaviour is only required for mouse interaction and not for keyboard triggered state changes?

Source/NSButton.m
```diff
@@ -543,4 +547,43 @@ - (NSSound *) sound
...
+{
+  [super mouseDown: event];
+
+  if ([[self superview] isKindOfClass: [NSMatrix class]] == NO)
+  {
+    NSImage *i = [[self cell] image];
```
fredkiefer, line 556: Maybe we should extract the decision whether an NSButton is a radio button into a separate helper method. That would make it easier to change the actual implementation when we find out that different criteria are needed.

Source/NSButton.m
```diff
@@ -543,4 +547,43 @@ - (NSSound *) sound
...
+            {
+              NSButton *b = (NSButton *)v;
+              NSImage *image = [[b cell] image];
+              BOOL flag = (image == [NSImage imageNamed: @"NSRadioButton"]);
+              
+              if (flag)
+                {
+                  if ([self action] == [b action] && b != self)
+                    {
+                      if ([self state] == NSOnState)
```
fredkiefer, line 576: This condition should be checked outside of the loop, probably right at the start of the method

Source/NSButton.m
```diff
@@ -543,4 +547,43 @@ - (NSSound *) sound
   return [_cell sound];
 }
 
+- (void) mouseDown: (NSEvent *)event
+{
+  [super mouseDown: event];
+
+  if ([[self superview] isKindOfClass: [NSMatrix class]] == NO)
```
fredkiefer, line 554: Did you test, whether the exclusion of NSMatrix is enforced on Cocoa? And if so, which other behaviour is expected there. Do we actually implement this behaviour?

fredkiefer (changes requested):
This looks like a good start

Source/NSButton.m
gcasa: Fixed.

Source/NSButton.m
gcasa: The NSMatrix class is deprecated on Cocoa.   I added this to avoid any potential conflicts with existing logic in GS.

Source/NSButton.m
gcasa: I believe it's needed for all.  I will find a way to do it in one location.

Source/NSButton.m
gcasa: Done.

Source/NSButton.m
```diff
@@ -543,4 +547,43 @@ - (NSSound *) sound
...
+    FOR_IN(NSView*, v, subviews)
+      {
+        if ([v isKindOfClass: [NSButton class]])
+          {
+            NSButton *b = (NSButton *)v;
+            if ([b _isRadioButton])
+              {
+                if ([self action] == [b action] && b != self)
+                  {
+                    if ([self state] == NSOnState)
```
fredkiefer, line 577: This check is still inside the loop.

Source/NSButton.m
fredkiefer: One option could be to call this logic from within the setState: method of the cell. The implementation would still need to be on the controller, that is here within NSButton though.

gcasa: I moved it to sendAction:toTarget: which I believe is central enough to be called in all cases.   Let me know if you consider this a good place for it as well.

Source/NSButton.m
```diff
@@ -543,4 +547,59 @@ - (NSSound *) sound
...
+          [NSImage imageNamed: @"NSRadioButton"]);
+}
+
+- (void) _flipState: (NSButton *)b
+{
+  if ([b _isRadioButton])
+    {
+      if ([self action] == [b action] && b != self)
+        {
+          if ([self state] == NSOnState)
```
fredkiefer, line 563: As I already commented twice this statement could be taken out of the loop. Maybe check it on line 576 or close to that. Or did you mean to get the state from b? That is `if ([b state] == NSOnState)`? That would make sense as well, maybe check both?

Source/NSButton.m
fredkiefer: You don't need to worry about that. The logic in NSMatix is based on cells not subviews. Normally an NSMatrix won't have subviews at all.

Source/NSButton.m
gcasa: I believe this should now be resolved.  I have moved it to a location in the code where it is called in all cases.

Source/NSButton.m
gcasa: I have removed the need for this check, so it is gone.

Source/NSButton.m
gcasa: Removed as stated.   It is no longer needed.

gcasa:
The repo https://gitub.com/gcasa/NSRadio_test contains the test for this change.

Source/NSButton.m
```diff
@@ -36,12 +36,20 @@
 #import "AppKit/NSButtonCell.h"
 #import "AppKit/NSEvent.h"
 #import "AppKit/NSWindow.h"
+#import "AppKit/NSImage.h"
```
fredkiefer, line 39: You no longer need these imports

Source/NSButtonCell.m
```diff
@@ -1922,4 +1922,19 @@ - (id) initWithCoder: (NSCoder*)aDecoder
...
 }
 
+// Implement 10.7 NSRadio button behavior...
+
+- (BOOL) _isRadio
+{
+  return ([self image] == [NSImage imageNamed: @"NSRadioButton"]);
+}
+
+- (NSInteger) nextState
```
fredkiefer, line 1932: I don't understand this method, could you please give me a short explanation? This surely implement the "prevent switch to off if button is currently on and is a radio button" part of your commit message but what is it needed for?

gcasa: nextState changes the state from the previous state to the state it should be in after... NSOffState -> NSOnState, etc.   On 10.7+ if you click on a radio button that is on it remains on.   This change is to ensure that happens.

Source/NSButton.m
gcasa: Removed.

Source/NSButtonCell.m
fredkiefer: Thank you for the explanation. Does this new behaviour happen for a single radio button as well? I understand how this makes sense for a group of radio buttons, but on a single it would prevent switching it off.

gcasa: I just tested it by removing one of the buttons from the test mentioned above.   You can also do this by adding another radio button connected to a different selector.   It does come on when clicked, but it does not go off when clicked again.

gcasa: @fredkiefer Let me know if this latest change looks good.

fredkiefer (approved):

PR #169, opened 2023-02-04
Implement NSAutoresizingMaskLayoutConstraint class

BennyKJohnson:
This PR implements the NSAutoresizingMaskLayoutConstraint class. This class is used to translate the legacy autoresizing mask and frame of a view into the equivalent NSLayoutConstraints.

### DIFF
--- a/Source/NSAutoresizingMaskLayoutConstraint.m
+++ b/Source/NSAutoresizingMaskLayoutConstraint.m
@@ -25,10 +25,10 @@
 
 @implementation NSAutoresizingMaskLayoutConstraint
 
-NSUInteger minXAttribute = 32;
-NSUInteger minYAttribute = 33;
-NSUInteger maxXAttribute = 36;
-NSUInteger maxYAttribute = 37;
+static NSUInteger minXAttribute = 32;
+static NSUInteger minYAttribute = 33;
+static NSUInteger maxXAttribute = 36;
+static NSUInteger maxYAttribute = 37;
 
 + (NSArray *)constraintsWithAutoresizingMask:
                  (NSAutoresizingMaskOptions) autoresizingMask
@@ -68,6 +68,9 @@ NSUInteger maxYAttribute = 37;
                                           bounds: bounds];
 
   NSMutableArray *constraints = [NSMutableArray arrayWithCapacity:4];
+
+  // The order of constraints was determined from MacOS behaviour.
+  // For the majority of users it will not matter, but we should assume that users have coupled to the ordering.
   if ((autoresizingMask & NSViewMinXMargin)
       && (autoresizingMask & NSViewMaxXMargin))
     {
--- a/Source/NSAutoresizingMaskLayoutConstraint.m
+++ b/Source/NSAutoresizingMaskLayoutConstraint.m
@@ -1,0 +1,292 @@
+/* Implementation of class NSAutoresizingMaskLayoutConstraint
+   Copyright (C) 2023 Free Software Foundation, Inc.
+
+   By: Benjamin Johnson <benjaminkylejohnson@gmail.com>
+   Date: 2023
+
+   This file is part of the GNUstep Library.
+
+   This library is free software; you can redistribute it and/or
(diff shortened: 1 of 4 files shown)


Source/NSAutoresizingMaskLayoutConstraint.m
```diff
@@ -0,0 +1,292 @@
...
+   You should have received a copy of the GNU Lesser General Public
+   License along with this library; if not, write to the Free
+   Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+   Boston, MA 02110 USA.
+*/
+#import "NSAutoresizingMaskLayoutConstraint.h"
+
+@implementation NSAutoresizingMaskLayoutConstraint
+
+NSUInteger minXAttribute = 32;
```
ivucica, line 28: These should be marked `static` so they are module-local, otherwise they pollute the global symbol namespace in the binary / process.

Source/NSAutoresizingMaskLayoutConstraint.m
```diff
@@ -0,0 +1,292 @@
...
+  NSMutableArray *constraints = [NSMutableArray arrayWithCapacity:4];
+  if ((autoresizingMask & NSViewMinXMargin)
+      && (autoresizingMask & NSViewMaxXMargin))
+    {
+      [constraints addObject: widthConstraint];
+      [constraints addObject: xConstraint];
+    }
+  else
+    {
+      [constraints addObject: xConstraint];
```
ivucica, line 79: From just this code, it is unclear why the order changes between the true and false paths of the if statement. Can you add mildly-detailed comments describing the reasoning?

ivucica, line 79: Is this correct for all of the remaining three cases:

- `autoresizingMask` has a flag for min X, but not max X
- `autoresizingMask` has a flag for max X, but not min X
- `autoresizingMask` has a flag for neither min X nor max X

If yes, can you write a comment explaining why?

If no, can you address this?

Same applies for Y.

ivucica (commented):
Just a driveby comment, as I am not familiar with the details of this class and it just came up during a meeting.

Source/NSAutoresizingMaskLayoutConstraint.m
BennyKJohnson: I'm not exactly clear on what you are asking. The purpose of the factory method is to convert a NSAutoresizingMask into the equivalent layout constraints. In order for a view to be correctly constrained it must have a x, y, width and height constraint. If the mask doesn't have a min/max flag for a dimension (x or y) we are going to use the frame origin dimension for the position dimensional constraint.

Source/NSAutoresizingMaskLayoutConstraint.m
BennyKJohnson: Updated

Source/NSAutoresizingMaskLayoutConstraint.m
BennyKJohnson: Yeah I agree this is strange behaviour. I've added a comment for clarity. It's the observed behaviour from MacOS. I don't think it matters that much for users but some users may have coupled themselves to the ordering, this includes me with my test suite I created for this class.

BennyKJohnson: Here is the test suite I wrote, which may offer a deeper understanding of the behaviour. I created the tests first on MacOS prior to writing the implementation. https://github.com/BennyKJohnson/ConstraintBasedLayoutTests/blob/master/NSAutoresizingLayoutConstraintTests.m

BennyKJohnson:
@gcasa @fredkiefer could I please get some feedback on this PR

fredkiefer:
Oh, sorry. I am just impressed by this and your other changes. I wanted to give the other GNUstep programmers some time to read through this without me interfering. That is why I mentioned it in the last GNUstep quarterly meeting. For me this just looks perfect.

fredkiefer (approved):
What we could do is add your external test code inside of GNUstep gui itself.

PR #170, opened 2023-02-14
Initial template for NSDictionaryController

gcasa:
This PR implements the NSDictionaryController functionality.

The test is here: https://github.com/gcasa/NSDictionaryController_test


Source/NSArrayController.m
```diff
@@ -585,6 +585,19 @@ - (void) encodeWithCoder: (NSCoder *)coder
     }
   else
     {
+      BOOL f;
+      f = _acflags.avoids_empty_selection;
+      [coder encodeValueOfObjCType: @encode(BOOL)
+                                at: &f];
+      f = _acflags.selects_inserted_objects;
```
fredkiefer, line 592: Why is `preservesSelection`missing here?

gcasa: Added missing data to non-keyed archiving logic.

gcasa:
The test for this class is located here: https://github.com/gcasa/NSDictionaryController_test

gcasa:
@fredkiefer Please take a look.  I believe it is functioning properly.

Source/NSDictionaryController.m
```diff
@@ -0,0 +1,480 @@
...
+  NSString *k = [obj key];
+
+  [super removeObject: obj];
+  [_contentDictionary removeObjectForKey: k];
+  [self rearrangeObjects];
+}
+
+- (void) removeObjects: (NSArray *)array
+{
+  [super addObjects: array];
```
fredkiefer, line 225: This should rather be `-removeObjects:`

gcasa: Oops!   Fixed

gcasa:
@fredkiefer Any further input?

gcasa:
@fredkiefer ??

gcasa:
@fredkiefer please take a look at these changes when you can.

Source/NSDictionaryController.m
```diff
@@ -0,0 +1,480 @@
...
+- (NSString *) _stringsFromDictionary
+{
+  NSEnumerator *en = [self keyEnumerator];
+  NSString *k = nil;
+  NSString *result = @"";
+
+  while ((k = [en nextObject]) != nil)
+    {
+      NSString *v = [self objectForKey: k];
+      result = [result stringByAppendingString:
```
fredkiefer, line 57: This format will only work for simple value types. Why not use one of the NSPropertyListSerialization formats?

Source/NSDictionaryController.m
```diff
@@ -0,0 +1,480 @@
...
+
+- (NSArray *) _buildArray: (NSDictionary *)content
+{
+  NSArray *allKeys = [content keysSortedByValueUsingSelector: @selector(compare:)];
+  NSMutableArray *result = [NSMutableArray arrayWithCapacity: [allKeys count]];
+
+  FOR_IN(id, k, allKeys)
+    {
+      NSDictionaryControllerKeyValuePair *kvp =
+	AUTORELEASE([[NSDictionaryControllerKeyValuePair alloc] init]);
```
fredkiefer, line 109: You could avoid the AUTORELEASE here by using a RELEASE after line 128 and in the skip case

Source/NSDictionaryController.m
```diff
@@ -0,0 +1,480 @@
...
+      id v = [content objectForKey: k];
+      NSString *localizedKey = [_localizedKeyDictionary objectForKey: k];
+
+      [kvp setLocalizedKey: localizedKey];
+      [kvp setKey: k];
+      [kvp setValue: v];
+      [kvp setExplicitlyIncluded: NO];
+
+      if ([_excludedKeys containsObject: k])
```
fredkiefer, line 118: This could be checked first before creating and setting kvp.

Source/NSDictionaryController.m
```diff
@@ -0,0 +1,480 @@
...
+	    triggerChangeNotificationsForDependentKey: @"arrangedObjects"];
+    }
+}
+
+- (void) _setup
+{
+  _contentDictionary = [[NSMutableDictionary alloc] init];
+  _includedKeys = [[NSArray alloc] init];
+  _excludedKeys = [[NSArray alloc] init];
+  _initialKey = @"key";
```
fredkiefer, line 160: Shouldn't we use ASSIGN here?

Source/NSDictionaryController.m
```diff
@@ -0,0 +1,480 @@
...
+- (void) observeValueForKeyPath: (NSString*)aPath
+		       ofObject: (id)anObject
+			 change: (NSDictionary*)aChange
+			context: (void*)aContext
+{
+  [NSException raise: NSInvalidArgumentException
+	      format: @"-%@ cannot be sent to %@ ..."
+	      @" create an instance overriding this",
+	      NSStringFromSelector(_cmd), NSStringFromClass([self class])];
+  return;
```
fredkiefer, line 353: Why this `return`?

fredkiefer (approved):
Looks good to me. You should check the comments I wrote but otherwise it is fine to merge.

gcasa:
> Looks good to me. You should check the comments I wrote but otherwise it is fine to merge.

Hey.  I saw the comment.  I will look into that before merging.

Source/NSDictionaryController.m
gcasa: Removed

Source/NSDictionaryController.m
gcasa: Since the initial retain count is 1, I think we are okay just using init methods here.  I have changed the strings to [[NSString alloc] initWithString: ...] so that the instance used is an NSString and not an NSConstantString.  Subtle difference, but I believe more correct.  In either case, they worked just fine in my test case.

Source/NSDictionaryController.m
gcasa: Yes, done.

Source/NSDictionaryController.m
gcasa: I feel like the AUTORELEASE addresses both cases, so it should be fine.   I think it keeps the code a little cleaner.  Let me know if you really would like me to do this the other way and I will do it as a quick fix.

Source/NSDictionaryController.m
gcasa: This shouldn't be an issue since it is only supposed to work for the strings format anyway.  I can find a function to convert from a strings list to a property list, but not the other way around. :/

gcasa: I will submit a quick fix for this if I find one or someone points it out.

PR #172, opened 2023-02-26
Implement bootstrapping auto layout

BennyKJohnson:


Source/NSLayoutConstraint.m
```diff
@@ -652,16 +653,92 @@ - (void) updateConstraints
 
 @implementation NSView (NSConstraintBasedLayoutInstallingConstraints)
 
+- (void) _bootstrapAutoLayout
+{
+  if ([self window])
+    {
+      [self.window _bootstrapAutoLayout];
```
fredkiefer, line 660: Could you please use `[self window]`here. All compilers support the dot notation, but some programmers don't like  it and mixing the two notations in one line just seems wrong.

Source/NSLayoutConstraint.m
```diff
@@ -652,16 +653,92 @@ - (void) updateConstraints
...
+
+- (void) layoutIfNeeded
+{
+  [self updateConstraintsIfNeeded];
+  [[self contentView] _layoutViewAndSubViews];
+}
+
+- (void) updateConstraintsIfNeeded
+{
+  [self.contentView updateConstraintsForSubtreeIfNeeded];
```
fredkiefer, line 736: Again `[self contentView]`should be used here and in the next method.

Source/NSView.m
```diff
@@ -5212,6 +5217,35 @@ - (void) _setNeedsUpdateConstraints: (BOOL)needsUpdateConstraints
...
+}
+
+- (GSAutoLayoutEngine*) _layoutEngine
+{  
+  return _layoutEngine;
+}
+
+- (void) _setLayoutEngine: (GSAutoLayoutEngine *)engine
+{
+  _layoutEngine = engine;
```
fredkiefer, line 5241: The reference handling for the layout engine is not clear to me. Who is responsible to retain and free it? The current implementation could leak it as well as release it multiple times. When a view gets removed from its superview and they have a shared layout engine.

fredkiefer (changes requested):

Source/NSLayoutConstraint.m
BennyKJohnson: Fixed

Source/NSView.m
```diff
@@ -5212,6 +5217,35 @@ - (void) _setNeedsUpdateConstraints: (BOOL)needsUpdateConstraints
...
+
+  NSArray *subviews = [self subviews];
+  FOR_IN (NSView *, subview, subviews)
+    [subview _setLayoutEngine:engine];
```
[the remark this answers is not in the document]
BennyKJohnson: Fixed

Source/NSView.m
BennyKJohnson: My intention was for it to be owned by the top most view of the view hierarchy. As the view hierarchy can be modified this introduces a lot of complexity. I've move the ownership of the layout engine to NSWindow, this simplifies things. However I am under the impression that AutoLayout works with a view without a window, so I might need to revisit this at a later stage.

Source/NSLayoutConstraint.m
```diff
@@ -652,16 +654,89 @@ - (void) updateConstraints
...
+}
+
+- (void) updateConstraintsIfNeeded
+{
+  [[self contentView] updateConstraintsForSubtreeIfNeeded];
+}
+
+- (void) _bootstrapAutoLayout
+{
+  [self _setLayoutEngine: [[GSAutoLayoutEngine alloc] init]];
```
fredkiefer, line 739: Most likely you should AUTORELEASE this here.

Source/NSView.m
```diff
@@ -5212,6 +5211,30 @@ - (void) _setNeedsUpdateConstraints: (BOOL)needsUpdateConstraints
...
+
+  NSArray *subviews = [self subviews];
+  FOR_IN (NSView *, subview, subviews)
+    [subview _layoutViewAndSubViews];
+  END_FOR_IN (subviews);
+}
+
+- (GSAutoLayoutEngine*) _layoutEngine
+{
+  if (!([self window] && [[self window] _layoutEngine]))
```
fredkiefer, line 5230: You don't need the second test. Just returning the value will be the same.

Source/NSLayoutConstraint.m
BennyKJohnson: I'm not clear on the reason for this. The layout engine should exist for the lifetime of the NSWindow. From my understanding (or lack of) the alloc init would have a +1 for the engine, which would be decremented to zero when the NSWindow is deallocated. Please explain.

Source/NSView.m
BennyKJohnson: Fixed

fredkiefer (changes requested):
Sorry, I did forget to send these two comments. Hopefully they make it clearer what I wanted to say with the AUTORELEASE comment. But I will try to explain it in more detail.

Source/NSLayoutConstraint.m
fredkiefer: Your thinking is correct here. The code as you wrote it is not a memory leak. The issue here is that you don't follow the established reference handling conventions of Objective-C. This could be confusing for readers of the code. One way to avoid this is to add some comments here and in the NSWindow code. Another way would be to stay with the expectations.

BennyKJohnson: No worries, fixed, thank you for the fast feedback while I'm still getting to grips with the conventions.

Source/NSWindow.m
```diff
@@ -6156,6 +6161,20 @@ - (void) _setWindow: (NSWindow *)window inactive: (BOOL)inactive
...
+@implementation NSWindow (NSConstraintBasedLayoutCorePrivateMethods)
+
+- (GSAutoLayoutEngine*) _layoutEngine
+{
+  return _layoutEngine;
+}
+
+- (void) _setLayoutEngine: (GSAutoLayoutEngine*)layoutEngine
+{
+  _layoutEngine = layoutEngine;
```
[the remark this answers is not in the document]
BennyKJohnson: Fixed

Source/NSWindow.m
```diff
@@ -1125,6 +1128,8 @@ - (id) initWithContentRect: (NSRect)contentRect
   _f.cursor_rects_enabled = YES;
   _f.cursor_rects_valid = NO;
 
+  _layoutEngine = nil;
```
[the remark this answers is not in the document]
BennyKJohnson: Fixed

Source/NSLayoutConstraint.m
```diff
@@ -652,16 +654,91 @@ - (void) updateConstraints
...
+- (void) updateConstraintsIfNeeded
+{
+  [[self contentView] updateConstraintsForSubtreeIfNeeded];
+}
+
+- (void) _bootstrapAutoLayout
+{
+  GSAutoLayoutEngine *layoutEngine = [[GSAutoLayoutEngine alloc] init];
+  [self _setLayoutEngine: layoutEngine];
+  AUTORELEASE(layoutEngine);
```
fredkiefer, line 741: Having this in a separate statement you could now even do a RELEASE, which is a bit cheaper..

Headers/AppKit/NSView.h
```diff
@@ -53,6 +53,7 @@
 @class NSScrollView;
 @class NSView;
 @class NSWindow;
+@class GSAutoLayoutEngine;
```
fredkiefer, line 56: Is this line still needed?

Source/NSWindow.m
```diff
@@ -6156,6 +6159,20 @@ - (void) _setWindow: (NSWindow *)window inactive: (BOOL)inactive
 }
 @end
 
+@implementation NSWindow (NSConstraintBasedLayoutCorePrivateMethods)
+
+- (GSAutoLayoutEngine*) _layoutEngine
```
fredkiefer, line 6164: You could also add a parameter to  this method, `createIfMissing` or so, and either return nil on missing or create one. That way you could remove the -_setLayoutEngine: and the -_bootstrapAutoLayout method and keep the creation of the GSAutoLayoutEngine in one place.

fredkiefer (approved):

fredkiefer:
Now it looks correct to me. There are still a few minor points you could change as for example take a look at the gcc warnings you can see in the checks. It would be great if somebody could test the code, I will only have time for that over the next weekend.

Source/NSWindow.m
BennyKJohnson: My reasoning for having a separate setter and lazy getter is so I could avoid the primary getter having a side effect,  this has certain tradeoffs as you mentioned.

Headers/AppKit/NSView.h
BennyKJohnson: Fixed

BennyKJohnson:
> Now it looks correct to me. There are still a few minor points you could change as for example take a look at the gcc warnings you can see in the checks. It would be great if somebody could test the code, I will only have time for that over the next weekend.

Cheers, I fixed one of the warnings introduced in this PR. I'm not really sure what you intent to test, the code doesn't do anything interesting because I'm trying my best to keep these PRs manageable. I hope soon it will start to come together and some automated tests can be introduced.

gcasa:
@BennyKJohnson if you're done with these changes, it's okay to merge them.  Please feel free to do so at your convenience.

BennyKJohnson:
@gcasa Please merge when ready, I don't think I have the ability to merge the PR.

gcasa:
Oh.  Ok.  I will do it then.  I will grant you that ability asap.

PR #176, opened 2023-03-17
Improvements to existing NSDataLink code

gcasa:
These changes will implement the rest of NSDataLink/NSDataLinkManager.   This class was not often used, but is relatively easy to implement.   This class was rarely used, but should be easy to finish.

### DIFF
--- a/Headers/AppKit/NSDataLink.h
+++ b/Headers/AppKit/NSDataLink.h
@@ -64,25 +64,25 @@ APPKIT_EXPORT_CLASS
   // Attributes
   @private
   // link info.
-  NSDataLinkNumber      linkNumber;
-  NSDataLinkDisposition disposition;
-  NSDataLinkUpdateMode  updateMode;
+  NSDataLinkNumber      _linkNumber;
+  NSDataLinkDisposition _disposition;
+  NSDataLinkUpdateMode  _updateMode;
 
   // info about the source.
-  NSDate                *lastUpdateTime;
-  NSString              *sourceApplicationName;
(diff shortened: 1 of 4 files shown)


rfm:
Looking at the diffs, it seems to me that the changes are almost all cosmetic (improving ivar names and changing whitespace).
I think it would make things far easier to review if there were separate changes:
1. a small set of functional changes to be reviewed carefully for correct logic
2. cosmetic changes, which can be passed very quickly

I generally try to do cosmetic stuff separately first.

gcasa:
> Looking at the diffs, it seems to me that the changes are almost all cosmetic (improving ivar names and changing whitespace). I think it would make things far easier to review if there were separate changes:
> 
> 1. a small set of functional changes to be reviewed carefully for correct logic
> 2. cosmetic changes, which can be passed very quickly
> 
> I generally try to do cosmetic stuff separately first.

They are at this point.  My original intention with this PR was to make the class actually work.   I implemented what is there a long time ago, but got stuck on how to monitor changes to a file without polling it (which is terribly inefficient) so, I wanted to come back to it and see if I could find a way to make it functional.  Sometimes i do the "cosmetic" stuff while I am thinking as it helps me think more cleanly.  I don't know if that makes sense, but it's my process. ;)

gcasa:
The test for this is here

https://github.com/gcasa/NSDataLink_test

I will incorporate this into the test suite after some refinement.

gcasa:
@copilot Add a ChangeLog entry for this.

PR #180, opened 2023-04-06
Implemented NSMenu removeAllItems method

qmfrederik:
Cherry-pick of https://github.com/keysight-eggplant/libs-gui/commit/61695272c58fc799cbd223b4054ba0114c0e2fb5


qmfrederik:
CI passed on my fork: https://github.com/qmfrederik/libs-gui/actions/runs/4633035130/, but is pending approval here.

Source/NSMenu.m
```diff
@@ -811,6 +811,15 @@ - (void) addItem: (id <NSMenuItem>)newItem
 			   atIndex: [_items count]];
 }
 
+- (void) removeAllItems
+{
+  [_items makeObjectsPerformSelector:@selector(setMenu:) withObject:nil];
+  [_items removeAllObjects];
+  _menu.needsSizing = YES;
+  [(NSMenuView*)_view setNeedsSizing: YES];
+  [self menuChanged];
```
fredkiefer, line 820: The big difference to removing each item separately is that you are not sending out any notification about that change. This might affect external NSMenu implementations, like the one that existed in Etoile. Did you leave this out on purpose or was it oversight?

qmfrederik: Good question.  Looks like [there isn't a notification for a bulk remove](https://developer.apple.com/documentation/appkit/nsmenudidremoveitemnotification), so would you expect a notification for item currently in the menu?

gcasa:
Are there still any outstanding issues with this? @fredkiefer @qmfrederik

Source/NSMenu.m
fredkiefer: Yes, that is the only solution I see. Just get the item count at the start of the method and add a notification for each.

gcasa: Why don't we do something akin to this here??:

```
- (void) removeAllItems
{
  NSUInteger count = [_items count];
  NSUInteger index = 0;
  
  for (index = 0; index < count; index++)
  {
    // since we are removing all items, the index of each item will change so we simply remove item 0 every time.
    [self removeItemAtIndex: 0]; 
  }
}
 ```
 
The removeItemAtIndex: method already contains all of the notification, cleanup, and resizing logic that is needed in this case.  Rather than repeating the code here for removing these items, we could just call the method.

Comment (gcasa) [reply Source/NSMenu.m:820]: Alternatively, following @fredkiefer's suggestion to the letter, we could simply do this...

```
- (void) removeAllItems
{
  NSUInteger count = [_items count]; 
  NSUInteger index = 0; 

  for (index = 0; index < count; index++) 
  { 
    NSDictionary *d = [NSDictionary dictionaryWithObject: [NSNumber numberWithInteger: index]
		                                                                     forKey: @"NSMenuItemIndex"];
    NSNotification *removed = [NSNotification
		notificationWithName: NSMenuDidRemoveItemNotification
		object: self
		userInfo: d];
		
    if (_menu.changedMessagesEnabled)
       [nc postNotification: removed];
    else
       [_notifications addObject: removed];
  }

  [_items makeObjectsPerformSelector: @selector(setMenu:) withObject: nil];
  [_items removeAllObjects];
  _menu.needsSizing = YES;
  [(NSMenuView*)_view setNeedsSizing: YES];
  [self menuChanged];
}  
```

While I think this would be faster (since we don't do a UI update on the removal of every item) I don't like it as much because of all of the repeated code.  These two comments are just my opinion and I could be wrong.  Please let me know what your thoughts are, please. :)

gcasa:
Similar or merged

Issue #182, opened 2023-05-25
NSCurrentServerThreadKey -> GSCurrentServerThreadKey should this be renamed?


gcasa:
https://github.com/gnustep/libs-gui/blob/fccf8771e6312216ab2ba442d1367c57dfae73eb/Source/GSDisplayServer.m#L65

This constant doesn't exist in either Cocoa or OpenStep.   Shouldn't this be renamed to have a GS prefix?

gcasa:
@fredkiefer the repo shows you committed this.  I can't find it in either the OPENSTEP or Cocoa documentation or header files.  Do you have further information on this?

I believe, for consistency, the prefix on this should be changed to be GS instead of NS.  I have looked into any apps this might break and the only one I see is gnumail and it will do so in a "soft" way.  That is to say it won't break its build.  However we should not allow inconsistencies to exist for the convenience of any one application. 

Let me know your thoughts.

fredkiefer:
I cannot remember adding this and the only commit I can find (2009) adds just another variable below this one. Which commit are you referring to?

I could not find any usage of this variable in the GNUstep code itself. To me it seems rather useless. So rather than renaming it we might get rid of it completely.

gcasa:
@fredkiefer My apologies. My eyesight must be bad. :) It was @afedor that added it.   I'm not sure why.  I think removing it entirely is a good idea.   If it isn't used otherwise, then it simplifies the code.   I am adding @rmottola here since GNUmail is his application and, apparently, the only one impacted.

rmottola:
@fredkiefer I came up with the same conclusion of removing it, GNUMail has just been updated.

PR #184, opened 2023-05-29
Remove now unused NSCurrentServerThreadKey key, previously it was use…

rmottola:
Remove now unused NSCurrentServerThreadKey key, previously it was used tor GSCurrentServer(), but it it is not needed anymore. Please use the function directly. Only known usage to me is GNUMail, it has just been converted.
Fixes https://github.com/gnustep/libs-gui/issues/182


fredkiefer (approved):

gcasa:
I discussed this with @rmottola at length.  Since @fredkiefer has approved, I am proceeding with merging this.  GC

PR #185, opened 2023-05-30
NSDockTile - Implement methods to update the icon and show badge

gcasa:
NSDockTile - Implement methods to update the icon and show badge


Source/NSDockTile.m
```diff
@@ -97,7 +108,19 @@ - (NSString *) badgeLabel
 
 - (void) setBadgeLabel: (NSString *)label
 {
+  NSImage *tempImage;
+  NSImageRep *rep;
+
   ASSIGNCOPY(_badgeLabel, label);
+
+  tempImage = [[NSImage alloc] initWithSize:[_appIconImage size]];
```
fredkiefer, line 116: Here and in the following lines a few spaces are missing after the colons.

Source/NSDockTile.m
```diff
@@ -97,7 +108,19 @@ - (NSString *) badgeLabel
...
+
+  rep = [[NSCustomImageRep alloc] initWithDrawSelector:@selector(draw) delegate:self];
+  [rep setSize: [_appIconImage size]];
+  [tempImage addRepresentation:rep];
+  [rep release];
```
fredkiefer, line 121: Why not use the `RELEASE`macro?

Source/NSDockTile.m
```diff
@@ -109,4 +132,74 @@ - (void) display
...
+  NSPoint text_location;
+  NSRect disc_rect;
+  NSSize disc_size;
+  int image_width, pad;
+  NSBezierPath *p;
+  NSPoint point;
+  CGFloat radius;
+  NSSize imageSize;
+
+  if (_showsApplicationBadge)
```
fredkiefer, line 147: What will get drawn if this is false but a label was set?

Source/NSDockTile.m
```diff
@@ -109,4 +132,74 @@ - (void) display
...
+    {
+      [_appIconImage compositeToPoint:NSZeroPoint operation:NSCompositeCopy];
```
fredkiefer, line 149: Again spaces are missing

Source/NSDockTile.m
```diff
@@ -109,4 +132,74 @@ - (void) display
...
+
+      imageSize = [_appIconImage size];
+      attrs = [[NSMutableDictionary alloc] init];
+      [attrs setObject: [NSFont boldSystemFontOfSize: imageSize.width/5]  forKey: NSFontAttributeName];
+      [attrs setObject: [NSColor blackColor]  forKey: NSForegroundColorAttributeName];
+
+      disc_size = [_badgeLabel sizeWithAttributes: attrs];
+
+      image_width = imageSize.width;
```
fredkiefer, line 158: This should be moved up a bit and used in line 153. Also a different name would be great as we use this value both for width and height and it is always confusing to see width and height in one computation.

Source/NSDockTile.m
```diff
@@ -109,4 +132,74 @@ - (void) display
...
+
+      pad = image_width / 10;
+      disc_size.height += pad;
+      disc_size.width += pad;
+
+
+      disc_rect = NSMakeRect(image_width-disc_size.width-4,
```
fredkiefer, line 165: Again a few spaces would help.

Source/NSDockTile.m
```diff
@@ -109,4 +132,74 @@ - (void) display
...
+			     image_width-disc_size.height-4,
+			     disc_size.width,
+			     disc_size.height);
+
+      text_location = NSMakePoint(image_width - (disc_size.width - ((disc_size.width - [_badgeLabel sizeWithAttributes: attrs].width)  * 0.5))-4,
```
fredkiefer, line 170: Why `[_badgeLabel sizeWithAttributes: attrs]`and not `disc_size`? That would help to let you see how to further simplify this code :-)

Source/NSDockTile.m
```diff
@@ -109,4 +132,74 @@ - (void) display
...
+				  image_width - (disc_size.height - ((disc_size.height - [_badgeLabel sizeWithAttributes: attrs].height) * 0.5))-4);
+
+      [[NSColor colorWithDeviceRed: 1.0
```
fredkiefer, line 173: Should this colour be configurable?

fredkiefer (changes requested):
Looks already great, a bit of clean up is needed though.

Source/NSDockTile.m
rmottola: no problem...

Source/NSDockTile.m
rmottola: copy&paste from GNUMail code :) fixed

Source/NSDockTile.m
rmottola: well disc_size has some padding added. However I think it can be simplified after subsitution to just pad.

```
      text_location = NSMakePoint(image_width - (disc_size.width - pad  * 0.5)) - 4,
				  image_width - (disc_size.height - pad * 0.5)) - 4);
```

Source/NSDockTile.m
rmottola: Didn't think about it yet... maybe in a second step? Define some new system colors? Then also update Thematic :)
For the time being I refined the code to use 3 colors and also draw something nicer.

rmottola:
I add another issue - would be interesting to test on Mac.
What happens if the string is too long? Does the font size diminish? get it truncated? "..." added for truncation?
Determining the maximum number of chars displayed is never trivial.
Hints from Apple doc or behaviour welcome.

gcasa:
> I add another issue - would be interesting to test on Mac.
> 
> What happens if the string is too long? Does the font size diminish? get it truncated? "..." added for truncation?
> 
> Determining the maximum number of chars displayed is never trivial.
> 
> Hints from Apple doc or behaviour welcome.

In my experience it gets truncated after about 5-6 characters. Just my 0.02.

Source/NSDockTile.m
rmottola: I fixed this, do you like it, @fredkiefer? I always composite the image.

Source/NSDockTile.m
rmottola: I cleaned up all width/height, removed some local variables too and use diff of sizes instead of fixed padding. How do you like it,  @fredkiefer ?

Source/NSDockTile.m
rmottola: Quite tidied up... please check.

rmottola:
> > I add another issue - would be interesting to test on Mac.
> > What happens if the string is too long? Does the font size diminish? get it truncated? "..." added for truncation?
> > Determining the maximum number of chars displayed is never trivial.
> > Hints from Apple doc or behaviour welcome.
> 
> In my experience it gets truncated after about 5-6 characters. Just my 0.02.

Tested on mac, it gets truncated indeed... we have smaller icons. I implemented truncating with size 5.

rmottola:
I get a warning since i need UTF-16 to encode the ellipsis for truncation... do you know another way of putting it in?

Source/NSDockTile.m
```diff
@@ -32,11 +44,16 @@ - (instancetype) init
   self = [super init];
   if (self != nil)
     {
-      NSRect rect = NSMakeRect(0,0,48,48);
-      _size = rect.size;
+      GSDisplayServer *server = GSCurrentServer();
+      NSSize size = [server iconSize]; 
+      NSRect rect = NSMakeRect(0,0,size.width,size.height);
```
fredkiefer, line 49: Here there are still no spaces after the commas.

Source/NSDockTile.m
```diff
@@ -90,13 +108,112 @@ - (NSString *) badgeLabel
 
 - (void) setBadgeLabel: (NSString *)label
 {
+  NSImage *tempImage;
+  NSImageRep *rep;
+
   ASSIGNCOPY(_badgeLabel, label);
+
+  tempImage = [[NSImage alloc] initWithSize: [_appIconImage size]];
```
fredkiefer, line 116: This is now called whenever the label changes. Shouldn't the code rather go into a method like `setShowsApplicationBadge`?

Source/NSDockTile.m
```diff
@@ -90,13 +108,112 @@ - (NSString *) badgeLabel
...
+      NSColor *badgeTextColor;
+      NSString *displayString;
+
+      badgeBackColor = [NSColor redColor];
+      badgeDecorationColor = [NSColor lightGrayColor];
+      badgeTextColor = [NSColor whiteColor];
+
+      displayString = _badgeLabel;
+      if ([_badgeLabel length] > 5)
+	displayString = [NSString stringWithFormat:@"%@\u2026%@",[_badgeLabel substringToIndex:2],  [_badgeLabel substringFromIndex:[_badgeLabel  length]-2]];
```
fredkiefer, line 162: Distributing the spaces more evenly would help :-)

Source/NSDockTile.m
fredkiefer: It could still be simplified, but most likely the compiler will do that by itself.

Source/NSDockTile.m
```diff
@@ -90,13 +108,112 @@ - (NSString *) badgeLabel
...
+  RELEASE(rep);
+
+  [NSApp setApplicationIconImage: tempImage];
 }
 
 - (void) display
 {
   [_contentView setNeedsDisplay: YES];
+
+  if (_showsApplicationBadge)
```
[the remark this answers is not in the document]
rmottola: Indeed.. the actual drawing with the Custom Image Representation is not really view based, so I remove this again.

Source/NSDockTile.m
rmottola: sorry, mistyped·... but a hint about unicode ellispis?

gcasa:
> I get a warning since i need UTF-16 to encode the ellipsis for truncation... do you know another way of putting it in?

I don't think Cocoa uses ellipsis when it truncates the value.  Of course what we do is up to us.

Source/NSDockTile.m
```diff
@@ -90,13 +108,112 @@ - (NSString *) badgeLabel
...
+
+  tempImage = [[NSImage alloc] initWithSize: [_appIconImage size]];
+
+  rep = [[NSCustomImageRep alloc] initWithDrawSelector: @selector(draw) delegate: self];
+  [rep setSize: [_appIconImage size]];
+  [tempImage addRepresentation: rep];
+  RELEASE(rep);
+
+  [NSApp setApplicationIconImage: tempImage];
 }
```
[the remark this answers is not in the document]
rmottola: Fixed...

rmottola:
> > I get a warning since i need UTF-16 to encode the ellipsis for truncation... do you know another way of putting it in?
> 
> I don't think Cocoa uses ellipsis when it truncates the value. Of course what we do is up to us.



> 



> > I get a warning since i need UTF-16 to encode the ellipsis for truncation... do you know another way of putting it in?
> 
> I don't think Cocoa uses ellipsis when it truncates the value. Of course what we do is up to us.

Yes it does, it does. It displays 3 dots. It is common on Apple (e.g. menus and other parts of UI) to use ellipsis.
The string line above works, I just get a warning on my gcc version, didn't try older/newer or other compilers yet.

Headers/AppKit/NSUserActivity.h
```diff
@@ -0,0 +1,47 @@
+/* Definition of class NSUserActivity
```
fredkiefer, line 1: Why is this file included in this PR? It doesn't do any harm but most likely it was not planed to be here.

Source/NSApplication.m
```diff
@@ -2458,6 +2459,14 @@ - (NSWindow*) iconWindow
   return _app_icon_window;
 }
 
+- (NSDockTile *) dockTile
+{
+  if (!_dock_tile)
+    {
+      _dock_tile = [[NSDockTile alloc] init];
```
fredkiefer, line 2466: This needs to be released in the `dealloc`method.

Source/NSDockTile.m
```diff
@@ -45,6 +64,8 @@ - (oneway void) release
 {
   RELEASE(_contentView);
   RELEASE(_badgeLabel);
+  RELEASE(_appIconImage);
```
fredkiefer, line 67: As you did not retain this image a release is not needed. Or rather a retain would be needed on line 56.

Source/NSDockTile.m
fredkiefer: Now this code here makes even less sense.

Source/NSDockTile.m
```diff
@@ -90,13 +111,104 @@ - (NSString *) badgeLabel
...
+
+      badgeBackColor = [NSColor redColor];
+      badgeDecorationColor = [NSColor lightGrayColor];
+      badgeTextColor = [NSColor whiteColor];
+
+      imageSize = [_appIconImage size];
+
+      displayString = _badgeLabel;
+      if ([_badgeLabel length] > 5)
+	displayString = [NSString stringWithFormat:@"%@\u2026%@", [_badgeLabel substringToIndex:2], [_badgeLabel substringFromIndex:[_badgeLabel  length]-2]];
```
fredkiefer, line 158: Our coding standard suggests to use {} even for one line if statements. Also a few spaces are still missing in this line (at least four).

Source/NSApplication.m
rmottola: fixed

Source/NSDockTile.m
rmottola: should be fixed too

Source/NSDockTile.m
rmottola: I counted three... I don't like the {} for one liners, I'm decidedly against it, but GUI is your code.. so should be compliant to your liking now

gcasa: > I counted three... I don't like the {} for one liners, I'm decidedly against it, but GUI is your code.. so should be compliant to your liking now

It is part of the GNU coding standard.  I, personally, think it makes things clearer and easier to follow, but to each their own.

Source/#NSDockTile.m#
```diff
@@ -0,0 +1,216 @@
...
+}
+
+- (BOOL) showsApplicationBadge
+{
+  return _showsApplicationBadge;
+}
+
+- (void) setShowsApplicationBadge: (BOOL)flag
+{
+  _showsApplicationBadge = flag;
```
gcasa: This method and setBadgeLabel: should both simply set the properties/ivar.   The display method should be called in each of these methods so that it can re-render and display the icon appropriately.

Source/#NSDockTile.m#
```diff
@@ -0,0 +1,216 @@
...
+}
+
+- (NSString *) badgeLabel
+{
+  return _badgeLabel;
+}
+
+- (void) setBadgeLabel: (NSString *)label
+{
```
gcasa: Same comment as above for setShowsApplicationBadge.  These methods should have no logic other than setting the ivar.  They should call [self display] to re-render the icon.

Source/#NSDockTile.m#
```diff
@@ -0,0 +1,216 @@
...
+  [rep setSize: [_appIconImage size]];
+  [tempImage addRepresentation: rep];
+  RELEASE(rep);
+
+  [NSApp setApplicationIconImage: tempImage];
+}
+
+- (void) display
+{
+  [_contentView setNeedsDisplay: YES];
```
gcasa: Render all icon content HERE and only here.   This includes the subview which isn't implemented.   This way all display logic for this class is centralized.

Source/NSDockTile.m
fredkiefer: Could you please change the `retain` in line 56 to `RETAIN()`. Just to be consistent.

Source/#NSDockTile.m#
fredkiefer: You commented on an outdated version of the code. This does not help as it is hard to say whether this comments are still relevant or not.

rmottola: THe code uses a NSCustomImageRep, thus it does not use the draw method directly, since that one has its own delegate. It is a side-effect of not being able to just "draw" a standard bitmap image rep.

Furthermore setShowsApplicationBadge is strange... apple's Doc says it exists since 10.5, as the DockTile and badge itself. However it says this strange thing:

> In macOS 10.5 and later, application tiles do not support the application badge.

What does this mean? Before 10.5 the badge didn't even exist... maybe application tile != dock tile and this tries to specify this?


and trying (and googling confirms) on my mac, I get this:
`setShowsApplicationBadge: is not yet implemented for the NSApp dockTile`
I don't know if it has been implemented very recently, but apparently up to Big Sur this is the message...

Source/#NSDockTile.m#
rmottola: should..
It is the best place found to allocate and set the tempImage as Application Image.
One would thin it is best to do it it on init, but that doesn't work.
Set each time it changes works.
One could do it at each allocation, but it would be a waste.

What I did appears to be the optimum empyrically found. I don't know why it can be done on init.
Doint it on setBadgeLevel still has the advantage that no resources are wasted if no badge is set.

Source/NSDockTile.m
fredkiefer: Did you see these comments here? The code here now belongs into the constructor or the `setShowsApplicationBadge`method in the case where we switch the display on and the off case would be to undo this.

rmottola: Yes, I saw then... but I also commented that it doesn't work. If I allocate the tempImage either in init or in setShowsApplicationBadge, then the application hangs and no badge is displayed.
Furthermore I commented that in all MacOS versions I have at hand, setShowsApplicationBadge does nothing beyond throwing a log of being not implemented.
This still means we can "respect" that setting of course, but I'd be careful in putting code there.

rmottola:
to sum up the hacking session of yesterday, @fredkiefer @gcasa 
While remaining for now NSCustomImageRep and its draw method. We use the AppIcon view now, so we do not need to reallocate the image every time we change the badge. In theory it works, is cleaner and more efficient.
Right now, it works only the first time. We tracked it down to caching. If I explicitly disable the caching code in NSImage, it works.
It is a mystery however why explicitly setting caching to none doesn't work. I would say "bug one".
Furthermore, probably "bug two" caching of custom image reps should either not happen or only under certain conditions.

fredkiefer:
As far as I remember from last night this description is not fully correct. After you moved the switching off of the caching to the right image it worked. So, the first bug you list does not exist, at least that is what I remember. The second one, that custom drawn images should never be cached, looks like a real issue, but this code here has a work around for that.

Source/NSDockTile.m
```diff
@@ -45,6 +71,9 @@ - (oneway void) release
 {
   RELEASE(_contentView);
   RELEASE(_badgeLabel);
+  RELEASE(_appIconImage);
+  RELEASE(_imageRep);
```
fredkiefer, line 75: This could now be released after line 64.

Source/NSDockTile.m
```diff
@@ -91,12 +121,101 @@ - (NSString *) badgeLabel
 - (void) setBadgeLabel: (NSString *)label
 {
   ASSIGNCOPY(_badgeLabel, label);
+  [self display];
 }
 
 - (void) display
 {
-  [_contentView setNeedsDisplay: YES];
+  [_contentView display];
```
fredkiefer, line 129: This should be `setNeedsDisplay:`, you forgot to change it back.

rmottola:
> As far as I remember from last night this description is not fully correct. After you moved the switching off of the caching to the right image it worked. So, the first bug you list does not exist, at least that is what I remember. The second one, that custom drawn images should never be cached, looks like a real issue, but this code here has a work around for that.

You remember correct. Disabling the cache is an effective workaround. I opened an issue separate from this.
If you like the last commits regarding releases, is this pull request now clean?

gcasa:
I'll merge once @rmottola says ok.  I am not sure he merge can since I opened this. He has been the one principally working on it.

PR #186, opened 2023-06-15
Implement GSCSTableau, GSCSEditInfo, GSCSEditVariableManager

BennyKJohnson:
This PR implements the supporting classes for the GSCSCassowarySolver. 

I expect there to be a couple more PRs before the whole solution is in place :), one for the cassowary solver itself and one to connect everything up.

### DIFF
--- a/Source/GSCSTableau.m
+++ b/Source/GSCSTableau.m
@@ -1,0 +1,700 @@
+/* Copyright (C) 2023 Free Software Foundation, Inc.
+
+   By: Benjamin Johnson
+   Date: 12-6-2023
+   This file is part of the GNUstep Library.
+
+   This library is free software; you can redistribute it and/or
+   modify it under the terms of the GNU Lesser General Public
+   License as published by the Free Software Foundation; either
+   version 2.1 of the License, or (at your option) any later version.
+
+   This library is distributed in the hope that it will be useful,
+   but WITHOUT ANY WARRANTY; without even the implied warranty of
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+   Lesser General Public License for more details.
+
+   You should have received a copy of the GNU Lesser General Public
+   License along with this library; if not, write to the Free
+   Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+   Boston, MA 02110 USA.
+*/
+
+#import "GSCSTableau.h"
+#import "GSCSFloatComparator.h"
+#import "GSFastEnumeration.h"
+
+@interface GSCSTableau (PrivateMethods)
+
+- (void) recordUpdatedVariable: (GSCSVariable *)variable;
+
+- (void) addTermVariablesForExpression: (GSCSLinearExpression *)expression
+                              variable: (GSCSVariable *)variable;
+
+- (void) substituteOutTermInExpression: (GSCSLinearExpression *)expression
+                    newExpression:
+                      (GSCSLinearExpression *)newExpression
+                          subject: (GSCSVariable *)subject
+                            term: (GSCSVariable *)term
+                      multiplier: (CGFloat)multiplier;
+
+- (CGFloat) calculateNewCoefficent: (CGFloat)coefficentInExistingExpression
+         coefficentInNewExpression: (CGFloat)coefficentInNewExpression
+                        multiplier: (CGFloat)multiplier;
+
+- (void) removeColumnVariable: (GSCSVariable *)variable
+                      subject: (GSCSVariable *)subject;
+
+- (void) recordUpdatedVariable: (GSCSVariable *)variable;
+
+- (void) addNewExpression: (GSCSLinearExpression *)newExpression
+             toExpression: (GSCSLinearExpression *)existingExpression
+                        n: (CGFloat)n
+                  subject: (GSCSVariable *)subject;
+
+- (NSString *) columnsDescription;
+
+- (NSString *) rowsDescription;
+
+@end
+
+@implementation GSCSTableau
+
+- (instancetype) init
+{
+  self = [super init];
+  if (self != nil)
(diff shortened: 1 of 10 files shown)


Source/GSCSTableau.m
```diff
@@ -0,0 +1,700 @@
...
+{
+  GSCSLinearExpression *expression = [_rows objectForKey: variable];
+  if (expression == nil)
+    {
+      NSException *missingExpressionException = [NSException
+        exceptionWithName: NSInvalidArgumentException
+                   reason: @"No expression exists for the provided variable"
+                 userInfo: nil];
+      [missingExpressionException raise];
+      return;
```
fredkiefer, line 216: Why a `return` here, the raise should never return.

Source/GSCSTableau.m
```diff
@@ -0,0 +1,700 @@
...
+      [expression addVariable: term coefficient: updatedCoefficent];
+      [self addMappingFromExpressionVariable: term toRowVariable: subject];
+    }
+}
+
+- (CGFloat) calculateNewCoefficent: (CGFloat)coefficentInExistingExpression
+         coefficentInNewExpression: (CGFloat)coefficentInNewExpression
+                        multiplier: (CGFloat)multiplier
+{
+  CGFloat newCoefficent
```
fredkiefer, line 326: Why use an intermediate variable?  You could just return the computation.

Source/GSCSTableau.m
```diff
@@ -0,0 +1,700 @@
...
+
+- (GSCSLinearExpression *) rowExpressionForVariable: (GSCSVariable *)variable
+{
+  return [_rows objectForKey: variable];
+}
+
+- (void) pivotWithEntryVariable: (GSCSVariable *)entryVariable
+                   exitVariable: (GSCSVariable *)exitVariable
+{
+// The expression "expr" represents the exit variable, which is about to be removed from the basis.
```
fredkiefer, line 464: Not sure what this comment refers to as there is no variable "expr"

Source/GSCSTableau.m
```diff
@@ -0,0 +1,700 @@
...
+}
+
+- (NSString *) description
+{
+  NSMutableString *description =
+    [NSMutableString stringWithString: @"Tableau Information\n"];
+  [description appendFormat: @"Rows: %ld (%ld constraints)\n", [_rows count],
+                            [_rows count] - 1];
+  [description appendFormat: @"Columns: %ld\n", [_columns count]];
+  [description appendFormat: @"Infesible rows: %ld\n", [_infeasibleRows count]];
```
fredkiefer, line 493: Here an "a" seems to be missing

Source/GSCSTableau.m
```diff
@@ -0,0 +1,700 @@
...
+  [description
+    appendFormat: @"External basic variables: %ld\n", [_externalRows count]];
+  [description appendFormat: @"External parametric variables: %ld\n\n",
+                            [_externalParametricVariables count]];
+
+  [description appendFormat: @"Columns: \n"];
```
fredkiefer, line 499: Why `appendFormat:`, this looks like a string to me.

fredkiefer (commented):
I must admit, I don't understand most of this code. It looks like a mix of debugging output and expression solving to me. And I still think that we should move all this code into a subdirectory to keep it apart from the rest of GNUstep gui,

Source/GSCSTableau.m
```diff
@@ -0,0 +1,700 @@
...
+    {
+      [_externalRows removeObjectForKey: variable];
+    }
+  NSArray *expressionTermVariables = [expression termVariables];
+  FOR_IN(GSCSVariable *, expressionTermVariable, expressionTermVariables)
+    [self removeMappingFromExpressionVariable: expressionTermVariable
+                                toRowVariable: variable];
+    if ([expressionTermVariable isExternal])
+      {
+        [_externalParametricVariables addObject: expressionTermVariable];
```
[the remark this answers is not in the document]
BennyKJohnson: This is correct, if the expression term variable is external it needs to be solved for.

Source/GSCSTableau.m
BennyKJohnson: Fixed

Source/GSCSTableau.m
BennyKJohnson: Fixed

Source/GSCSTableau.m
BennyKJohnson: Fixed

Source/GSCSTableau.m
BennyKJohnson: Fixed

Source/GSCSTableau.m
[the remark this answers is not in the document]
fredkiefer: Could you please explain this a bit more? Perhaps this is easier for me to accept if you call `recordUpdatedVariable:` here.

fredkiefer:
My main concern is really that we have so many new files in the Source directory. It would be easier to have them in a separate subdirectory. We haven't done that in gui before but base started that some time ago and it helps keeping track of files that belong together.

Source/GSCSTableau.m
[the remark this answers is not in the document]
BennyKJohnson: Honestly it's been a while since I wrote the code and I can't recall it's purpose. I have taken a look at the original implementation found here https://sourceforge.net/projects/cassowary/ and I can't find the check either. I have removed it.

BennyKJohnson:
> My main concern is really that we have so many new files in the Source directory. It would be easier to have them in a separate subdirectory. We haven't done that in gui before but base started that some time ago and it helps keeping track of files that belong together.

Hey @fredkiefer, this makes sense. I have done this in a separate PR just to keep the changes clean. https://github.com/BennyKJohnson/libs-gui/pull/1

PR #188, opened 2023-08-14
Fixed contentRectForFrameRect

austintatiousness:
fixed contentRectForFrameRect to take better account of userSpaceScaleFactor


austintatiousness:
Something is still off. Will pull again if verified fixed.

PR #189, opened 2023-08-14
Updated contentRectForFrameRect:

austintatiousness:
contentRectForFrameRect  updated to scale origin by userSpaceScaleFactor.

### DIFF
--- a/Source/GSWindowDecorationView.m
+++ b/Source/GSWindowDecorationView.m
@@ -108,8 +108,8 @@ static inline NSRect RectWithSizeScaledByFactor(NSRect aRect, CGFloat factor)
       
       aRect = RectWithSizeScaledByFactor(aRect, 1/factor);
     } else {
-  	aRect.size.width -= l + (r);
-  	aRect.size.height -= t + (b);
+  	aRect.size.width -= l + r;
+  	aRect.size.height -= t + b;
   	aRect.origin.x += (l);
  	aRect.origin.y += (b);
     }
--- a/Source/GSWindowDecorationView.m
+++ b/Source/GSWindowDecorationView.m
@@ -101,17 +101,18 @@ static inline NSRect RectWithSizeScaledByFactor(NSRect aRect, CGFloat factor)
 
       CGFloat factor = [[NSScreen mainScreen] userSpaceScaleFactor]; 
 
-      aRect.size.width -= (l + r)/factor;
-      aRect.size.height -= (t + b)/factor;
+      aRect.size.width -= (l + r);
+      aRect.size.height -= (t + b);
       aRect.origin.x += (l/factor);
       aRect.origin.y += (b/factor);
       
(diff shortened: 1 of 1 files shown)


fredkiefer:
I am really unsure about this change. At least it need so also adjust the following method, as this tries to do the exact opposite and they need to be in synch. But most likely we are fixing an issue here at the wrong place.
Which sort of window decoration are you using? The ones provided by the window manager or are they drawn by GNUstep itself. In the later case the issue might be instead in the class GSStandardWindowDecorationView where the offset method takes the scale factor only for one of the four values into account.

austintatiousness:
This affected every window whose widgets are drawn by a WindowMaker where the GSScaleFactor was set to 2.

austintatiousness:
I have been pursuing the code and I am lead to my original conclusion that is also coupled with the facts presented on the mailing list. The frame of a window is always defined in pixels, regardless of scale factor. The following is the code as presented currently in the GNUStep repository. 

```
+ (NSRect) contentRectForFrameRect: (NSRect)aRect
			 styleMask: (NSUInteger)aStyle
{
  float t = 0.0, b = 0.0, l = 0.0, r = 0.0;

  [self offsets: &l : &r : &t : &b forStyleMask: aStyle];
  aRect.size.width -= l + r;
  aRect.size.height -= t + b;
  aRect.origin.x += l;
  aRect.origin.y += b;
 
  if (0 == (aStyle & NSUnscaledWindowMask))
    {
      // FIXME: This method should probably take a screen parameter
      // rather than assuming the mainScreen
      
      CGFloat factor = [[NSScreen mainScreen] userSpaceScaleFactor]; 
      aRect = RectWithSizeScaledByFactor(aRect, 1/factor);
    }
 return aRect;
}
```

As you can see, it scales only the size of the `aRect` by the scale factor, as it should. The origin of the NSRect in most cases, I assume, will be (0,0), as most decorations provided by the window manager don't have left, right, or bottom insets.  However, WindowMaker does have a bottom inset for the bottom decorations of about 10 or 11 pixels. Because the result of this function will be multiplied by the scale factor, instad of it being inset by 10 pixels, it gets inset by 20 pixels (up from the bottom). You would never notice this in GNOME, for example, because there is no bottom decoration. 

Instead of just `RectWithSizeScaledByFactor(NSRect, CGFloat)` we need a function `RectScaledByFactor(NSRect, CGFloat)`

In my version of the code, I simply also scale aRect.origin.x and aRect.origin.y by doing the following 

```
 aRect.origin.x += l / factor;
 aRect.origin.y += b / factor;
```

I have tested this on all window managers that I have installed and it persists.  Additionally,I think that  we will need to adjust `+ (NSRect) frameRectForContentRect: (NSRect)aRect styleMask: (NSUInteger)aStyle` as well account for the origin as well. 

Again, this would go unnoticed on any platform that doesn't have bottom decorations.

PR #190, opened 2023-08-26
NSMenuItemCell: Fix small typo that causes display character check to fail

eukara:
NSMenu's don't seem to display the Shift character. I believe this change makes it display as intended.


fredkiefer:
Thank you for this patch!

PR #193, opened 2023-08-30
Update code in GSXib5KeyedUnarchiver.m to read alternative toolTip format

gcasa:
Fix for issue#192.  Read alternate toolTip format.

### DIFF
--- a/Source/GSXib5KeyedUnarchiver.m
+++ b/Source/GSXib5KeyedUnarchiver.m
@@ -3186,6 +3186,20 @@ didStartElement: (NSString*)elementName
       else if ([object respondsToSelector: @selector(setHeaderToolTip:)])
         [object setHeaderToolTip: [element attributeForKey: @"toolTip"]];
     }
+  else if ([element elementForKey: @"toolTip"])
+    {
+      GSXibElement *toolTipElement = [element elementForKey: @"toolTip"];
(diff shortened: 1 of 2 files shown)


Source/GSXib5KeyedUnarchiver.m
```diff
@@ -3186,6 +3186,20 @@ - (id) decodeObjectForXib: (GSXibElement*)element
       else if ([object respondsToSelector: @selector(setHeaderToolTip:)])
         [object setHeaderToolTip: [element attributeForKey: @"toolTip"]];
     }
+  else if ([element elementForKey: @"toolTip"])
+    {
+      GSXibElement *toolTipElement = [element elementForKey: @"toolTip"];
+      NSString *v = [toolTipElement value];
+      if ([object respondsToSelector: @selector(setToolTip:)])
```
fredkiefer, line 3193: We should try to share some code with the lines above by extracting the whole `if` block into a new method.

gcasa: Good point. I will do that asap.

gcasa: I have done this, per your suggestion.  Resolve this comment, if you feel as though it fits what you wanted.

PR #195, opened 2023-08-31
Add dummy awakeFromNib dummy method

gcasa:
This adds a  dummy method for NSView so that subclasses can call awakeFromNib.  I have run into production code which does this and does not fail on macOS.  I also have a test which I will add to git and reference here.

### DIFF
--- a/Source/NSView.m
+++ b/Source/NSView.m
@@ -5231,6 +5231,10 @@ static NSView* findByTag(NSView *view, NSInteger aTag, NSUInteger *level)
   return 0;
 }
 
+- (void) awakeFromNib
+{
+}
+
 @end
 
 #if OS_API_VERSION(MAC_OS_X_VERSION_10_7, GS_API_LATEST)


gcasa:
Here is the test... https://github.com/gcasa/NSView_subclass_awakeFromNib_test

fredkiefer:
This new method will result in a lot of useless calls but if Apple has it, we will need it too.

gcasa:
@fredkiefer One additional suggestion I have on this... even though it is merged.   You lament that we are "making a lot of useless calls" yet it appears to me that in order to avoid making said useless calls we are invoking "respondsToSelector" in an if and THEN making the call instead of just assuming the method is always there, so, for the calls we don't have we are doing an if AND a respondsToSelector call anyway and for those we do have we are doing both of those AND the call anyway.

I am wondering if it is more or less expensive to simply assume it's presence rather then check for it, but that's a discussion for another time.

fredkiefer:
Sounds like a valid solution to me. In that case we should move this method to NSObject.

gcasa:
I would guess/bet that this method is implemented on NSObject via a category in AppKit.   I am going to do some research today to see if that's the case.

Issue #196, opened 2023-09-01
NSMenu implement removeAllItems


gcasa:
Currently there is no version of this method in our codebase and it is needed for compatibility.

gcasa:
Merged fix

PR #197, opened 2023-09-01
Fixes issue #196: Add implementation of removeAllItems to NSMenu

gcasa:
Adds a simple implementation of this method since it is missing in the existing codebase.


fredkiefer:
This should do. We could think about switching off notifications before the loop and activating it after the loop again. But this shouldn't#t be much of a difference.

fredkiefer (approved):

PR #199, opened 2023-09-12
Warning Fix

rmottola:
catched some - not all, but catched some - done with Greg.

### DIFF
--- a/Source/GNUmakefile
+++ b/Source/GNUmakefile
@@ -684,19 +684,6 @@ GSXibElement.h \
 GSXibLoading.h \
 GSXibKeyedUnarchiver.h \
 GSHelpAttachment.h \
-GSAutoLayoutVFLParser.h \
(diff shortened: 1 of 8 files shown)


Source/GNUmakefile
```diff
@@ -684,19 +684,6 @@ GSXibElement.h \
 GSXibLoading.h \
 GSXibKeyedUnarchiver.h \
 GSHelpAttachment.h \
```
fredkiefer, line 686: You should remove the trailing backslash as well.

fredkiefer (approved):
Part from this minor issue, which you should fix anyway, this looks great to me.

Source/GNUmakefile
rmottola: it is convenient to add later, but yes, cleaner. Removed.

PR #201, opened 2023-09-20
add gstheme hooks

KyleCardoza:
This PR includes several very small changes to a few classes, all of which give GSTheme more ability to control the visual presentation of the UI. Each GSTheme hook is accompanied by a default implementation in GSTheme itself which simply does whatever the library was going to do already; the net effect is that these changes change nothing for existing apps and themes, but allow new themes more control.


Source/NSMenu.m
```diff
@@ -280,6 +282,12 @@ - (void) _organizeMenu
       appItem = [self itemWithTitle: appTitle];
       appMenu = [appItem submenu];
 
+      if ([[GSTheme theme] organizesMainMenu])
```
fredkiefer, line 285: This can get rather confusing. now the theme may rearrange the menu and afterwords the code here will do something different with it. Wouldn't it be simpler to move all the code from here into the default implementation in the theme and allow other themes to override it?

Source/NSMenuItemCell.m
```diff
@@ -290,6 +290,8 @@ - (NSString*) _keyEquivalentString
                           (shift != NO) ? shiftKeyString : @"",
                           (m & NSCommandKeyMask) ? commandKeyString : @"",
                           key];
+
+	  key = [[GSTheme theme] keyForKeyEquivalent: key];
```
fredkiefer, line 294: Again I am wondering whether we should move more of the original code into the default theme. As the code is now, the theme must be fully aware on how we build up the key here, parse that and rewrite it.
The only thing that speaks against that move is that this code here also changes the key equivalent font. So it is not side effect free.

Source/NSMenuView.m
```diff
@@ -791,6 +791,9 @@ - (void) sizeToFit
           NSMenuItemCell *aCell = [self menuItemCellForItemAtIndex: i];
           float titleWidth = [aCell titleWidth];
```
fredkiefer, line 792: This should be an CGFloat and that type should also be used in your new method.

fredkiefer (changes requested):
Looks good to me. There are a few places where more discussion is needed but the general direction is great.

Source/GSThemeDrawing.m
```diff
@@ -2073,6 +2072,38 @@ - (void) drawWindowBorder: (NSRect)rect
...
+  NSSize buttonSize = [[closeButton image] size];
+  buttonSize = NSMakeSize(buttonSize.width + 3, buttonSize.height + 3);
+  
+  [closeButton setFrame: NSMakeRect(viewSize.width - buttonSize.width - 4,
+                   		   (viewSize.height - buttonSize.height) / 2,
+  		                    buttonSize.width, 
+				    buttonSize.height)];
+}
+
+- (NSRect) minimizeButtonFrameForBounds: (NSRect)bounds
```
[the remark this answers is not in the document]
KyleCardoza: Right, my bad. Consider that one fixed.

Source/NSMenu.m
```diff
@@ -265,6 +265,8 @@ - (void) _organizeMenu
   NSString *servicesString = _(@"Services");
   int i;
 
+
```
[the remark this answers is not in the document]
KyleCardoza: Removed, must have been a slip of the hands.

Source/NSMenu.m
KyleCardoza: You know, I was thinking about that -- there isn't any really good reason for that not to be the case. The whole point of GSTheme, as I understand it, is to allow overriding default behaviour without breaking existing code -- so by moving the default behaviour of things into the base GSTheme class, we would accomplish that end with more flexibility for theme authors and no cost to those who aren't interested in fancy new themes.

So anyway, I'll get that change done too.

Source/NSMenuItemCell.m
KyleCardoza: I think you're right, as much as possible of the "default" drawing behaviour should be moved into the default GSTheme class, both for architectural reasons, and because it affords theme authors far 
more power and flexibility in customizing the visual presentation.

So perhaps the implementation of `_keyEquivalentFont` and `_setKeyEquivalentFont` should defer to the theme, as well as replacing my implementation of `GSTheme-[keyForKeyEquivalent]` with the existing body of `NSMenuItemCell-[_keyEquivalentString]`, and have that method itself defer to the matching theme method?

Source/NSMenuView.m
KyleCardoza: Got it, will fix.

rmottola:
I have tested this with the standard theme and with the WinClassic theme which sets a horizontal menu and I see no issue.

KyleCardoza:
Excellent to hear. Is there anything else I should work on before marking it ready for final review?

rmottola:
> Excellent to hear. Is there anything else I should work on before marking it ready for final review?

that's up to @fredkiefer - I was just adding some additional tests & checks.

Source/GSThemeMenu.m
```diff
@@ -175,5 +184,239 @@ - (BOOL) menuShouldShowIcon
...
+              if (index != 0)
+                {
+                  RETAIN (appItem);
+                  [menu removeItemAtIndex: index];
+                  [menu insertItem: appItem atIndex: 0];
+                  RELEASE (appItem);
+                }
+            }
+
+	  if ([[GSTheme theme] menuShouldShowIcon])
```
fredkiefer, line 263: `[GSTheme theme]` should just be `self` here.

Source/GSThemeDrawing.m
```diff
@@ -2073,6 +2072,38 @@ - (void) drawWindowBorder: (NSRect)rect
     }
 }
 
+- (void) setFrameForCloseButton: (NSButton *)closeButton
```
fredkiefer, line 2075: Not your fault, but we shouldn't need this second method to determine the close button size.

Source/NSMenuItemCell.m
fredkiefer: I fully agree with your comment, but your code does not reflect this.

fredkiefer (approved):

Source/NSMenuItemCell.m
KyleCardoza: I was waiting for feedback on the idea before implementing it -- but I'm glad we're on the same page there. I will get that done, probably in my next PR, which I plan to be specifically about moving more default behaviour into the base GSTheme class, as opposed to the minimal set of hooks I needed for my own personal passion project.

gcasa:
Hey @KyleCardoza if you deem this ready you are okay to merge it.   If you can't then either @fredkiefer or I can do it.  I may have to add you to the project to allow you to have permission.   I will do that now.

KyleCardoza:
I didn't think I could do that, but if that's how we want to do it I am happy and proud to accept.

PR #204, opened 2023-10-01
Allow popover to be dismissed

gcasa:


Source/NSPopover.m
```diff
@@ -232,23 +232,27 @@ - (void) close
 {
   [_realPanel close];
   [_realPanel setDelegate:nil];
+  [_realPanel orderOut: nil];
```
fredkiefer, line 235: Why is the `orderOut:` needed here and in the next method? The close Methods should already take care of that.

Source/NSViewController.m
```diff
@@ -164,6 +186,45 @@ - (NSBundle *)nibBundle
...
+  [viewController viewDidDisappear: NO];
+}
+
+- (void) dismissController: (id)sender
+{
+  [self dismissViewController: self];
+}
+
+// NSSeguePerforming methods...
+- (void)performSegueWithIdentifier: (NSStoryboardSegueIdentifier)identifier
```
fredkiefer, line 201: Here and in the next methods a space after the return type is missing.

gcasa: Fixed, resolve if you agree.

Source/NSPopover.m
gcasa: This should now be addressed.

gcasa:
I'm fixing some positioning issues with the popover now.

fredkiefer (approved):

gcasa:
I am opening another PR for the positioning issue as well as the segue problems we discussed a while ago.

PR #205, opened 2023-10-01
Avoid accessing instance variables in inline functions when compiling with msvc

qmfrederik:
`DPSOperators.h` exports various inline functions which access the `methods` variable of the `NSGraphicsContext` class.  This will result in accessing instance variables across module boundaries if these functions are used in external libraries.

Accessing instance variables across module boundaries is not supported when building on Windows using the Visual Studio toolchain.

This patch adds a `methods` function on `NSGraphicsContext` which returns the variable, and reworks the inline functions to use the `methods` function instead of directly accessing the variable.

Re-submit of !177


fredkiefer:
As this was reviewed in detail in the last pull request I just trust you that this are the same changes. Hopefully we have more luck this time.

PR #210, opened 2023-11-25
Move awakeFromNib implementation to NSObject instead of NSView, as di…

gcasa:
As discussed in the documentation

### DIFF
--- a/Headers/AppKit/NSNibLoading.h
+++ b/Headers/AppKit/NSNibLoading.h
@@ -47,6 +47,15 @@
  */
 - (void) awakeFromNib;
 
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_8, GS_API_LATEST)
+/**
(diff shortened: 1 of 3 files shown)


gcasa:
https://github.com/gnustep/libs-base/issues/352

gcasa:
https://github.com/gnustep/libs-gui/issues/211

gcasa:
https://developer.apple.com/documentation/objectivec/nsobject/1402907-awakefromnib?language=objc

fredkiefer:
The change itself is OK, but now we also could remove all the checks whether an objects responds to `awakeFromNib`

fredkiefer (approved):

gcasa:
> The change itself is OK, but now we also could remove all the checks whether an objects responds to `awakeFromNib`

I didn't see this comment. I will do this in a separate PR.

PR #212, opened 2023-11-27
Update calls to awakeFromNib

gcasa:
and add call to prepareForInterfaceBuilder if we are in Gorm.  This is per our discussion @fredkiefer.

### DIFF
--- a/Source/GSGormLoading.m
+++ b/Source/GSGormLoading.m
@@ -6,7 +6,7 @@
 
    Author: Gregory John Casamento
    Date: July 2003.
-   
+
    This file is part of the GNUstep GUI Library.
 
    This library is free software; you can redistribute it and/or
@@ -21,10 +21,10 @@
 
    You should have received a copy of the GNU Lesser General Public
    License along with this library; see the file COPYING.LIB.
-   If not, see <http://www.gnu.org/licenses/> or write to the 
-   Free Software Foundation, 51 Franklin Street, Fifth Floor, 
+   If not, see <http://www.gnu.org/licenses/> or write to the
+   Free Software Foundation, 51 Franklin Street, Fifth Floor,
    Boston, MA 02110-1301, USA.
-*/ 
+*/
 
 #import <Foundation/NSCoder.h>
 #import <Foundation/NSDictionary.h>
@@ -55,7 +55,7 @@ static const int currentVersion = 1; // GSNibItem version number...
 @end
 
 /*
- * This private class is used to collect the nib items while the 
+ * This private class is used to collect the nib items while the
  * .gorm file is being unarchived.  This is done to allow only
  * the top level items to be retained in a clean way.  The reason it's
  * being done this way is because old .gorm files don't have any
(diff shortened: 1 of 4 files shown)


gcasa:
The whitespace changes are because I did whitespace-cleanup before committing.

Source/GSNibLoading.m
```diff
@@ -2005,22 +2004,33 @@ proxy objects (-1 = File's Owner, -2 = First Responder,
...
+	  // If we are in Gorm/IB, set the object up for display
+	  if ([NSClassSwapper isInInterfaceBuilder] == YES)
+	    {
+	      [obj prepareForInterfaceBuilder];
+	    }
+	}
     }
 
   // awaken the owner
   if ([owner respondsToSelector: @selector(awakeFromNib)])
```
fredkiefer, line 2025: Maybe this is also no longer needed?

Source/GSNibLoading.m
```diff
@@ -2005,22 +2004,33 @@ proxy objects (-1 = File's Owner, -2 = First Responder,
...
     {
       [owner awakeFromNib];
+
+      // If we are in Gorm/IB, set the object up for display
+      if ([NSClassSwapper isInInterfaceBuilder] == YES)
+	{
+	  [obj prepareForInterfaceBuilder];
```
fredkiefer, line 2032: I think that `obj` will be undefined here. Did you mean `owner`? Not sure whether that is needed.

Source/GSGormLoading.m
```diff
@@ -223,46 +223,55 @@ - (void) awakeWithContext: (NSDictionary *)context
...
-		  if ([o respondsToSelector: @selector(awakeFromNib)])
+		  // send the awake message, every object should respond to this message since
+		  // it is defined on NSObject(NSNibAwaking)
+		  [o awakeFromNib];
+
+		  // Call prepareForInteraceBuilder if we are in IB/Gorm.
+		  if ([self respondsToSelector: @selector(isInInterfaceBuilder)])
 		    {
-		      [o awakeFromNib];
+		      if ([self isInInterfaceBuilder] == YES)
```
fredkiefer, line 254: In the NIB and XIB loading code we make this call on NSClassSwapper, why is it on self here?

fredkiefer (changes requested):
Looks good but there seem to be a few issues

Source/GSNibLoading.m
gcasa: The call to awakeFromNib for the owner is because it's not one of the objects I archived but it still needs to have this call done.  The check is no longer needed for respondsToSelector:.

Source/GSNibLoading.m
gcasa: I changed this to owner.

Source/GSNibLoading.m
gcasa: I have updated this to remove the check.

Source/GSGormLoading.m
gcasa: This is because we do things slightly differently with .gorm files since they were created first.  I learned some lessons and did it a bit more correct in Nib (and that was inherited by Xib since they are related).

gcasa:
{;ease review new changes and resolve if you agree.

Source/GSGormLoading.m
fredkiefer: Don't you think it would be about time to change this? But maybe not in this MR.

fredkiefer (approved):

PR #213, opened 2023-11-27
Add initial test for NSStackView, also adding tests for other new classes...

gcasa:
I am going to incorporate many of the independent tests into GUI's test suite as well as add some additional logic to check that things are being positioned / computed correctly.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,12 @@
+2023-11-27 Gregory John Casamento <greg.casamento@gmail.com>
+
+	* Source/GSGormLoading.m
+	* Source/GSNibLoading.m
+	* Source/GSXibLoading.m: Remove calls to respondsToSelector:
+	for awakeFromNib.  This method is now implemented on a
+	category on NSObject called NSNibAwaking.  Also add a
+	call to prepareForInterfaceBuilder when we are in Gorm/IB.
+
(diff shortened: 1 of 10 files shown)


gcasa:
These are simple tests that only test the loading of these new classes.   They are a start for more in depth tests.

fredkiefer:
With this sort of tests I am a bit unsure where we should put them. The tests in the libs-gui/Test directory are intended to be run from the CI pipeline. And these new tests don't make much sense in this respect. Maybe we could put them into the examples project? On the other hand nobody ever executes the tests there. Maybe a new directory in the libs-gui project would do? As I said, I am unsure here and maybe we should start a discussion on the mailing list?

gcasa:
> With this sort of tests I am a bit unsure where we should put them. The tests in the libs-gui/Test directory are intended to be run from the CI pipeline. And these new tests don't make much sense in this respect. Maybe we could put them into the examples project? On the other hand nobody ever executes the tests there. Maybe a new directory in the libs-gui project would do? As I said, I am unsure here and maybe we should start a discussion on the mailing list?

I think that moving them to examples would be sufficient.  I really would like more tests in gui but I am not entirely sure what kind of tests we can add as gui is harder to test.

gcasa:
Per discussion i am going to move these to the examples directory.

gcasa:
I'm closing this.

PR #216, opened 2023-12-16
Fix handling of custom classes for XIB v5.  This corrects an issue in…

gcasa:
This fixes an issue in Gorm as well.   Previously custom classes were not being handled properly and causing a crash when loading a xib and saving as a gorm.   This fix helps to avert that issue.


fredkiefer (approved):
I don't see how moving code would break anything.

gcasa:
> I don't see how moving code would break anything.

Thanks, Fred. 

I added code to initialize the dictionary.  I was considering doing it as a quick fix but I had already created the PR.

Issue #217, opened 2023-12-21
NSShadow does not work


optimisme:
Here is an example:
```objectivec
    NSRect shadowFrame = NSInsetRect(frame, 5.0, 5.0);
    shadowFrame.origin.x += 5.0;
    shadowFrame.origin.y += 5.0;

    // Definir el color de l'ombra
    NSColor *shadowColor = [NSColor redColor];

    // Definir l'ombra
    NSShadow *shadow = [[NSShadow alloc] init];
    [shadow setShadowColor:shadowColor];
    [shadow setShadowOffset:NSMakeSize(0, -1)];
    [shadow setShadowBlurRadius:10]; 

    // Aplicar l'ombra
    [NSGraphicsContext saveGraphicsState];
    [shadow set];

    // Dibuixar l'ombra (Això dibuixa una "caixa" d'ombra al voltant de l'àrea del botó)
    NSBezierPath *shadowPath = [NSBezierPath bezierPathWithRect:shadowFrame];
    [[NSColor clearColor] set];
    [shadowPath fill];
    // Restaurar l'estat gràfic per a que la configuració de l'ombra no afecti el fons del botó
    [NSGraphicsContext restoreGraphicsState];
```
   
   Here another example that doesn't do anything:
```objectivec
       [NSGraphicsContext saveGraphicsState];
    
    // Definir el color de l'ombra
      NSColor *shadowColor = [NSColor blackColor];

      // Crear l'ombra
      NSShadow *shadow = [[NSShadow alloc] init];
      [shadow setShadowColor:shadowColor];
      [shadow setShadowOffset:NSMakeSize(0, 1)]; // Desplaçament de l'ombra 1 píxel cap avall
      [shadow setShadowBlurRadius:20]; // Radi de difuminació de 20 píxels

      // Aplicar l'ombra al botó
      [buttonDefault setShadow:shadow];

      // Restaurar el context gràfic
      [NSGraphicsContext restoreGraphicsState];
```

fredkiefer:
Yes, at the moment NSShadow is missing most of the real drawing code. If I remember correctly, it would be really hard to implement this as it requires a different drawing concept from the one we have implemented in GNUstep gui at the moment. Any patchers to improve this are welcome.

optimisme:
In my opinion, this is not an improvement, it is a bug. A very serious bug. I don't know who will fix it, but without a doubt, GNUStep cannot be considered serious if this bug is not fixed.

gcasa:
[...]

PR #229, opened 2023-12-24
Update NSView with setShadow:/shadow methods

gcasa:
Please see the bug here: https://github.com/gnustep/libs-gui/issues/217

### DIFF
--- a/Source/GSXib5KeyedUnarchiver.m
+++ b/Source/GSXib5KeyedUnarchiver.m
@@ -325,6 +325,11 @@ static NSArray      *XmlBoolDefaultYes  = nil;
                                            @"middleViews", @"NSStackViewMiddleContainer",
                                            @"endViews", @"NSStackViewEndContainer",
 					   @"collectionViewLayout", @"NSCollectionViewLayout",
+					   @"shadow", @"NSViewShadow",
+					   @"blurRadius", @"NSShadowBlurRadius",
+					   @"width", @"NSShadowHoriz",
+					   @"height", @"NSShadowVert",
+					   @"color", @"NSShadowColor",
                                            nil];
           RETAIN(XmlKeyMapTable);
 
@@ -870,12 +875,12 @@ didStartElement: (NSString*)elementName
               // Need to store element for making the connections...
               [self addConnection: element];
             }
-	  /*
+          /*
           else if ([XmlConstraintRecordTags containsObject: elementName])
             {
               [self objectForXib: element]; // decode the constraint...
             }
-	  */
+          */
         }
       else
         {
@@ -3186,13 +3191,13 @@ didStartElement: (NSString*)elementName
   if (toolTipString != nil)
     {
       if ([object respondsToSelector: @selector(setToolTip:)])
-	{
-	  [object setToolTip: toolTipString];
-	}
+        {
+          [object setToolTip: toolTipString];
+        }
       else if ([object respondsToSelector: @selector(setHeaderToolTip:)])
-	{
-	  [object setHeaderToolTip: toolTipString];
-	}
+        {
+          [object setHeaderToolTip: toolTipString];
+        }
     }
 }
 
--- a/Source/GSXib5KeyedUnarchiver.m
+++ b/Source/GSXib5KeyedUnarchiver.m
@@ -448,8 +448,8 @@ static NSArray      *XmlBoolDefaultYes  = nil;
                @"decodeSecondAttribute:", @"NSSecondAttribute",
                @"decodeRelation:", @"NSRelation",
                @"decodeTransitionStyle:", @"NSTransitionStyle",
-	    @"decodeShadowOffsetHoriz:", @"NSShadowHoriz",
-	    @"decodeShadowOffsetVert:", @"NSShadowVert",
+	       @"decodeShadowOffsetHoriz:", @"NSShadowHoriz",
+	       @"decodeShadowOffsetVert:", @"NSShadowVert",
                  nil];
           RETAIN(XmlKeyToDecoderSelectorMap);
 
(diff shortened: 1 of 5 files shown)


gcasa:
I think that this might give some clue as to how to do this in the backend.  I am not sure that shadows necessarily need layers to work.
https://stackoverflow.com/questions/30630280/how-do-i-create-a-circular-shadow-effect-with-cairo

gcasa:
Also of interest MIGHT be... https://github.com/rcairo/rcairo/blob/master/lib/cairo/context/blur.rb

I know these examples are in different languages, but they are both using cairo, so I wanted to note them here.

gcasa:
It seems as though the keys being used are these...
```
[heron@nomad Base.lproj] % strings MainMenu.nib | grep -i shadow                                     [0]
NSShadowHoriz
IBNSShadowedSymbolConfiguration
NSShadowVert
NSViewShadow
NSShadowBlurRadius
NSShadowColor
NSShadow_Test
NSShadow_Test
Quit NSShadow_Test
Quit NSShadow_Test
About NSShadow_Test
About NSShadow_Test
NSShadow_Test Help
NSShadow_Test Help
Hide NSShadow_Test
Hide NSShadow_Test
NSShadow_Test
NSShadow_Test
NSShadow
```
The new NIBArchive format is proving to be a menace at analyzing nib files for keys... but since we can read xib files it is not a problem, but it is a format we might want to consider looking at in the future. GC

gcasa:
Yet more info about blur... https://discourse.gnome.org/t/cairo-gaussian-blur-for-shadows/2174

Source/NSView.m
```diff
@@ -4634,7 +4646,7 @@ - (id) initWithCoder: (NSCoder*)aDecoder
   NSEnumerator *e;
   NSView	*sub;
   NSArray	*subs;
-
+
```
fredkiefer, line 4649: Here is some leftover space

gcasa: Should be fixed.

gcasa:
Okay, so now I have got everything decoding for the NSShadow from the XIB properly.   I am going to work on the backend now.   Some changes might be needed in libs-gui, but I doubt it.

fredkiefer:
Let me try to explain this once more:
- You should start off by implementing direct drawing of the shadow. For this you will need to add a setShadow: method on the GSGraphicsContext class to be called from `[NSShadow set]`. After that you just need to implement that method or rather the actual handling of the shadow while drawing in the Cairo graphics context in the backend.
- Implementing the shadow drawing for a view requires more changes in gui. When drawing a view we always clip to the frame of the view. This means, when drawing the shadow from the view drawing code, it would never be visible. Most likely we need to draw it from the super view context somewhere in the whole view hierarchy drawing code. But this is only part of the problem. You also need to figure out what parts of the screen need to be invalidated when a view with a shadow requires a redraw. Please take your time here. A complete day just trying to understand the existing code would be well spend. Then figure out what extensions will be required and before implementing them, let us discuss your approach.

I would suggest to merge this PR here separately and not wait until you have the complete implementation in place.

gcasa:
Okay.  Thank you for the great input.  I will change this PR to ready and if you find it suitable, then please review.

gcasa:
I will open another PR for the shadow work in NSGraphicsContext.

gcasa:
@fredkiefer please approve if it's okay to merge.  I am in the process of looking at what is involved on the backend.

gcasa:
https://github.com/gnustep/libs-gui/pull/232 PR for set method on libs-gui

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -445,6 +448,8 @@ + (void) initialize
                @"decodeSecondAttribute:", @"NSSecondAttribute",
                @"decodeRelation:", @"NSRelation",
                @"decodeTransitionStyle:", @"NSTransitionStyle",
+	    @"decodeShadowOffsetHoriz:", @"NSShadowHoriz",
```
fredkiefer, line 451: Looks like the indentation is incorrect

Source/GSXib5KeyedUnarchiver.m
```diff
@@ -2984,6 +2989,20 @@ - (id) decodeTransitionStyle: (GSXibElement *)element
   return num;  
 }
 
+- (id) decodeShadowOffsetHoriz: (GSXibElement *)element
+{
+  NSSize size = [self decodeSizeForKey: @"offset"];
+  NSNumber *num = [NSNumber numberWithFloat: size.width];
```
fredkiefer, line 2995: This local variable is not needed.

Source/NSView.m
```diff
@@ -4737,12 +4749,19 @@ - (id) initWithCoder: (NSCoder*)aDecoder
 	  [self didAddSubview: sub];
 	}
 
+      // Decode the shadow...
+      if ([aDecoder containsValueForKey: @"NSViewShadow"])
+	{
+	  _shadow = [aDecoder decodeObjectForKey: @"NSViewShadow"];
```
fredkiefer, line 4755: Here a RETAIN would be required.

Source/GSXib5KeyedUnarchiver.m
gcasa: Removed

Source/GSXib5KeyedUnarchiver.m
gcasa: Corrected

Source/NSView.m
```diff
@@ -4798,6 +4817,12 @@ - (id) initWithCoder: (NSCoder*)aDecoder
 	  [self didAddSubview: sub];
 	}
       RELEASE(subs);
+
+      // Decode the shadow if this is version 2 or greater...
+      if (version >= 2)
+	{
+	  _shadow = [aDecoder decodeObject];
```
fredkiefer, line 4824: Again you probably should retain the decoded object as you release it in the dealloc method.

Source/NSView.m
gcasa: Done

Source/NSView.m
gcasa: Done.

fredkiefer (changes requested):
Looks mostly OK, but the RETAIN issue needs to be resolved. You could also use the setter instead.

fredkiefer (approved):

gcasa:
> Looks mostly OK, but the RETAIN issue needs to be resolved. You could also use the setter instead.

I added both retains, I think using the setter is cleaner.  I am going to go ahead and make that change.  Give me a moment.

gcasa:
I changed the code to use the setter, thanks. :). Merging.

gcasa:
Forgot to add the NSView.m file to the commit... I will correct it on master. Sorry.

gcasa:
Oh... it got them.  :). Nevermind.

PR #242, opened 2024-02-06
Make PACKAGE_SCOPE @public on MinGW

qmfrederik:
The clang compiler actually enforces visibility on MinGW (by not exporting the instance variable offsets); there isn't a functional equivalent of `@package` on Windows.

Without this patch, compiling libs-gui fails like this:

```
Linking bundle libgmodel ...
ld.lld: error: undefined symbol: __objc_ivar_offset_NSView._sub_views.☺
>>> referenced by ./obj/libgmodel.obj/GMAppKit.m.o:(.refptr.__objc_ivar_offset_NSView._sub_views.☺)
```

### DIFF
--- a/Headers/AppKit/AppKitDefines.h
+++ b/Headers/AppKit/AppKitDefines.h
@@ -72,7 +72,7 @@
 
 #endif
 
-#ifdef __clang__
+#if defined(__clang__) && !defined(__MINGW32__)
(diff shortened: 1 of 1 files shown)


fredkiefer (approved):
As long as we agree that this is a bug of clang on MINGW this change is fine for me.

PR #247, opened 2024-03-31
Remove unused variables

qmfrederik:
Suppresses a couple of build warnings (-Wunused-but-set-variable)

### DIFF
--- a/Source/GSToolbarCustomizationPalette.m
+++ b/Source/GSToolbarCustomizationPalette.m
@@ -100,7 +100,6 @@
   float vAccumulator = 0.0;
   NSEnumerator *e = [[self subviews] objectEnumerator];
   NSView *layoutedView = nil;
-  int index = 0;
 
   // Loop over all subviews
   while ((layoutedView = [e nextObject]) != nil)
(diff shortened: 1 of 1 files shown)


Source/GSHorizontalTypesetter.m
```diff
@@ -1035,9 +1034,9 @@ the first tab would move (exactly) to the next tab
 		  g->nominal = NO;
 		p.x += last_p.x;
 		p.y += last_p.y;
+	    last_p = g->pos = p;
 	      }*/
 
-	    last_p = g->pos = p;
```
fredkiefer, line 1040: I am not sure whether the setting of `g->pos` that got removed here is really not needed. If so, could you please explain why?

qmfrederik: You are right, this was too greedy on my side.

qmfrederik:
@fredkiefer Thanks for reviewing this.  I reduced the scope of this patch to focus on the first file only.  You are probably right and the proper fix is to re-enable the commented out code, but that's a bit beyond the mindset of "let's see if there's low hanging fruit to reduce compiler warnings" I had when I opened these PRs.

fredkiefer (approved):

PR #251, opened 2024-04-03
GSXib5KeyedUnarchiver: Add mapping for NSSegmentItemLabel, NSSegmentItemImage

qmfrederik:
Also fixes some whitespace issues (replace tabs by spaces to be consistent)


fredkiefer:
The change itself looks good, but what is causing the checks to fail and why are there four extra lines of change?

qmfrederik:
The failed checks are, I think, a GitHub system failure; the four extra changes were tabs-to-spaces conversion (to be consistent with the other lines)

PR #252, opened 2024-04-03
[NSTableView initWithCoder]: Read alternatingRowBackgroundColors, from tableViewFlags

qmfrederik:


fredkiefer (changes requested):
Looks good, but shouldn't#t we encode these new values as well? Otherwise a roundtrip won't work. (Line 5508)

qmfrederik:
Thanks, @fredkiefer . I've added the encoding of those values in 7b29a2fd355371b8ac7be4dc2261ea8aa1317c79.

PR #254, opened 2024-04-03
NSBox: (re-)implement isOpaque; decode NSBorderColor2 and NSFillColor2

qmfrederik:
This (re-)implements logic where a box is marked as opaque if it is a custom box which is not transparent and adds the decoding of the `NSBorderColor2` and `NSFillColor2` attributes.

This fixes the rendering of custom boxes with a custom fill color; the current implementation would mark them as non-opaque and hence the fill color would not be rendered.

- The original implementation (from 1999) always returned `YES` (53bcb5024020fe85e370d9d49297aca4320d7768)
- A more complete implementation was backported from the Testplant branch in 2012 (e85b16bc05d5d513788c09b470fd09ad5f888882, by @fredkiefer, downstream commit 36e77b95f7921e9539fdc0c115fe4039fa573150)
- In 2013, a return value of `NO` was hardcoded (02bc49e2d5336e9092001e3d56cb43750a8466a3, by @ericwa)

### DIFF
--- a/Source/NSBox.m
+++ b/Source/NSBox.m
@@ -464,18 +464,7 @@
 
 - (BOOL) isOpaque
 {
-  // FIXME: Depends on theme; if always returning NO is a performance hit
-  // we can check if GSTheme is going to draw an old-style opaque box
-  // or not.
-  return NO;
-  // if (_box_type == NSBoxCustom)
-  //   {
-  //     return !_transparent;
-  //   }
-  // else
-  //   {
-  //     return YES;
-  //   }
+  return [[GSTheme theme] isBoxOpaque: self];
 }
 
 - (NSColor*) fillColor
--- a/Headers/Additions/GNUstepGUI/GSTheme.h
+++ b/Headers/Additions/GNUstepGUI/GSTheme.h
@@ -1339,6 +1339,8 @@ APPKIT_EXPORT_CLASS
 		 clipRect: (NSRect)clipRect
 		   inView: (NSView *)view;
 
+- (BOOL) isBoxOpaque: (NSBox *)box;
+
 - (void) drawBoxInClipRect: (NSRect)clipRect
 		   boxType: (NSBoxType)boxType
 		borderType: (NSBorderType)borderType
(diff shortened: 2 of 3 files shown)


Source/NSBox.m
```diff
@@ -629,6 +628,23 @@ - (id) initWithCoder: (NSCoder*)aDecoder
...
+          NSColor *color = [aDecoder decodeObjectForKey:@"NSBorderColor2"];
+          if (color != nil)
+            [self setBorderColor:color];
+        }
+
+      [self setFillColor:[NSColor clearColor]];
+      if ([aDecoder containsValueForKey:@"NSFillColor2"])
+        {
+          NSColor *color = [aDecoder decodeObjectForKey:@"NSFillColor2"];
+          NSLog(@"Got box fill color %@", [color description]);
```
fredkiefer, line 644: This seems to be a left over from debugging.

Source/NSBox.m
```diff
@@ -629,6 +628,23 @@ - (id) initWithCoder: (NSCoder*)aDecoder
           [self setContentView: cv];
           RELEASE(cv);
         }
+
+      [self setBorderColor:[NSColor clearColor]];
```
fredkiefer, line 632: Here and in the following lines the space after the colon is missing.

fredkiefer:
The decoding of the missing values is fine. What I am worried about is the opaque method. The code here really cannot know what the theme is going to do with the box. There all the set colours could get ignored. A proper solution would be to move your code (which is the old code) into the theme and have themes that behave differently override that method.

qmfrederik:
> A proper solution would be to move your code (which is the old code) into the theme and have themes that behave differently override that method.

I don't know much about themes in GNUstep, so where would that code go?  In `GSThemeDrawing drawBoxInClipRect` or somewhere else?

fredkiefer:
What would be needed is a new method `isBoxOpaque:` close to the other one you mentioned. The code being just like the one you tried to put here. And the method in NSBox would call the theme method with itself as an argument. Annoying and complicated, but the only way the behaviour matches.

fredkiefer (approved):

PR #256, opened 2024-04-05
Fix NSImageView intercepting mouse events when not editable

williameveretteggplant:
This adds a condition to the mouseDown: event for NSImageView. Basically, if the image view is not editable, it invokes the mouseDown: method on the nextResponder for the ImageView.

This is necessary, because as it currently works, the imageView will intercept any mouse clicks on the image view, even though it is not editable and should be transparent to mouse events. This means that if the image view is a subview of another view that you want to click on, such as an NSCollectionViewItem, the clicks will not register if you click on the image view.

One can also make the imageView not respond to mouse events by setting enabled to false, but that has side effects such as changing the appearance of the image.

### DIFF
--- a/Source/NSImageView.m
+++ b/Source/NSImageView.m
@@ -308,6 +308,12 @@ static Class imageCellClass;
 
 - (void) mouseDown: (NSEvent*)theEvent
 {
+  if (![self isEditable])
+    {
+      [[self nextResponder] mouseDown:theEvent];
+      return;
+    }
+
   if ([self initiatesDrag])
     {
       NSPasteboard *pboard;
--- a/Source/NSImageView.m
+++ b/Source/NSImageView.m
@@ -308,11 +308,6 @@ static Class imageCellClass;
 
 - (void) mouseDown: (NSEvent*)theEvent
 {
-  if (![self isEditable])
-    {
-      [[self nextResponder] mouseDown:theEvent];
-      return;
-    }
 
   if ([self initiatesDrag])
     {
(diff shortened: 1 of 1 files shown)


fredkiefer:
I am unsure about this change. On first sight it looks like a hack just to avoid the super call that would happen at the end of the method and actually it is just that. So either we are doing something wrong in NSControl or we have a very special case here. Some analysis what Apple is doing here would really help to clear up things. I see how this change would fix the behaviour for you tough.

williameveretteggplant:
Well the current implementation of NSTextField on master has this:

>   if ([self isSelectable] == NO || [self isEnabled] == NO)
    {
      if (_next_responder)
        [_next_responder mouseDown: theEvent];
      else
        [super mouseDown: theEvent];
      return;
    }
So maybe we should copy that pattern. 
Basically, NSControl doesn't use the 'editable' property, so it has no way to determine if "editable" status of the cell should be used. It only checks enabled, which is a property on NSControl. If it is enabled it is going to capture that mouse event and it can't move on to the next responder.
I can tell you that in the Apple approach it works correctly, which is to say the ImageView does not block mouse events when it is not editable. As to exactly how they implement that, I have no way of knowing.

fredkiefer:
This sounds like a plan. We just replace the call to super at the end of the method with something similar you what you copied over from NSTextField.

williameveretteggplant:
Ok. Change made.

fredkiefer (approved):
Let's try with this version.

PR #257, opened 2024-04-09
Set editable property in NSCell default to false

williameveretteggplant:
Unarchiving a NSCell element from a xib will have the editable property set to false unless explicitly set to true. This is consistent with how Apple does it. A textFieldCell element that has been set to not be editable in XCode will not have an "editable" property. And this will be interpreted by Cocoa as a non-editable text label.
The previous version where cells had editable set to true by default was resulting in a bunch of text labels being editable in UI when they were not supposed to be.


fredkiefer (approved):
Sounds reasonable to me

PR #258, opened 2024-04-11
Change to decoding of EquivalentKeyModifierFlags

williameveretteggplant:
In the existing code, when the keyEquivalentModifierMask is shifted right by 8 bits when decoding, and shifted left by 8 bits when encoding. I can see no reason for this. 

Furthermore, when attempting to decode a standard xib, this eliminates the standard modifier keys that are used. For example, if "command" is present in the xib, the mask is ored with NSCommandKeyMask, which is 0x10 0000. But then when initializing with this encoder, this is shifted right by 8 bits, producing 0x00 1000, and then anded with NSDeviceIndependentModifierFlagsMask (0xffff 0000) resulting in 0. So the command key modifier is lost.

Like I said, I think this 8 bit shift is simply a mistake, and should be removed.

### DIFF
--- a/Source/NSButtonCell.m
+++ b/Source/NSButtonCell.m
@@ -1641,7 +1641,7 @@
       // style and border.
       bFlags2 |= [self showsBorderOnlyWhileMouseInside] ? 0x8 : 0;
       bFlags2 |= (([self bezelStyle] & 0x7) | (([self bezelStyle] & 0x18) << 2));
-      bFlags2 |= [self keyEquivalentModifierMask] << 8;
+      bFlags2 |= [self keyEquivalentModifierMask] & NSDeviceIndependentModifierFlagsMask;
 
       switch ([self imageScaling])
 	{
@@ -1805,7 +1805,7 @@
           bFlags2 = [aDecoder decodeIntForKey: @"NSButtonFlags2"];
           [self setShowsBorderOnlyWhileMouseInside: (bFlags2 & 0x8)];
           [self setBezelStyle: (bFlags2 & 0x7) | ((bFlags2 & 0x20) >> 2)];
-          [self setKeyEquivalentModifierMask: ((bFlags2 >> 8) & 
+          [self setKeyEquivalentModifierMask: (bFlags2 & 
                                                NSDeviceIndependentModifierFlagsMask)];
 
 	  switch ((bFlags2 >> 6) & 3)


rfm:
Looking at the source, it seems to me that the original code is wrong/broken from a portability point of view, and the suggested 'fix' is not much better.

The problem here is that the flags are defined using bitfields in GSCodingFlags.h (there appears to be an attempt to get the compiler to order them the same way depending on the hardware byte order),  but the layout of bitfields is specifically compiler dependent (undefined) so you can't depend on they layout of bits at all.

This means that the result of a call to -encodeIntForKey

rfm:
I may be missing something, but looking at the source, it seems to me that the original code is wrong/broken from a portability point of view, and the suggested 'fix' is not much better.

The problem here is that the flags are defined using bitfields in GSCodingFlags.h
There appears to be an attempt to get the compiler to order them the same way depending on the hardware byte order (and since there are 24 relevant bits which would have been in a 32bit bitfield, the 8bit shift may be related to that endianness stuff),  but the layout of bitfields in C/C++/ObjC is specifically compiler dependent (undefined) so you can't depend on the layout of bits at all.

This means that the result of a call to encode_NSUInteger(aCoder, &_keyEquivalentModifierMask) is encoding an undefined value into the archive and the corresponding call to unarchive is getting an undefined value: if the archiving and unarchiving are done by different compilers the results may be garbage.

So the correct/portable way to do this would be to replace the archiving code with longer code which uses a temporary unsigned integer variable,  sets bits at explicit positions in that variable, and then encodes the temporary variable (unsigned integer) to the archive (with similar code for decoding from the archive and setting bitfields from the contents of the decoded value).

Alternatively, perhaps it would be better to stop using bitfields at all, and use shifting/masks everywhere that bitfields are used now.  That makes archiving/unarchiving simple, but means a lot of code changes.

rfm:
PS. as long as all compilers (I guess we only target GCC, Clang) use identical layout algorithms on all operating systems/architectures, my above comment is pretty academic.  I don't know that they do though.

williameveretteggplant:
Your concerns about encoding and then decoding on different platforms are valid, but resolving them is a much larger task. Honestly, I'm not sure when, in practice, you would ever encode an NSButtonCell for later decoding.

What specifically this is meant to fix is the common, and in fact routine, scenario of initializing a button from a XIB. In that case the XIB is decoded using something like GSXib5KeyedUnarchiver which creates the decoder used to initialize the NSButtonCell. And this solves that situation.

williameveretteggplant:
@fredkiefer don't want to nag, but is it possible you could take a look at this? It would be appreciated.

fredkiefer:
Sorry for not replying earlier. Like Richard I think that this change addresses some specific issues that you are facing but still the code looks as wrong as the previous one. What I would really like to understand is, on which platform your change works better than the previous one and if you have specific NIB or XIB files where one can see the difference.

williameveretteggplant:
It works better on all platforms. And it isn't a specific xib file, it's any that uses modifier keys with key equivalents.

Look at GSXib5KeyedUnarchiver.m : 1415. If "command" is present in the attributes it adds NSCommandKeyMask to the mask. As I wrote above 

> For example, if "command" is present in the xib, the mask is ored with NSCommandKeyMask, which is 0x10 0000. But then when initializing with this encoder, this is shifted right by 8 bits, producing 0x00 1000, and then anded with NSDeviceIndependentModifierFlagsMask (0xffff 0000) resulting in 0. So the command key modifier is lost.

This isn't specific to any platform. It's all platforms. And it isn't some specific edge case, it is any time you have modifier keys with key equivalents in a xib. The current implementation just completely breaks that. This fixes it.

fredkiefer:
I just did a few tests on MacOSX and there if I set the key equivalent to NSEventModifierFlagShift (1 << 17) it gets encoded as the value 1 << 25. This means the old code is correct, the values from the key equivalent mask get moved to the left by 8 bits. 
Now that you explained the issue that you are trying to fix, I would suggest that we correct the code in GSXib5KeyedUnarchiver.m instead. There the shift to the left is missing. It was my fault, I should have checked these changes from Eggplant more carefully. I was under the impression that this was actual production code that had been tested and used. 
Are you willing to provide the correct fix or should I do it myself? And are you going to do this in this pull request or will you open up a new one?

williameveretteggplant:
Ok. I think I will create a new pull request. This is going to require going back to the original, so I might as well start a new branch.
I think this code is something like 17 years old, and for a long time we've been operating off our own fork that just fixed it in a different way. It's only now that we are trying to get back to the master branch we are discovering these things.

PR #259, opened 2024-04-19
Left shift key equivalent modifier mask in xib unarchiver

williameveretteggplant:
This is necessary to offset the 8 bit right shift that occurs in the initWithCoder: method in NSButtonCell


fredkiefer (approved):
Looks good. We also need the same left shift (maybe inline?) on line 1398. Could you please add this?

williameveretteggplant:
Done

PR #260, opened 2024-04-23
Revert "Left shift key equivalent modifier mask in xib unarchiver"

williameveretteggplant:
This fix appears to have created serious side effects. It has broken input to text fields. I'm not sure why, but this needs to be reverted.


fredkiefer:
I am willing to revert this, as the change is still very fresh. But it really would help if you could provide an example that got broken by this. That way more people would be able to work on a proper fix.

PR #262, opened 2024-04-25
Reduce code duplication in drawViewCell:... method

gcasa:
Per discussions with @fredkiefer I am working through the code here to clean up and reduce duplication.


Source/NSTableColumn.m
```diff
@@ -609,7 +615,9 @@ - (id) initWithCoder: (NSCoder*)aDecoder
 
 	  if (version >= 4)
 	    {
+	      _tableView = [aDecoder decodeObject]; // not retained, tableView retains us...
```
fredkiefer, line 618: Did you add another value to the encoded format without increasing the version number? Hopefully nobody did create an archive with the intermediate format.

fredkiefer:
If I remember correctly there is also quite some duplication in drawing the special NSOutlienView column. Could you please extract that as well?

gcasa:
Yep.  Will do.

Source/NSTableColumn.m
gcasa: I considered this, but I think it's likely okay because we didn't do a release in the interim.   The new decoding is because the methods for registering the prototype views are on NSTableView according to Apple documentation.

gcasa:
Made some adjustments to the mouse interaction.  Please take a look and see if you approve of all of the changes.  Thanks @fredkiefer I hope you feel better soon.

fredkiefer (approved):
Looks good to me

gcasa:
> Looks good to me

Thanks @fredkiefer 🙏

PR #263, opened 2024-04-25
Improve decoding of keyEquivalentModifierMask in .xib files

qmfrederik:
The `keyEquivalentModifierMask` modifier mask:
- Maps to `NSKeyEquivModMask` for `NSMenuItem`
- Is packed in the `ButtonFlags2` value for `NSButtonCell`

If a `keyEquivalentModifierMask` element is not present, a default value is used (`NSCommandKeyMask` for `NSMenuItem` and `0` for `NSButtonCell`).  The presence of a `keyEquivalent` element does not influence this behavior.

This commit:
1. Updates `[GSXib5KeyedUnarchiver decodeModifierMaskForElement]` to return a `NSNumber`, so that it can signal the absence of the `keyEquivalentModifierMask` element by returning `nil` and the presence of a `keyEquivalentModifierMask` with no modifier flags set by returning `0`
2. Updates `[GSXib5KeyedUnarchiver decodeButtonFlags2ForElement]` to shift the modifier mask left by 8 bits when storing it in the `NSButtonFlags2` value.
3. Adds unit tests to test for this behavior.

This brings the bavior in line with that observed on macOS.

### DIFF
--- a/Source/GSXib5KeyedUnarchiver.m
+++ b/Source/GSXib5KeyedUnarchiver.m
@@ -1382,21 +1382,9 @@ didStartElement: (NSString*)elementName
   id            object     = nil;
   NSDictionary *attributes = [[element elementForKey: @"keyEquivalentModifierMask"] attributes];
 
-  // ??? SKIP modifier mask processing if BASE64-UTF8 string being used ???
   if (attributes == nil)
     {
-      if (([element elementForKey: @"keyEquivalent"]) &&
-          ([[element elementForKey: @"keyEquivalent"] attributeForKey: @"base64-UTF8"]))
-      {
-        object = [NSNumber numberWithUnsignedInt: 0];
-      }
-    else
-      {
-        // Seems that Apple decided to omit this attribute IF certain default keys alone
-        // are applied.  If this key is present WITH NO setting then the following is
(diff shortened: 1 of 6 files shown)


qmfrederik:
@williameveretteggplant I _think_ that the regression you observed with #259 was that your change shifted the `NSKeyEquivModMask` values left by 8 bits for `NSMenuItem`, rendering them useless -- the equivalent of not specifying any modifier key at all. As a result, the menus would run without a modifier key for their key equivalent, so you'd trigger a CTRL+O action by simply typing O in an editor window.  That correlates with your "but this needs to be reverted" -- I've seen the same behavior.

@fredkiefer Your note here https://github.com/gnustep/libs-gui/pull/258#issuecomment-2064984870 was correct; but for `NSButtonCell` only where the value gets stored in `NSButtonFlags2`.

I've double checked the behavior by writing a couple of XCTests which assert this behavior on macOS; you can see that test here: https://github.com/qmfrederik/xib-tests/blob/main/DemoTests/DemoTests.m.  That GitHub repo also contains a pipeline which runs the test to prove that they pass.

fredkiefer (approved):
Looks great. Thank you!

PR #268, opened 2024-05-03
Add NSMenuToolbarItem implementation

gcasa:
Add implementation of NSMenuToolbarItem.


gcasa:
Hey @fredkiefer let me know if there are any issues with this.  Thanks. :)

gcasa:
@fredkiefer This is a relatively simple change.   Please take a moment to review.

Source/NSMenuToolbarItem.m
```diff
@@ -0,0 +1,90 @@
...
+*/
+
+#import "AppKit/NSMenuToolbarItem.h"
+#import "AppKit/NSMenu.h"
+#import "AppKit/NSImage.h"
+
+#import "GNUstepGUI/GSTheme.h"
+
+@interface NSToolbarItem (Private)
+- (NSView *) _backView;
```
fredkiefer, line 32: You don‘t see to use this method. Is it a left over from a previous attempt implement this class? Maybe that other version was a better fit?

Source/NSMenuToolbarItem.m
```diff
@@ -0,0 +1,90 @@
...
+}
+
+- (void) setMenu: (NSMenu *)menu
+{
+  ASSIGN(_menu, menu);
+}
+
+- (void) _showMenu: (id)sender
+{
+  [[GSTheme theme] rightMouseDisplay: _menu
```
fredkiefer, line 86: Where is this menu being displayed? Should we fake an event here to pass a location or are there better ways to control that?

fredkiefer (commented):
Looks basically good but with two tiny questions.

Source/NSMenuToolbarItem.m
gcasa: The menu shows when you right-click on the NSToolbarItem icon that represents it.

Source/NSMenuToolbarItem.m
gcasa: This method is something I thought I would need but ended up not using.  I have removed it.

Source/NSMenuToolbarItem.m
gcasa: Here is an example on GNUstep:

menu_item_toolbar_test

And on the mac:

mac_menu_toolbar_item

Both are using the test code referenced in the previous comments.

gcasa:
Sorry about the lateness of my reply, I haven't been feeling well for several days.

fredkiefer (approved):
Look great

PR #271, opened 2024-05-08
Improve Endianness support for BitmapRep

rmottola:
Let's start with TIFF

### DIFF
--- a/Source/nsimage-tiff.h
+++ b/Source/nsimage-tiff.h
@@ -57,7 +57,9 @@ typedef struct {
     int     error;
     float   xdpi;
     float   ydpi;
-    uint16_t bitmapFormat;  /* NSBitmapFormat*/
+  short    isBigEndian;
+  short    is16Bit;
+  short    is32Bit;
 } NSTiffInfo; 
 
 typedef struct {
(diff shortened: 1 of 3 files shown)


Source/nsimage-tiff.h
```diff
@@ -57,6 +57,9 @@ typedef struct {
     int     error;
     float   xdpi;
     float   ydpi;
+  short    isBigEndian;
```
rmottola, line 60: @fredkiefer is this a good way to hold BOOLs ?
This file is Objective-C agnostic, so no BOOL available.

rmottola:
@fredkiefer First iteration done yesterday worked and fixes GWorkspace! So I committed and started a branch. Pull Request should make it easier to work together and track changes. Not done yet.

I didn't like stuffing NSBitmapFormat into uint16_t, so I created separate flags, this allows also some code simplification.

rmottola:
@fredkiefer I do not find in the code where NSBitmapFormatSixteenBitBigEndian is actually set for TIFF, so how does it get the information about Endianness and interprets bits. I want to copy/move the logic in NSTiffGetInfo to keep the info consistent.

rmottola:
@fredkiefer we discussed that three different methods needed to be updated in NSBitmapImageRep. But with the trick of having the logic in tiff.m and by setting the flags in `fillTIFFInfo: usingCompression: factor:`, are we covered in all cases or are still other places left?

PR #272, opened 2024-05-12
Revert to defaultParagraphStyle tab stops from 10 years ago

adamfox-tp:
As we are trying to move away from using our own custom forks and branches of GNUstep, we discovered this discrepancy between the main fork/branch and ours, and without this block of code, the tab stops on GNUstep platforms now have approximately twice as wide tab stops.  Looking back at [the commit that commented this block out 10 years ago](https://github.com/gnustep/libs-gui/commit/47dc5486848644580ab62678b35373c29738fa67), looks like the reason was:

```
Comment out tabs in default paragraph style as Apple doesn't seem to have them.
```

So, one way or another, we need to fix this in our product, and it seems there is a visual discrepancy between Apple and GNUstep.  We can fix it in our product by going through all the places we're currently leveraging the defaultParagraphStyle and using a custom style instead, but we wanted to figure out what the general consensus is here before doing that.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,7 +1,8 @@
-2024-05-12  Adam Fox <adam.fox@keysight.com>
+2024-05-13  Adam Fox <adam.fox@keysight.com>
 
-	* Source/NSParagraphStyle.m (+defaultParagraphStyle): Restore
-	tabs in default paragraph style and move it to -init.
+	* Source/NSParagraphStyle.m (-init): Apply the old default
+	tab stop behavior from +defaultParagraphStyle to -init,
+	and therefore also to +defaultParagraphStyle.
 
 2024-03-18 Fred Kiefer <FredKiefer@gmx.de>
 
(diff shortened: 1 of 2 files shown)


fredkiefer:
After this change things are surely more correct than before. But it looks like this code rather belongs into the -init method as Apple seems to set these values not just on the defaultParagraghStyle. Do you want to correct this or should we just push this change through as is and clean up later?

adamfox-tp:
I'm happy to move this to init, thanks for the swift reply!

adamfox-tp:
@fredkiefer I think the current changes are good, but please let me know if you have any other suggestions!

ChangeLog
```diff
@@ -1,3 +1,8 @@
+2024-05-12  Adam Fox <adam.fox@keysight.com>
+
+	* Source/NSParagraphStyle.m (+defaultParagraphStyle): Restore
```
fredkiefer, line 3: This does not match the new implementation, but we just ignore that.

fredkiefer (approved):

ChangeLog
adamfox-tp: Ah yeah, made sense in my head at the time but that's misleading, I'll change it.

adamfox-tp: Okay @fredkiefer I updated the change description, please let me know if there's anything else, thank you!

PR #275, opened 2024-05-13
ImageMagick TGA support fix

rmottola:
seek for trailing signature and recognize v2 TGA files and hint that to ImageMagick

works for v2 TGA files

### DIFF
--- a/Source/GSImageMagickImageRep.m
+++ b/Source/GSImageMagickImageRep.m
@@ -2,9 +2,11 @@
 
    <abstract>ImageMagick image representation.</abstract>
 
-   Copyright (C) 2011 Free Software Foundation, Inc.
+   Copyright (C) 2011-2024 Free Software Foundation, Inc.
    
    Author:  Eric Wasylishen <ewasylishen@gmail.com>
+            Riccardo Mottola
+
    Date: June 2011
    
    This file is part of the GNUstep Application Kit Library.
@@ -130,6 +132,7 @@
   ExceptionInfo *exception = AcquireExceptionInfo();
   ImageInfo *imageinfo = CloneImageInfo(NULL);
   Image *images, *image;
+  char signature[32];
   
   // Set the background color to transparent
   // (otherwise SVG's are rendered against a white background by default)
@@ -138,6 +141,14 @@
 #else
   QueryColorDatabase("none", &imageinfo->background_color, exception);
 #endif
+
+  bzero(signature, 32);
+  [data getBytes:signature range:NSMakeRange([data length]-18, 18)];
+  if (strncmp(signature, "TRUEVISION-XFILE.", 17) == 0)
+    {
+      NSWarnLog(@"Targa file detected!, giving a magick hint...");
+      strcpy(imageinfo->magick, "TGA");
+    }
 
   images = BlobToImage(imageinfo, [data bytes], [data length], exception);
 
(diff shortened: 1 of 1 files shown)


rmottola:
@eukara I think this might interest you

Source/GSImageMagickImageRep.m
```diff
@@ -139,6 +142,14 @@ + (NSArray*) imageRepsWithData: (NSData *)data allImages: (BOOL)allImages
   QueryColorDatabase("none", &imageinfo->background_color, exception);
 #endif
 
+  bzero(signature, 32);
+  [data getBytes:signature range:NSMakeRange([data length]-18, 18)];
```
fredkiefer, line 146: Here some additional spaces would improve readability

Source/GSImageMagickImageRep.m
```diff
@@ -129,7 +131,9 @@ + (NSArray*) imageRepsWithData: (NSData *)data allImages: (BOOL)allImages
 
   ExceptionInfo *exception = AcquireExceptionInfo();
   ImageInfo *imageinfo = CloneImageInfo(NULL);
-  Image *images, *image;
+  Image *images;
+  Image *image;
+  char signature[32];
```
fredkiefer, line 136: Why is the signature 32 bytes long when you only use 18 of them?

Source/GSImageMagickImageRep.m
```diff
@@ -139,6 +143,14 @@ + (NSArray*) imageRepsWithData: (NSData *)data allImages: (BOOL)allImages
   QueryColorDatabase("none", &imageinfo->background_color, exception);
 #endif
 
+  memset(signature, 0, 32);
+  [data getBytes: signature range: NSMakeRange([data length] - 18, 18)];
+  if (strncmp(signature, "TRUEVISION-XFILE.", 17) == 0)
+    {
+      NSWarnLog(@"Targa file detected!, giving a magick hint...");
+      strcpy(imageinfo->magick, "TGA");
```
fredkiefer, line 151: Although it is pretty save here, we normally don't use `strcpy`. Rather something like `strncpy`. Here it makes no difference but static code analysis tools may complain.

fredkiefer (approved):

Source/GSImageMagickImageRep.m
rmottola: Spaced better!

Source/GSImageMagickImageRep.m
rmottola: As a safety for future uses, we could try to handle more than one signature.
I was looking and there are other files misbehaving, so one could reuse it.

If you wish I could reduce it to exactly 18, I don't think 14bytes are a big issues though.

Source/GSImageMagickImageRep.m
rmottola: > Although it is pretty save here, we normally don't use `strcpy`. Rather something like `strncpy`. Here it makes no difference but static code analysis tools may complain.

is it needed even if the right value is a constant? looks overkill, a code analyzer shouldn't complain.

eukara:
Thank you very much Riccardo!

Source/GSImageMagickImageRep.m
rmottola: @fredkiefer to appease both, I defined a local constant. So I declare for now just the 18 bytes needed for TGA, but it is more evident that if you tinker with it, you need to update it. And memset will be correct too. Fine?

rmottola:
Tested on ImageMagick 6 & 7. Going to merge. Details can be smoothed out, @fredkiefer!

PR #265, opened 2024-04-25
NSTableRowView class implementation

gcasa:
Recent versions of macOS since 10.7+ use an NSTableRowView to represent a row in NSTableView and it's subclasses when the table is view-based.   This PR seeks to implement that functionality.

PR #277, opened 2024-06-19
Updated PR for NSOutlineView/NSTreeController due to refactor

gcasa:
Subject summarizes.  This PR is meant to implement NSTreeController and NSOutlineView integration.  I made a new branch since NSOutlineView was heavily modified in the PR addressing NSTableRowView [here](https://github.com/gnustep/libs-gui/pull/265).


gcasa:
@fredkiefer This is not yet ready for review.   I will convert it when it is.  Thanks.

gcasa:
Interesting indeed...
```
Found 20 methods on '_NSControllerTreeProxy'
	'_NSControllerTreeProxy' has method named 'count' of encoding 'Q16@0:8'
	'_NSControllerTreeProxy' has method named 'indexPath' of encoding '@16@0:8'
	'_NSControllerTreeProxy' has method named 'parentNode' of encoding '@16@0:8'
	'_NSControllerTreeProxy' has method named '_invokeSelector:withArguments:onKeyPath:ofObjectAtIndexPath:' of encoding 'v48@0:8:16@24@32@40'
	'_NSControllerTreeProxy' has method named '_setValue:forKeyPath:ofObjectAtIndexPath:' of encoding 'v40@0:8@16@24@32'
	'_NSControllerTreeProxy' has method named '_validateValue:forKeyPath:ofObjectAtIndexPath:error:' of encoding 'c48@0:8^@16@24@32^@40'
	'_NSControllerTreeProxy' has method named '_valueForKeyPath:ofObjectAtIndexPath:' of encoding '@32@0:8@16@24'
	'_NSControllerTreeProxy' has method named 'childNodes' of encoding '@16@0:8'
	'_NSControllerTreeProxy' has method named 'countForIndexPath:' of encoding 'Q24@0:8@16'
	'_NSControllerTreeProxy' has method named 'countOfChildNodes' of encoding 'Q16@0:8'
	'_NSControllerTreeProxy' has method named 'descendantNodeAtIndexPath:' of encoding '@24@0:8@16'
	'_NSControllerTreeProxy' has method named 'initWithController:' of encoding '@24@0:8@16'
	'_NSControllerTreeProxy' has method named 'isExpandableAtArrangedObjectIndexPath:' of encoding 'c24@0:8@16'
	'_NSControllerTreeProxy' has method named 'isLeaf' of encoding 'c16@0:8'
	'_NSControllerTreeProxy' has method named 'mutableChildNodes' of encoding '@16@0:8'
	'_NSControllerTreeProxy' has method named 'nodeAtIndexPath:' of encoding '@24@0:8@16'
	'_NSControllerTreeProxy' has method named 'objectAtIndexPath:' of encoding '@24@0:8@16'
	'_NSControllerTreeProxy' has method named 'objectInChildNodesAtIndex:' of encoding '@24@0:8Q16'
	'_NSControllerTreeProxy' has method named 'representedObject' of encoding '@16@0:8'
	'_NSControllerTreeProxy' has method named 'sortWithSortDescriptors:recursively:' of encoding 'v28@0:8@16c24'
```
I am introducing an equivalent class GSControllerTreeProxy to fill this role, it will be a subclass of NSTreeNode, just as _NSTreeControllerProxy is.   Apparently, NSTreeController is not meant to handle an NSArray of objects.   This is apparent from the documentation as ```- (NSTreeNode *) arrangedObjects``` is defined as such.

I will need to work out the nuances of how this handles various bindings.

gcasa:
The test for this is here... https://github.com/gcasa/NSTreeController_test

gcasa:
Here is NSOutlineView using a dataSource...

outline_normal

And here it is using a binding...

outline_tree_controller

This demonstrates that the new code does not affect outline view functionality in other applications.

gcasa:
I am marking this as ready for review and proceeding to optimize and clean up the code in preparation for merge.

gcasa:
Of course, Gorm works as well...

gorm_of_course

gcasa:
@fredkiefer These changes are mainly about the binding between NSTreeController and NSOutlineView.  More changes might be forthcoming, but I will do some of the other methods as separate PR.   Please take a moment to review.

Source/NSKeyValueBinding.m
```diff
@@ -122,6 +122,40 @@ - (void) unbind: (NSString *)binding
 
 @end
 
+@implementation NSString (PrivatePathExtension)
+
+- (NSString *) stringByRemovingFirstKeyPath
```
fredkiefer, line 127: This method looks like overkill. Why not just look for the first dot and remove everything before including the dot? And if there is no dot just return the string itself.
On the other hand we should also be careful not to provide to many additional methods on standard classes. This could easily clash with application code that adds a method of the same name. In my view it would be best to avoid this method, if there isn't something similar on Apple.

gcasa: I've moved the method to NSOutlineView as a private method.   I'm going to re-implement it to make it simpler.

gcasa: I have removed the private method entirely. :)

gcasa:
With the last change, it is now possible to update values from the outline view.  Please review if possible as I feel it is now ready for merge.  LMK if you have any constructive comments @fredkiefer.

gcasa:
Additional changes to NSTreeNode (adding children at specific index paths) will be addressed in a separate PR. This one is big enough already and they belong mainly in NSTreeNode and not NSTreeController.

gcasa:
Index insertion is working propertly... as is deletion.
index_insert

gcasa:
I am currently working on selection logic and refining some of the other methods since they are selection-based.   Once that is done I believe this PR should be reviewed and merged.   The next step for the NSTreeController changes should be related to integration with NSBrowser and that can be taken on as a separate PR.

gcasa:
It appears there is an adaptor between NSTableView and it's binding...

```
frame #2: 0x00007ff812b5bbf8 AppKit`-[NSBinder _setValue:forKeyPath:ofObject:mode:validateImmediately:raisesForNotApplicableKeys:error:] + 440
AppKit`-[NSBinder _setValue:forKeyPath:ofObject:mode:validateImmediately:raisesForNotApplicableKeys:error:]:
->  0x7ff812b5bbf8 <+440>: xorl   %r13d, %r13d
    0x7ff812b5bbfb <+443>: movq   -0x30(%rbp), %rax
    0x7ff812b5bbff <+447>: movq   0x30(%rax), %rdi
    0x7ff812b5bc03 <+451>: movq   0x3f3054c6(%rip), %rsi    ; "unsuppressSpecificNotificationFromObject:keyPath:"
(lldb) up
frame #3: 0x00007ff812b5b9f0 AppKit`-[NSBinder setValue:forBinding:error:] + 237
AppKit`-[NSBinder setValue:forBinding:error:]:
->  0x7ff812b5b9f0 <+237>: movl   %eax, %r12d
    0x7ff812b5b9f3 <+240>: movq   (%rbx), %rdx
    0x7ff812b5b9f6 <+243>: cmpq   %rdx, %r13
    0x7ff812b5b9f9 <+246>: je     0x7ff812b5ba1d            ; <+282>
(lldb) up
frame #4: 0x00007ff812b6882f AppKit`-[NSTableBinder tableView:didChangeToSelectedRowIndexes:] + 118
AppKit`-[NSTableBinder tableView:didChangeToSelectedRowIndexes:]:
->  0x7ff812b6882f <+118>: testb  %al, %al
    0x7ff812b68831 <+120>: je     0x7ff812b68877            ; <+190>
    0x7ff812b68833 <+122>: movq   (%r12), %rdx
    0x7ff812b68837 <+126>: movq   0x3f2f95ba(%rip), %rsi    ; "valueForBinding:resolveMarkersToPlaceholders:"
(lldb) up
frame #5: 0x00007ff812b5e89a AppKit`-[_NSBindingAdaptor tableView:didChangeToSelectedRowIndexes:] + 152
AppKit`-[_NSBindingAdaptor tableView:didChangeToSelectedRowIndexes:]:
->  0x7ff812b5e89a <+152>: incq   %r14
    0x7ff812b5e89d <+155>: cmpq   %r14, %r12
    0x7ff812b5e8a0 <+158>: jne    0x7ff812b5e854            ; <+82>
    0x7ff812b5e8a2 <+160>: addq   $0x28, %rsp
```

I may need to do something similar to get this working on GNUstep.

gcasa:
@fredkiefer Please take a look at this.  This PR is quite sizable, so I think this is a good stopping point.   The selection logic changes are done, the following things need to be addressed in a separate PR:

* methods that move nodes. 
* NSBrowser integration with NSTreeController et al.

Thanks for your patience and I am going to start any cleanup or optimization I can see in the current code and incorporate your suggestions as they are made.  TY. GC

Source/GSThemeDrawing.m
```diff
@@ -3531,8 +3534,12 @@ - (void) drawOutlineViewRow: (NSInteger)rowIndex
   NSInteger i;
   id dataSource = [outlineView dataSource];
   NSTableColumn *outlineTableColumn = [outlineView outlineTableColumn];
+  GSKeyValueBinding *theBinding = nil;
 
-  if (dataSource == nil)
+  theBinding = [GSKeyValueBinding getBinding: NSContentBinding
```
fredkiefer, line 3539: We should not need to access bindings from the theme drawing code. If we really have to, we should rearrange the code to get better input parameters.

Source/NSKeyValueBinding.m
```diff
@@ -192,8 +192,12 @@ + (GSKeyValueBinding *) getBinding: (NSString *)binding
   if (!objectTable)
     return nil;
 
+  // NSLog(@"+++ called with %@, %@", binding, anObject);
```
fredkiefer, line 195: Do we want to leave this debug code in? If so, I would suggest to change it to `NSDebugLog` instead of commenting it out.

Source/NSOutlineView.m
```diff
@@ -440,11 +457,57 @@ - (CGFloat) indentationPerLevel
...
 {
-  if (item == nil)
+  BOOL result = NO;
+  GSKeyValueBinding *theBinding = [GSKeyValueBinding getBinding: NSContentBinding
+						      forObject: self];
+  if (theBinding != nil)
     {
-      return NO;
+      BOOL leaf = YES;
+      id observedObject = [theBinding observedObject];
```
fredkiefer, line 466: I will need to inspect this usage of binding in more detail. At first look this is rather surprising as it breaks the loose coupling that bindings should provide but there may be valid reasons for this approach.

Source/NSOutlineView.m
```diff
@@ -2024,6 +2278,7 @@ - (void)_openItem: (id)item
   id object;
   id sitem = (item == nil) ? (id)[NSNull null] : (id)item;
 
+  // NSLog(@"openItem: %@", item);
```
fredkiefer, line 2281: Remove this

Source/NSTableView.m
```diff
@@ -3300,80 +3310,87 @@ - (void) validateEditing
       NSString *string;
       id newObjectValue = nil;
       BOOL validatedOK = YES;
+      GSKeyValueBinding *theBinding = nil;
```
fredkiefer, line 3313: I would prefer two definitions closer to the usages.

fredkiefer (changes requested):
I will need to spend more time on this to better understand the usages of the bindings.

gcasa:
> I will need to spend more time on this to better understand the usages of the bindings.

Understood. The included test shows the ones I was trying to support.  Let me know if you need my input or any clarifications.

Tests/gui/GSXib5KeyedUnarchiver/menu.m
```diff
@@ -35,7 +36,7 @@ int main()
   data = [NSData dataWithContentsOfFile:@"Menu.xib"];
   unarchiver = [GSXibKeyedUnarchiver unarchiverForReadingWithData:data];
   rootObjects = [unarchiver decodeObjectForKey: @"IBDocument.RootObjects"];
-  enumerator = [rootObjects objectenumerator];
+  enumerator = [rootObjects objectEnumerator];
```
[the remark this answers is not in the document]
gcasa: removed

Source/NSTreeController.m
```diff
@@ -127,123 +151,252 @@ - (BOOL) selectsInsertedObjects
...
 {
-  // FIXME
-  return [super content];
+  [self willChangeValueForKey: @"arrangedObjects"];
+  DESTROY(_arranged_objects);
+
+  if ([_content isKindOfClass: [NSArray class]])
+    {
+      NSMutableDictionary *dictionary = [GSControllerTreeProxy dictionaryWithChildren: _content];
+      _arranged_objects = [[GSControllerTreeProxy alloc] initWithRepresentedObject: dictionary
```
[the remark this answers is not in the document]
gcasa: true.  I will do this.

Source/GSControllerTreeProxy.m
```diff
@@ -0,0 +1,91 @@
...
+#import "AppKit/NSTreeController.h"
+
+#import "GSControllerTreeProxy.h"
+#import "GSBindingHelpers.h"
+
+@implementation GSControllerTreeProxy
+
++ (NSMutableDictionary *) dictionaryWithChildren: (NSMutableArray *)children
+{
+  NSMutableDictionary *dictionary = [NSMutableDictionary dictionaryWithObject: children
```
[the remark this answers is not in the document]
gcasa: I'll eliminate it.

Source/GSThemeDrawing.m
gcasa: I will take a look at this.

Source/NSKeyValueBinding.m
gcasa: Will do.

Source/NSOutlineView.m
gcasa: Done.

Source/NSTableView.m
gcasa: Done.

Source/GSThemeDrawing.m
gcasa: I am not certain this is even needed.

gcasa: The reason this is checked here is only because we need to make sure there is some data source that is available whether it is a traditional one or a binding.   I can add something to check this so that it doesn't need to be checked locally.  Either adding a parameter or adding a private ivar/accessor.

Source/NSOutlineView.m
gcasa: The issue here was that the selection changes were being reflected in the tree controller based on what was selected in the outline view.  I was unsure of how to achieve this without doing this.

Source/GSThemeDrawing.m
gcasa: I have moved the check into [NSOutlineView drawRow:clipRect] so that it is not theme specific.  I have checked both view based on non-view based outline view tests and they work fine.

Source/GSControllerTreeProxy.m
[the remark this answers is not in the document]
gcasa: Done.

Source/NSTreeController.m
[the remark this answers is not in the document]
gcasa: Done.

Source/GSControllerTreeProxy.h
```diff
@@ -0,0 +1,61 @@
...
+#if	defined(__cplusplus)
+extern "C" {
+#endif
+  
+@interface GSControllerTreeProxy : NSTreeNode
+{
+  NSTreeController *_controller;
+}
+
++ (NSMutableDictionary *) dictionaryWithChildren: (NSMutableArray *)children;
```
fredkiefer, line 41: This method no longer needs to be on the public interface. You might even remove it completely.

gcasa: Will do.

Source/NSOutlineView.m
gcasa: I won't resolve this until you say okay.

Source/GSControllerTreeProxy.h
gcasa: One thing... I am working on some fixes to when a child is added to a selected item.  I will go ahead and fix this and resolve this.   Then I will fix the insertion logic.

Source/NSOutlineView.m
gcasa: What I can say is that the behavior is the same as on macOS.   Whether it's not the same way is another matter. :/

gcasa:
behavior of addChild: method is now consistent with macOS

gcasa:
@fredkiefer Please review again, if possible.

fredkiefer (changes requested):
Looks better but it is really to big to review at once. Could you break out some changes into a separate MR?

gcasa:
> Looks better but it is really to big to review at once. Could you break out some changes into a separate MR?

I will take a look to see if any of the functionality can be broken out.   The issue is that much of this works together, but I think I may be able to break out a couple of things.  Will notify you.

Source/GSControllerTreeProxy.m
```diff
@@ -0,0 +1,90 @@
...
+			 forKey: @"value"];
+}
+
+// These return the value in the cases where the parent class method is called...
+- (NSArray *) childNodes
+{
+  return [self children];
+}
+
+- (NSArray *) mutableChildNodes
```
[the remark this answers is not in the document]
gcasa: I think it should be.  Will update.

ChangeLog
```diff
@@ -1,3 +1,21 @@
+2024-06-30 Gregory John Casamento <greg.casamento@gmail.com>
+
+	* Headers/AppKit/NSTreeController.h: Update header
+	* Headers/AppKit/NSTreeNode.h: Update header
+	* Source/GNUmakefile: Add new class
+	* Source/GSBindingHelpers.h: Add NSString category
+	* Source/GSControllerTreeProxy.[hm]: Proxy class
+	for NSTreeController, a subclass of NSTreeNode.
+	* Source/GSThemeDrawing.m: Optimize code a bit and
+	add changes for binfings.
```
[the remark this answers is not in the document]
gcasa: Will update.

Source/GSControllerTreeProxy.m
```diff
@@ -0,0 +1,90 @@
...
+#import <Foundation/NSArray.h>
+#import <Foundation/NSDictionary.h>
+#import <Foundation/NSString.h>
+
+#import "AppKit/NSTreeController.h"
+
+#import "GSControllerTreeProxy.h"
+#import "GSBindingHelpers.h"
+
+@implementation GSControllerTreeProxy
```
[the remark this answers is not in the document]
gcasa: Because, according to the test, the object being created and stored when the user causes NSTreeController is a subclass of NSTreeNode that contains a _representedObject that is a mutable NSDictionary with the keys value or children.   So while it may have been sufficient, it's not the same as would be expected if the user pulls the item from the NSTreeController whereas the GSControllerTreeProxy is analogous enough to the _NSControllerTreeProxy (see earlier in the comments) to be used compatibly.

Source/NSTreeNode.m
```diff
@@ -47,6 +47,7 @@ - (NSMutableArray*) _childNodes
 
 - (void) _setParentNode: (NSTreeNode*)parentNode
 {
+  NSLog(@"parentNode = %@", parentNode);
```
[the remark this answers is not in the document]
gcasa: Done

Source/NSTableView.m
```diff
@@ -2083,6 +2084,8 @@ - (void) dealloc
 {
   [self abortEditing];
 
+  [GSKeyValueBinding unbindAllForObject: self];
```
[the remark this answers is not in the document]
gcasa: Moved.

Source/NSKeyValueBinding.m
```diff
@@ -192,8 +192,11 @@ + (GSKeyValueBinding *) getBinding: (NSString *)binding
   if (!objectTable)
     return nil;
 
+  NSDebugLog(@"+++ called with %@, %@", binding, anObject);
```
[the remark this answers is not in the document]
gcasa: Done

Source/NSTableColumn.m
```diff
@@ -678,6 +686,11 @@ - (id) valueForKey: (NSString*)aKey
       // FIXME
       return [NSNumber numberWithBool: YES];
     }
+  else if ([aKey isEqual: NSEditableBinding])
+    {
+      // FIXME
+      return [NSNumber numberWithBool: YES];
```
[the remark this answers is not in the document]
gcasa: Made those changes

Source/NSTableView.m
```diff
@@ -6840,17 +6878,101 @@ - (BOOL) _isCellEditableColumn: (NSInteger) columnIndex
     }
 }
 
+- (void) _applyBindingsToCell: (NSCell *)cell
```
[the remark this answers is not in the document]
gcasa: I am in the process of moving this.

gcasa:
> Looks better but it is really to big to review at once. Could you break out some changes into a separate MR?

I have been looking at how to break this up.   It is going to be difficult to break it out functionality-wise because much of it works together.   I think we need to bite the bullet and just review it in chunks as it is.   I will make sure to do future PRs in smaller chunks.  I know this one is big.  My apologies @fredkiefer. :(

Source/NSTableView.m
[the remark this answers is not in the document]
gcasa: Moved.

gcasa:
@fredkiefer Please take another look.   Again, future PRs will be smaller.  Apologies

Source/NSTreeController.m
```diff
@@ -127,123 +169,258 @@ - (BOOL) selectsInsertedObjects
...
+
+      for (pos = 0; pos < length; pos++)
+	{
+	  NSUInteger i = [indexPath indexAtPosition: pos];
+	  id node = [children objectAtIndex: i];
+
+	  children = [node valueForKeyPath: _childrenKeyPath];
+	}
+
+      lastIndex = [indexPath indexAtPosition: length];
```
fredkiefer, line 351: This code here looks the same as the code inside the loop. If this is correct you could just run the loop one more iteration. Which could be achieved by removing the `- 1` when computing the `length`.

Source/NSTreeController.m
```diff
@@ -22,55 +22,92 @@
...
-*/ 
+@end
 
-#import <Foundation/NSArray.h>
-#import <Foundation/NSIndexPath.h>
-#import <Foundation/NSString.h>
-#import <Foundation/NSSortDescriptor.h>
+@implementation NSOutlineView (__NSTreeController_private__)
+
+- (NSArray *) _selectedIndexPaths
```
fredkiefer, line 58: I don't see where this method ever gets called. What do we need it for?

Source/NSTableView.m
```diff
@@ -2338,28 +2341,35 @@ - (void) setDataSource: (id)anObject
...
+	     format: @"Data Source doesn't respond to "
+	     @"tableView:objectValueForTableColumn:row:"];
+	  */
+	}
+
+      _dataSource_editable = [anObject respondsToSelector: sel_c];
+    }
+  else
+    {
+      /* Based on testing on macOS, this should default to YES if there is a binding...
```
fredkiefer, line 2368: If this is true, you don't have to check for the binding in line 3338. Looking at this variable should be enough.

Source/NSOutlineView.m
```diff
@@ -1815,14 +2035,32 @@ - (id) _objectValueForTableColumn: (NSTableColumn *)tb
 {
   id result = nil;
 
-  if ([_dataSource respondsToSelector:
-    @selector(outlineView:objectValueForTableColumn:byItem:)])
+  // If we have content binding the data source is used only
+  // like a delegate
+  NSDictionary *info = [GSKeyValueBinding infoForBinding: NSValueBinding forObject: tb];
```
fredkiefer, line 2040: Looking at the binding of another object looks wrong to me. This code (and other similar one) belongs into NSTableColumn. I don't quite understand how it should be packaged up there, though.

Source/NSTreeNode.m
```diff
@@ -154,12 +154,13 @@ - (NSTreeNode*) descendantNodeAtIndexPath: (NSIndexPath*)path
 
 - (NSIndexPath*) indexPath
 {
+  NSLog(@"In %@", NSStringFromSelector(_cmd));
```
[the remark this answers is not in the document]
gcasa: Missed this one. Sorry.

Source/NSTreeController.m
gcasa: Fixed

Source/NSTreeController.m
gcasa: Removed

Source/NSTableView.m
gcasa: Removed

Source/NSTableColumn.m
```diff
@@ -650,6 +654,85 @@ - (NSString *) title
   return [_headerCell stringValue];
 }
 
+- (void) _applyBindingsToCell: (NSCell *)cell
+			atRow: (NSInteger)index
+{
+  GSKeyValueBinding *theBinding = nil;
+  NSFont *font = nil;
+  
+  theBinding = [GSKeyValueBinding getBinding: NSEditableBinding
```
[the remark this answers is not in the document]
gcasa: What confused/concerned me here was that the cells were created and populated into the table column, but the binding existed on the NSTableColumn itself.   That's why I wrote this method.  I am not sure the bindings will be automatically applied here.

Source/NSOutlineView.m
gcasa: The issue here is this.   I believe that NSOutlineView on macOS works a bit differently than the one we have implemented.   Ours preloads the visible items and then gets them progressively as the items are expanded.   As such it is not always getting the self.whateverProperty pattern of stuff from the items.   Sometimes it is necessary to get things as shown in that method.   I believe it can be made more elegant, but I am not sure how at this point.

Source/NSTableView.m
gcasa: Both instances.

gcasa:
@fredkiefer Are there any additional changes you feel are needed here?

Source/NSOutlineView.m
```diff
@@ -194,8 +223,11 @@ - (id) initWithFrame: (NSRect)frame
 
 - (void) dealloc
 {
+  [GSKeyValueBinding unbindAllForObject: self];
```
fredkiefer, line 226: Do we still need this, now that you have the call on NSControl?

gcasa: I don't think so.  I actually thought I had removed it.

gcasa: Removed.

gcasa:
@fredkiefer Can we take an iterative approach on this?  I don't want us to get stuck in a cycle of review.  My concern is trying to refactor NSOutlineView to make things may be a ton of work (in effort if not in volume).

Source/NSTableColumn.m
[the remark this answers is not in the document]
fredkiefer: If you are unsure, please test it. That is the quickest way of gathering information about code. In my view the line 
`[cell setEditable: _is_editable];` should be sufficient for this specific binding.

Source/NSTableView.m
```diff
@@ -6839,18 +6786,20 @@ - (BOOL) _isCellEditableColumn: (NSInteger) columnIndex
       return [cell isEditable];
     }
 }
-
```
fredkiefer, line 6842: There should be an empty line between the two methods.

gcasa: Done

Source/NSTableColumn.m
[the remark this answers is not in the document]
gcasa: Changed.

Source/NSTreeController.m
```diff
@@ -80,13 +102,21 @@ - (void) dealloc
   RELEASE(_countKeyPath);
   RELEASE(_leafKeyPath);
   RELEASE(_sortDescriptors);
+  RELEASE(_arranged_objects);
+
```
fredkiefer, line 106: Shouldn't we release `_selection_index_paths` here as well?

gcasa: Done

Source/NSTreeController.m
```diff
@@ -127,123 +154,252 @@ - (BOOL) selectsInsertedObjects
...
+	{
+	  NSMutableArray *newContent = [NSMutableArray arrayWithArray: [self content]];
+	  GSControllerTreeProxy *node = [[GSControllerTreeProxy alloc]
+					  initWithContent: newObject
+					   withController: self];
+
+	  [newContent addObject: node];
+
+	  [self setContent: newContent];
+	  RELEASE(newObject);
```
fredkiefer, line 289: Don't we need to release `node` as well?

Source/NSTreeController.m
```diff
@@ -127,123 +154,252 @@ - (BOOL) selectsInsertedObjects
...
+- (IBAction) addChild: (id)sender
 {
-  // FIXME
+  NSIndexPath *p = [self selectionIndexPath];
+  id newObject = [self newObject];
+
+  if (p != nil)
+    {
+      [self insertObject: newObject atArrangedObjectIndexPath: p];
+    }
```
fredkiefer, line 302: In the last method you did release `newObject`, in this you don't seem to do this. Which way is correct?

Source/NSTreeController.m
gcasa: Done

Source/NSTreeController.m
gcasa: Done

gcasa:
@fredkiefer Please review and flag any final issues.  Currently, the test is working as expected and matches the behavior that is evident on macOS.  Can any additional concerns be addressed in a separate PR?

Source/NSTreeController.m
```diff
@@ -127,123 +155,242 @@ - (BOOL) selectsInsertedObjects
...
-- (BOOL) setSelectionIndexPath: (NSIndexPath*)indexPath
+- (BOOL) setSelectionIndexPath: (NSIndexPath *)indexPath
 {
-  // FIXME
-  return NO;
+  BOOL f = [self commitEditing];
+
+  if (YES == f)
+    {
+      [_selection_index_paths addObject: indexPath];
```
fredkiefer, line 164: The name of the method rather suggests that it replaces the old content. Just like the next method.

Source/NSTreeController.m
```diff
@@ -127,123 +155,242 @@ - (BOOL) selectsInsertedObjects
...
+- (NSIndexPath *) selectionIndexPath
 {
-  // FIXME
-  return nil;
+  return [_selection_index_paths objectAtIndex: 0];
+}
+
+- (NSArray *) selectionIndexPaths
+{
+  return _selection_index_paths;
```
fredkiefer, line 249: Shouldn't we return an immutable copy here?

Source/NSTreeController.m
```diff
@@ -127,123 +155,242 @@ - (BOOL) selectsInsertedObjects
...
 }
 
-- (void) addChild: (id)sender
+- (IBAction) add: (id)sender
 {
-  // FIXME
+  NSIndexPath *p = [NSIndexPath indexPathWithIndex: 0];
+  id newObject = [self newObject];
+  [self insertObject: newObject atArrangedObjectIndexPath: p];
 }
```
fredkiefer, line 277: Now the release of `newObject` is missing here. According to my comment on the next method it was the other way around before. I still think it should be similar in both methods and as `newObject` returns a retained object, we need to release it here.

fredkiefer (commented):
This PR is so huge, whenever I look I still find wrong bits I never inspected before. And then there are new bugs that must have been inserted with recent changes. But most of all I am still uncomfortable with the way KVO is used here. Either we did implement the basic mechanism of KVO wrong or this new code is using these wrongly. I cannot decide.

As for merging and fixing stuff later on. I don't believe the later on part will ever happen. But if you feel more comfortable with your changes you can always go ahead and merge.

gcasa:
> This PR is so huge, whenever I look I still find wrong bits I never inspected before. And then there are new bugs that must have been inserted with recent changes. But most of all I am still uncomfortable with the way KVO is used here. Either we did implement the basic mechanism of KVO wrong or this new code is using these wrongly. I cannot decide.
> 
> As for merging and fixing stuff later on. I don't believe the later on part will ever happen. But if you feel more comfortable with your changes you can always go ahead and merge.

As always I would rather wait for your approval.  I realize it's a big PR.  I have trouble splitting it because it's all functional together and it's also difficult to verify unless it's working properly.  Also, I didn't know it would be this big beforehand.  I understand your misgivings about how KVO is used in a couple of places, but I am not sure we can get around that without a complete rewrite of NSOutlineView.  It was hell to get it to work in the first place and it's going to be hell to re-think it.   I am going to spend some time today trying to rethink how this was done so I can improve it and maybe it will make us BOTH a little happier with merging.

The NSBrowser change is also part of this as it too can use NSTreeController.  The current NSBrowser changes don't incorporate the bindings as I was going to get that one merged and then do the NSTreeController and binding changes there as well and revisit the ones done here to some degree.  It was then I would revisit it.  

OFF TOPIC: Just FYI, I am working on refactoring the NSStoryboard code (I recall this is something you were worried about us not revisiting).   More on that later.

Source/NSOutlineView.m
gcasa: Ahhh, yes, I will do that.   I see what you mean.

Source/NSOutlineView.m
```diff
@@ -1642,6 +1741,84 @@ before the actual editing is started (otherwise you easily end up
...
+	    }
+	}
+    }
+
+  return nil;
+}
+
+- (NSIndexPath *) _indexPathForItem: (id)item
+{
+  id rootItem = nil;
```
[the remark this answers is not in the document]
gcasa: Done

Source/NSTreeController.m
gcasa: Ah, yes, okay... fixing.

Source/NSTreeController.m
gcasa: Done

Source/NSTreeController.m
gcasa: Done

Source/NSOutlineView.m
```diff
@@ -1642,6 +1741,84 @@ before the actual editing is started (otherwise you easily end up
...
+  return [self _findIndexPathForItem: item
+			  parentItem: rootItem];
+}
+
+- (void) _indexPathsFromSelectedRows
```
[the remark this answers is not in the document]
gcasa: Done.

Source/NSTableView.m
```diff
@@ -6873,16 +6823,41 @@ - (id) _objectValueForTableColumn: (NSTableColumn *)tb
...
+  GSKeyValueBinding *theBinding;
+
+  theBinding = [GSKeyValueBinding getBinding: NSValueBinding
+				   forObject: tb];
+  if (theBinding != nil)
+    {
+      NSArray *items = [[theBinding observedObject] arrangedObjects];
+      if (items != nil)
+	{
+	  NSDictionary *info = [GSKeyValueBinding infoForBinding: NSValueBinding
```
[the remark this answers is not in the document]
gcasa: I am going to come back to this in a bit.

Source/NSOutlineView.m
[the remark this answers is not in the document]
gcasa: I made this change, but it seems to crash if I autorelease it.

Source/NSTableView.m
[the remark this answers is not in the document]
gcasa: This has been refactored.  I added the suggested method, so info is no longer retrieved. Should we make the change always to get the observedObject from info?  I added the convenience method so we wouldn't always have to do that.   Let me know what you think, then resolve this if you feel as if the change is good.

gcasa:
> This PR is so huge, whenever I look I still find wrong bits I never inspected before. And then there are new bugs that must have been inserted with recent changes. But most of all I am still uncomfortable with the way KVO is used here. Either we did implement the basic mechanism of KVO wrong or this new code is using these wrongly. I cannot decide.
> 
> As for merging and fixing stuff later on. I don't believe the later on part will ever happen. But if you feel more comfortable with your changes you can always go ahead and merge.

I believe the issue is two fold... 
---
1) NSOutlineView stores it's data in a map where it's basically... item -> (item1, item2, item3)... item1 -> (child1, child2, child3) etc...   it does this by calling _loadDictionary... in order to not have to re-write NSOutlineView I is necessary to grab the last parts of the key rather than the whole key since we no longer have the "root" to reference, just the object.   I think this is "okay" but it's certainly not ideal.

2) The issue it the reverse setting of the selected indexes is the issue that is, I think, of the main concern.  I believe that KVO was done right but I don't know how to explain this behavior in macOS as it certainly does change it in the AppController when the value is updated on NSTreeController.

I am trying to think of why this is necessary and if there is a cleaner way to do it.   I believe that code is certainly needed to do this as I don't think it's going to happen automatically.   I am going to do more tests on macOS to see how it is handled there either to justify how we are doing it or help to discover a better way.

gcasa:
@fredkiefer The more I dig into this the more I am convinced that this is a special case for NSOutlineView.   There are a couple of reasons for this...

* While debugging I found an object called NSBindingAdapter (or similar) when using NSTreeController.   I believe this plays a role in pushing back the selectedIndexPaths to the NSTreeController
* Based on my tests our implementation of KVO is correct, so I don't see a problem there.  If this is the case, then the data set in selectedIndexPaths shouldn't automatically be reflected UNLESS there is a binding and there isn't in the test program so this MUST be something that is being done behind the scenes.

As we have seen before sometimes Apple sometimes implements special cases  *sigh*. I have spent the last day or two thinking into this and investigating and I can't see any other solution.

Source/NSOutlineView.m
gcasa: I added this method.  It seems to clean things up a lot.

Source/NSTableView.m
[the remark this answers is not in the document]
gcasa: I added the suggested method.  I think leaving the convenience method in is okay at this point.  I am going to mark this as resolved.  If you have an objection, please re-open.

fredkiefer:
@gcasa if you have to, merge this. The code reminds me a lot of the mess we had in NSArrayController before I rewrote the complete implementation there. After that most of the bits just fell in place and everything was a lot easier and almost understandable. But it did cost me a lot of time and this case seems even more complicated. The real power of bindings is that it not only is easy to use but also should provide a consistent inner working, which we haven't achieved here. And I don't believe in anybody trying to clean up later on. This just never happens.

fredkiefer (approved):
The code is working, maybe this is as much as we can achieve.

gcasa:
> The code is working, maybe this is as much as we can achieve.

I am doing some experiments to see if we can improve the code at least in NSTableView.  I am trying to get rid of some of the strangeness and I am not ruling out that this might be due to my lack of understanding.

gcasa:
I'm going to merge what we have and keep working on this and simplify it as much as possible.  There is additional integration with NSBrowser and I am HOPING it might shed some light on what else can be done to improve it.

gcasa:
> @gcasa if you have to, merge this. The code reminds me a lot of the mess we had in NSArrayController before I rewrote the complete implementation there. After that most of the bits just fell in place and everything was a lot easier and almost understandable. But it did cost me a lot of time and this case seems even more complicated. The real power of bindings is that it not only is easy to use but also should provide a consistent inner working, which we haven't achieved here. And I don't believe in anybody trying to clean up later on. This just never happens.

Given that I have a follow on PR in this very same code, I will.  I just don't like things sitting out in PRs for months as they tend to code rot. :/

gcasa:
@fredkiefer Merging, with the proviso that I come back to this in the PR about NSBrowser bindings...  I will create that one tonight.  I already have some changes with look like they could simplify things in NSTableView to use bindings more consistently.   I am hoping once I understand that better that I can translate that knowledge to NSOutlineView.

fredkiefer:
Just to make this clear. I am impressed by what you did here. I am also not able to solve the code issues in a pleasing way. Still I am convinced there must be one, although it is hard to find.

gcasa:
> Just to make this clear. I am impressed by what you did here. I am also not able to solve the code issues in a pleasing way. Still I am convinced there must be one, although it is hard to find.

Thank you, Fred, that really means a lot to me.  @fredkiefer I have been puzzling over simplifying this code for a week.  I am hoping the new changes for NSBrowser will help illuminate something.  I have a personal saying and that is that "the code will always show you the way" because you can intuitively feel when you're forcing something to happen and you shouldn't have to.

PR #278, opened 2024-06-23
Minor renamings of some APIs

ethanc8:
This commit implements Apple's renaming of some APIs, such as another enum `NSProgressIndicatorStyle` that Apple had renamed in order to allow it to be parsed by the Swift compiler, the new typedef `typedef NSString *NSPasteboardType;`, and the `-[NSViewController view`{`Will`|`Did`}{`Appear`|`Disappear`}`]` variants which take no arguments.


gcasa (changes requested):
It would be best to, instead of renaming them, do a similar version ifdef for the version of macOS this change was made in.   We always want to be as backwards compatible as possible.   It is okay in an enum for multiple names to be assigned the same value.

ethanc8:
> It would be best to, instead of renaming them, do a similar version ifdef for the version of macOS this change was made in. We always want to be as backwards compatible as possible. It is okay in an enum for multiple names to be assigned the same value.

I did not rename any APIs -- I just added the new APIs, and the enums have been defined. I might need to add some guards around the new enum names, though.

gcasa:
> > It would be best to, instead of renaming them, do a similar version ifdef for the version of macOS this change was made in. We always want to be as backwards compatible as possible. It is okay in an enum for multiple names to be assigned the same value.
> 
> I did not rename any APIs -- I just added the new APIs, and the enums have been defined. I might need to add some guards around the new enum names, though.

I apologize, I wasn't clear, this is what I meant.  It's so that if someone is using a particular version of the API they don't have access to the enums under the new names.

fredkiefer (approved):
Looks good

fredkiefer:
Could you please add the #ifdef that Gregory requested so we can merge this.

ethanc8:
> Could you please add the #ifdef that Gregory requested so we can merge this.

I will do that tomorrow.

ethanc8:
I have rebased it on `master` and will try updating the patches soon.

ethanc8:
@fredkiefer @gcasa I believe this is ready to merge into master. The CI errors are due to unrelated issues with the workflow YAML definitions.

Source/NSViewController.m
```diff
@@ -83,6 +83,7 @@ - (void)viewDidLoad
 
 - (void) viewWillAppear: (BOOL)animated
 {
+  [self viewWillAppear];
```
fredkiefer, line 86: I think the call order should be the other way around. That is in viewWillAppear we should call viewWillAppear: with NO.

ethanc8: [Apple's docs indicate](https://developer.apple.com/documentation/appkit/nsviewcontroller/viewwillappear()?language=objc) that developers are supposed to override `viewWillAppear`. `viewWillAppear:` seems to be a GNUstep extension; it never appears in [NSViewController.h](https://github.com/phracker/MacOSX-SDKs/blob/master/MacOSX11.3.sdk/System/Library/Frameworks/AppKit.framework/Versions/C/Headers/NSViewController.h).

ethanc8: So if in `viewWillAppear` we call `viewWithAppear:NO` people will subclass NSViewController and implement `viewWillAppear`, but any calls to `viewWillAppear:NO` in gnustep-gui will bypass their implementation. But if in `viewWillAppear:` we call `viewWithAppear`, we will use people's implementations in their subclasses whenever gnustep-gui calls `viewWithAppear:NO`.

fredkiefer: Apple once defined these methods with an animation flag, but then decided to remove it again. Now there may be old code out there that has the methods with the flag or new code that doesn't have the flag. And we need to be able to handle both cases.
There is one horrible way to do so, that I implemented in NSDocumentController. We would check for each method whether it was redefined from the version in NSViewController and in that code call the redefined implementation.
But I think in this case we could have it easier. When we replace all the calls to the old methods in NSViewController to ones to the new ones and from the new method stubs call the old ones, as I suggested above, then we should automatically get any overridden method. One could argue that your implementation already does the same, with less code changes. To me it still looks the wrong way around. But the only way to resolve this would be to check how the implementation on Apple is internally working.
So I will merge your version and we will see if anybody complains.

fredkiefer (approved):

PR #281, opened 2024-07-19
Add new 10.6 delegate methods to NSBrowserDelegate

gcasa:
The purpose of this branch is to add the NSBrowserDelegate methods defined in the documentation for macOS 10.6+ here...

https://developer.apple.com/documentation/appkit/nsbrowserdelegate/1407572-browser?language=objc

The test for this branch is here...

https://github.com/gcasa/NSBrowser_test

### DIFF
--- a/Headers/AppKit/NSBrowser.h
+++ b/Headers/AppKit/NSBrowser.h
@@ -100,6 +100,9 @@ APPKIT_EXPORT_CLASS
   int _lastVisibleColumn;
   NSString *_columnsAutosaveName;
   NSBrowserColumnResizingType _columnResizing;
+
+  BOOL _itemBasedDelegate;
+  id _lastItemLoaded;
 }
 
 //
@@ -357,6 +360,8 @@ canDragRowsWithIndexes: (NSIndexSet *)rowIndexes
        withEvent: (NSEvent *)event;
 #endif
 #if OS_API_VERSION(MAC_OS_X_VERSION_10_6, GS_API_LATEST)
+- (NSInteger) browser: (NSBrowser *)browser
+              numberOfChildrenOfItem: (id)item;
 - (id) browser: (NSBrowser *)browser
          child: (NSInteger)index
         ofItem: (id)item;
--- a/Headers/AppKit/NSBrowser.h
+++ b/Headers/AppKit/NSBrowser.h
@@ -37,7 +37,6 @@
 @class NSArray;
 @class NSIndexPath;
 @class NSIndexSet;
-@class NSMutableDictionary;
 
 @class NSCell;
 @class NSEvent;
--- a/Headers/AppKit/NSBrowser.h
+++ b/Headers/AppKit/NSBrowser.h
@@ -37,13 +37,13 @@
 @class NSArray;
 @class NSIndexPath;
 @class NSIndexSet;
+@class NSMutableDictionary;
 
 @class NSCell;
 @class NSEvent;
 @class NSMatrix;
 @class NSScroller;
 @class NSViewController;
-//@class NSBox;
 
 #if OS_API_VERSION(MAC_OS_X_VERSION_10_3, GS_API_LATEST)
 enum _NSBrowserColumnResizingType
@@ -63,7 +63,6 @@ APPKIT_EXPORT_CLASS
   Class _browserMatrixClass;
   NSString *_pathSeparator;
   
-  //NSBox *_horizontalScrollerBox;
   NSScroller *_horizontalScroller;
   NSTimeInterval _lastKeyPressed;
   NSString *_charBuffer;
@@ -102,7 +101,6 @@ APPKIT_EXPORT_CLASS
   NSBrowserColumnResizingType _columnResizing;
 
   BOOL _itemBasedDelegate;
-  id _lastItemLoaded;
 }
 
 //
--- a/Headers/AppKit/NSBrowser.h
+++ b/Headers/AppKit/NSBrowser.h
@@ -42,6 +42,7 @@
 @class NSEvent;
 @class NSMatrix;
 @class NSScroller;
+@class NSViewController;
 //@class NSBox;
 
 #if OS_API_VERSION(MAC_OS_X_VERSION_10_3, GS_API_LATEST)
@@ -355,6 +356,29 @@ canDragRowsWithIndexes: (NSIndexSet *)rowIndexes
         inColumn: (NSInteger)column
        withEvent: (NSEvent *)event;
 #endif
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_6, GS_API_LATEST)
+- (id) browser: (NSBrowser *)browser
+         child: (NSInteger)index
+        ofItem: (id)item;
+- (BOOL) browser: (NSBrowser *)browser
+      isLeafItem: (id)item;
+- (BOOL) browser: (NSBrowser *)browser
+  shouldEditItem: (id)item;
+- (id) browser: (NSBrowser *)browser
+       objectValueForItem: (id)item;
+- (id) browser: (NSBrowser *)browser
+       setObjectValue: (id)object
+       forItem: (id)item;
+- (id) rootItemForBrowser: (NSBrowser *)browser;
+- (NSViewController *) browser: (NSBrowser *)browser
+                       previewViewControllerForLeafItem: (id)item;
+- (NSViewController *) browser: (NSBrowser *)browser
+                       headerViewControllerForItem: (id)item;
+- (void) browser: (NSBrowser *)browser
+         didChangeLastColumn: (NSInteger)oldLastColumn
+        toColumn: (NSInteger)column;
+#endif
+
 @end
 
 #if OS_API_VERSION(MAC_OS_X_VERSION_10_3, GS_API_LATEST)
--- a/Headers/AppKit/NSBrowser.h
+++ b/Headers/AppKit/NSBrowser.h
@@ -359,7 +359,7 @@ canDragRowsWithIndexes: (NSIndexSet *)rowIndexes
 #endif
 #if OS_API_VERSION(MAC_OS_X_VERSION_10_6, GS_API_LATEST)
 - (NSInteger) browser: (NSBrowser *)browser
-              numberOfChildrenOfItem: (id)item;
+numberOfChildrenOfItem: (id)item;
 - (id) browser: (NSBrowser *)browser
          child: (NSInteger)index
         ofItem: (id)item;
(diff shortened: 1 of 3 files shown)


gcasa:
I am going to do some cleanup, also there are some trivial changes needed to support loading an NSViewController if one is necessary for the title view, etc.

gcasa:
I believe that support for the following delegate methods should be left for a separate PR:
```
- (BOOL) browser: (NSBrowser *)browser
  shouldEditItem: (id)item;
- (id) browser: (NSBrowser *)browser
       setObjectValue: (id)object
       forItem: (id)item;
- (NSViewController *) browser: (NSBrowser *)browser
                       previewViewControllerForLeafItem: (id)item;
- (NSViewController *) browser: (NSBrowser *)browser
                       headerViewControllerForItem: (id)item;
- (void) browser: (NSBrowser *)browser
         didChangeLastColumn: (NSInteger)oldLastColumn
        toColumn: (NSInteger)column;
```
As they might require significant changes to how NSBrowser is implemented.  I am researching this now to see if it's possible in the current implementation.

gcasa:
The code to call the delegate method 

```
- (NSViewController *) browser: (NSBrowser *)browser
                       headerViewControllerForItem: (id)item;
```

has been implemented.  The rest will have to wait for a separate PR as the ones involved in editing will require a more substantial change.  The other two are not strictly necessary as GS doesn't currently support column re-arrangement in NSBrowser and the leaf preview is optional.

gcasa:
@fredkiefer Please review when possible.  GC

Source/NSBrowser.m
```diff
@@ -3205,58 +3242,103 @@ - (void) _remapColumnSubviews: (BOOL)fromFirst
...
+{
+  id item = nil;
+
+  if (column == 0)
+    {
+      item = [_browserDelegate rootItemForBrowser: self];
+    }
+  else
+    {
+      NSInteger col = column > 0 ? column - 1 : column;
```
fredkiefer, line 3298: This line seems strange. Do we have negative columns? If not the value could either be 0 or bigger than zero and the zero case is already handled in the previous `if` case, so there is no need to be checked here.

Source/NSBrowser.m
```diff
@@ -3205,58 +3242,103 @@ - (void) _remapColumnSubviews: (BOOL)fromFirst
...
+      bc = [_browserColumns objectAtIndex: col];
+      if (bc != nil)
+	{
+	  NSMatrix *matrix;
+
+	  matrix = [bc columnMatrix];
+	  if (matrix != nil)
+	    {
+	      NSArray *selectedCells = [matrix selectedCells];
+	      if (selectedCells != nil && [selectedCells count] > 0)
```
fredkiefer, line 3310: The whole logic of this method seems strange to me. What is your intention here?
This will work for some very common cases, for example when a new column needs to be loaded because we double click on an item in the column before. But I am not sure that every time we need to load a new column there is always exactly one item selected in the previous column.

Source/NSBrowser.m
```diff
@@ -3306,68 +3388,88 @@ - (void) _performLoadOfColumn: (NSInteger)column
...
+      // Iterate over the children for the item....
+      for (i = 0; i < rows; i++)
+	{
+	  id aCell = [matrix cellAtRow: i column: 0];
+	  if (![aCell isLoaded])
+	    {
+	      BOOL leaf = YES;
+	      id val = nil;
+
+	      child = [_browserDelegate browser: self child: i ofItem: item];
```
fredkiefer, line 3416: We never make sure that the delegate actually supports all these methods.

fredkiefer (commented):
I think that this code will work in the most common case. I am unsure whether it is overall the correct solution. It might well be that I am missing some concept here.

Source/NSBrowser.m
gcasa: The idea is to find the item that needs to be loaded for a given column.   When multiple items are selected it will  show nothing on macOS.  I will make changes to make this code behave similarly.  See below this screenshot from macOS:

multiple_selection_macos

Source/NSBrowser.m
gcasa: From the test, it seems that when I comment out the code for various methods in this type of delegate that it doesn't throw an exception.

Source/NSBrowser.m
```diff
@@ -29,39 +32,43 @@
...
    You should have received a copy of the GNU Lesser General Public
    License along with this library; see the file COPYING.LIB.
-   If not, see <http://www.gnu.org/licenses/> or write to the 
-   Free Software Foundation, 51 Franklin Street, Fifth Floor, 
+   If not, see <http://www.gnu.org/licenses/> or write to the
+   Free Software Foundation, 51 Franklin Street, Fifth Floor,
    Boston, MA 02110-1301, USA.
 */
 
-#include <math.h>                  // (float)rintf(float x)
```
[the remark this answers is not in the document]
gcasa: I'll add it back.  It just seemed extraneous to me.

Source/NSBrowser.m
gcasa: We don't have negative columns.   The idea here is that if column is 0, then col will be zero and so we get the first column.   If it is greater than zero we get the previous column.   I will take a look and optimize this code.

Source/NSBrowser.m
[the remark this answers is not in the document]
gcasa: Added back.

Source/NSBrowser.m
fredkiefer: This sounds like they always check which methods are implemented.

Headers/AppKit/NSBrowser.h
```diff
@@ -355,6 +357,31 @@ canDragRowsWithIndexes: (NSIndexSet *)rowIndexes
...
+       setObjectValue: (id)object
+       forItem: (id)item;
+- (id) rootItemForBrowser: (NSBrowser *)browser;
+- (NSViewController *) browser: (NSBrowser *)browser
+                       previewViewControllerForLeafItem: (id)item;
+- (NSViewController *) browser: (NSBrowser *)browser
+                       headerViewControllerForItem: (id)item;
+- (void) browser: (NSBrowser *)browser
+         didChangeLastColumn: (NSInteger)oldLastColumn
+        toColumn: (NSInteger)column;
```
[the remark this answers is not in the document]
gcasa: Corrected for all declarations.

Source/NSBrowser.m
gcasa: I have simplified this code as discussed.

gcasa:
Multiple selection works just as on macOS:
multiple_selection_macos

Source/NSBrowser.m
gcasa: Hmm...  I'm a bit confused.   Like I said... all it does when I comment the code out for the implementations of these in the AppDelegate it just doesn't get the data.   I am guessing if these aren't implemented it doesn't set something that causes the browser to get the data.  I'll change the code to reflect this.

gcasa:
I believe the ones that are needed are
```
- (NSInteger) browser: (NSBrowser *)browser
numberOfChildrenOfItem: (id)item;
- (id) browser: (NSBrowser *)browser
         child: (NSInteger)index
        ofItem: (id)item;
- (BOOL) browser: (NSBrowser *)browser
      isLeafItem: (id)item;
```
I will make the changes to check for these.

Source/NSBrowser.m
gcasa: I believe the ones that are needed are
```
- (NSInteger) browser: (NSBrowser *)browser
numberOfChildrenOfItem: (id)item;
- (id) browser: (NSBrowser *)browser
         child: (NSInteger)index
        ofItem: (id)item;
- (BOOL) browser: (NSBrowser *)browser
      isLeafItem: (id)item;
```
I will make the changes to check for these.

gcasa: I believe this is the check that is happening.  On macOS the data simply doesn't load.   This change does the same thing.

gcasa:
> I think that this code will work in the most common case. I am unsure whether it is overall the correct solution. It might well be that I am missing some concept here.

I am actually unsure how it's not the correct solution.

gcasa:
Also...
1) what is the complex case here?
2) even if it does only handle the "simple case" it is infinitely better to be able to do that than to handle NONE of them that use this type of delegate.

gcasa:
I am going to enhance the test to see if I can address any of the other possible use-cases.

gcasa:
@fredkiefer I have added extra data to the test.   I believe that the changes I have made should cover most, if not all cases.  Please take a look.

gcasa:
@fredkiefer Are there any additional changes you feel are necessary here?

gcasa:
@fredkiefer The behavior here matches that on macOS. Do you have any final review comments?

Source/NSBrowser.m
fredkiefer: The implementation does not match the description you give above. For multiple selections the first gets used.

fredkiefer (approved):
Just fix these minor issues and then the code should be ready for a merge.

Headers/AppKit/NSBrowser.h
```diff
@@ -37,12 +37,13 @@
 @class NSArray;
 @class NSIndexPath;
 @class NSIndexSet;
+@class NSMutableDictionary;
```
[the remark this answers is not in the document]
gcasa: Done

gcasa:
As you can see from the above it does seem to function properly.   The _itemForColumn: method gets the list for the current column based on the previous column's selection.

fredkiefer:
> As you can see from the above it does seem to function properly. The _itemForColumn: method gets the list for the current column based on the previous column's selection. So given this, I believe it is okay to merge. I am going to do that.

In the discussion before this one, which you marked as resolved you wrote:

> The idea is to find the item that needs to be loaded for a given column. When multiple items are selected it will show nothing on macOS. I will make changes to make this code behave similarly.

I don't see in the code how the case of multiple selection gets handled in this way. This is a minor issue, so merging was fine, still it would have been nice to get a real answer to this question.

gcasa:
> > As you can see from the above it does seem to function properly. The _itemForColumn: method gets the list for the current column based on the previous column's selection. So given this, I believe it is okay to merge. I am going to do that.
> 
> 
> 
> In the discussion before this one, which you marked as resolved you wrote:
> 
> 
> 
> > The idea is to find the item that needs to be loaded for a given column. When multiple items are selected it will show nothing on macOS. I will make changes to make this code behave similarly.
> 
> 
> 
> I don't see in the code how the case of multiple selection gets handled in this way. This is a minor issue, so merging was fine, still it would have been nice to get a real answer to this question.

I apologize. If you run the test and select two things in a column the column next to it shows nothing. This is the correct behavior. I will take a look and provide a better explanation of this. I am not in front of my computer right now.

Source/NSBrowser.m
```diff
@@ -407,15 +415,15 @@ - (NSInteger) selectedRowInColumn: (NSInteger)column
...
    -browser:selectRow:inColumn: is implemented, this is its responsability
    to select the cell. This method adds a NSBrowser column if needed
-   and deselects other selections if the browser does not allows multiple 
-   selection.</p><p>See Also: -loadedCellAtRow:column: 
+   and deselects other selections if the browser does not allows multiple
+   selection.</p><p>See Also: -loadedCellAtRow:column:
    -browser:selectRow:inColumn: [NSMatrix-selectCellAtRow:column:]</p>
  */
-- (void) selectRow: (NSInteger)row inColumn: (NSInteger)column 
+- (void) selectRow: (NSInteger)row inColumn: (NSInteger)column
```
gcasa, line 426: @fredkiefer I believe this is the method where a multiple selection causes the last column to be unloaded.  The method you commented on earlier is responsible for getting the selection of the previous column so that the current column knows what to display (based on the children it retrieves).

PR #282, opened 2024-08-08
Modal display not on top for some platforms

williameveretteggplant:
The problem we have seen is that on some platforms (such as Redhat 8) modal windows, specifically NSAlert, are not reliably appearing above the other windows.

By moving the `orderFrontRegardless` call to a little later in the `beginModalSessionForWindow:` method, this appears to reliably display on top as expected.

### DIFF
--- a/Source/NSApplication.m
+++ b/Source/NSApplication.m
@@ -1785,6 +1785,7 @@ static BOOL _isAutolaunchChecked = NO;
 	  /*
 	   * Try to handle events for this session, discarding others.
 	   */
+    	  [theSession->window orderFrontRegardless];
 	  code = [self runModalSession: theSession];
 	  if (code == NSRunContinuesResponse)
 	    {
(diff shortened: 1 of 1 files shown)


Source/NSApplication.m
```diff
@@ -1689,6 +1689,7 @@ - (NSModalSession) beginModalSessionForWindow: (NSWindow*)theWindow
 	  [theWindow makeMainWindow];
 	}
     }
+  [theWindow orderFrontRegardless];
```
gcasa, line 1692: This change looks reasonable to me. Any thoughts @fredkiefer ?

fredkiefer: The code change looks OK. I have not tested the behaviour but if you have feel free to merge.

gcasa: I have tested this on Debian 12 and it works properly there.

gcasa: @fredkiefer per your comment, I am approving this so that @williameveretteggplant can merge.

gcasa (approved):
@williameveretteggplant Please go ahead and merge.

williameveretteggplant:
I don't see any way to merge.

gcasa:
> I don't see any way to merge.

This is one of the reasons I advised not creating forks and accepting my invitation instead.  If it was a branch on the repo you would have the option to merge.

I will do it for you.  One sec.

PR #289, opened 2024-09-12
GSMemoryPanel: Only call GSDebug* in debug mode

qmfrederik:
The GSDebug* methods are conditional on NDEBUG not being defined; so do not attempt to call these methods if this variable is not set.

The current implementation causes a build warning/error:

```
error: call to undeclared function 'GSDebugAllocationActive'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
```

### DIFF
--- a/Source/GSMemoryPanel.m
+++ b/Source/GSMemoryPanel.m
@@ -191,7 +191,9 @@ static GSMemoryPanel *sharedGSMemoryPanel = nil;
   NSButton *button;
 
   /* Activate debugging of allocation. */
+#ifndef	NDEBUG
   GSDebugAllocationActive (YES);
+#endif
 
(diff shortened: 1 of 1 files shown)


fredkiefer:
The problem is that these functions are defined in base and used here in gui. There is no way that we can guarantee that both libraries are compiled with matching flags. Still in some cases your change hopefully makes the setup more consistent.

fredkiefer (approved):

rfm:
Perhaps this should be considered a bug in base ... I don't think the presence of the functions should be determined by NDEBUG

qmfrederik:
Perhaps the implementation in base should be a no-op if `NDEBUG` is set, rather than the functions being undefined?

PR #290, opened 2024-09-14
Fall back to pkgconfig for resolving libtiff

qmfrederik:
In certain configurations (e.g. when built in debug mode using the msvc compiler), the libtiff link library may have a 'd' suffix -- 'tiffd.lib' instead of `tiff.lib`. This is an explicit choice in the libtiff build process: https://gitlab.com/libtiff/libtiff/-/blob/master/cmake/WindowsSupport.cmake#L28

In this case, `AC_CHECK_LIB(tiff, TIFFReadScanline)` will fail because it doesn't search for `tiffd`.

Rather than hardcoding the `tiffd` variant, use pkg-config to find libtiff.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -273,7 +273,16 @@ if test "$have_tiff" = yes; then
     GRAPHIC_LFLAGS="$with_tiff_library $GRAPHIC_LFLAGS"
     GRAPHIC_CFLAGS="$with_tiff_include $GRAPHIC_CFLAGS"
   else
-    have_tiff=no
+    PKG_CHECK_MODULES(
+      [TIFF],
+      [libtiff-4],
+      [
+        GRAPHIC_LFLAGS="$TIFF_LIBS $GRAPHIC_LFLAGS"
+        GRAPHIC_CFLAGS="$TIFF_CFLAGS $GRAPHIC_CFLAGS"
+      ],
+      [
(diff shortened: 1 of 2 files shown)


qmfrederik:
See also https://github.com/gnustep/tools-windows-msvc/blob/master/phases/43-libtiff.bat#L34-L36 - another workaround for the same problem.

configure.ac
```diff
@@ -273,7 +273,16 @@ if test "$have_tiff" = yes; then
     GRAPHIC_LFLAGS="$with_tiff_library $GRAPHIC_LFLAGS"
     GRAPHIC_CFLAGS="$with_tiff_include $GRAPHIC_CFLAGS"
   else
-    have_tiff=no
+    PKG_CHECK_MODULES(
+      [TIFF],
+      [libtiff-4],
+      [
+        GRAPHIC_CFLAGS="$TIFF_CFLAGS $GRAPHIC_CFLAGS"
```
fredkiefer, line 280: Why did you switch this two lines?

qmfrederik: I copy/pasted the code which used `PKG_CHECK_MODULES` to search for ImageMagick, and that had `LDFLAGS` and `CFLAGS` in this order.  Good catch; I amended the commit to keep the order consistent with the previous lines.

Issue #291, opened 2024-09-16
NSBox doesn't decode properly on 64bit big endian


rmottola:
in several Gorm files of GWorkspace, e.g. Attributes.gorm on PPC 64bit and SPARC 64bit boxes do not show their borders.
PPC 32bit works.

fredkiefer:
Thank you for your PR

rmottola:
Confirm fixed on SPARC and PPC 64bit.

PR #292, opened 2024-09-18
use macro to decode NSInteger into enum - fixes 64Bit Big Endian

rmottola:
Fix enum decoding issue of https://github.com/gnustep/libs-gui/issues/291


rmottola:
Could there be other lingering undetected? I guess the issue is signed vs. unsigned enum, which is significative on 64Bit BE... on 32bit BE or 64BIT LE it doesn't caused issues.

fredkiefer (approved):

PR #296, opened 2024-09-24
implement integerValue method in NSSliderCell

williameveretteggplant:
When you call setIntegerValue: method on an NSSliderCell, it calls the superclass method in NSCell, which then calls setObjectValue:. The implementation of setObjectValue: in NSSliderCell sets the _value ivar. The problem is, when you call integerValue: it calls the superclass method in NSCell, which returns a value based on the _object_value ivar, which has not been set.
By implementing the integerValue: method directly in NSSliderCell, this will return the correct value.

### DIFF
--- a/Source/NSSliderCell.m
+++ b/Source/NSSliderCell.m
@@ -658,6 +658,11 @@ double _doubleValueForMousePoint (NSPoint point, NSRect knobRect,
     }
 }
 
+- (NSInteger) integerValue
+{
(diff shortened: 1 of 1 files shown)


fredkiefer (approved):
It would be more consistent if you also add an `setIntegerValue:` method. I wait a while with merging to leave you some time for that.

fredkiefer:
Sadly it did not happen so I merge without it.

Issue #295, opened 2024-09-23
NSAttributedString rangeOfTextList:atIndex: locks up on adjacent lists


fjmilens3:
The current `rangeOfTextList:atIndex:` implementation will lock up when two paragraphs with different styles and different textlists are adjacent. I found this while looking into #294 (the issue is far more likely to surface when #294 is fixed because attribute ranges aren't incorrectly merged). A fix for #293 would also likely depend on this being fixed. 

The following example will reproduce the issue:

```
- (void) causeRangeOfTextListProblem 
{
  // Create two lists 
  NSTextList *list1 = [[NSTextList alloc] initWithMarkerFormat: @"{decimal}" options: 0];
  NSTextList *list2 = [[NSTextList alloc] initWithMarkerFormat: @"{box}" options: 0];
  
  // Create two paragraph styles
  NSMutableParagraphStyle *style1 = [[NSParagraphStyle defaultParagraphStyle] mutableCopy];
  NSMutableParagraphStyle *style2 = [[NSParagraphStyle defaultParagraphStyle] mutableCopy];
  
  // Change the paragraph styles so that they don't get merged together (equality testing problems)
  [style1 setHeadIndent: 36.0];
  [style2 setHeadIndent: 72.0];
  
  // Use one list per paragraph style
  [style1 setTextLists: [NSArray arrayWithObject: list1]];
  [style2 setTextLists: [NSArray arrayWithObject: list2]];

  // Create an attributed string for each list
  NSAttributedString *str1 = [[NSAttributedString alloc] 
    initWithString: [NSString stringWithFormat: @"%@ list1\n", [list1 markerForItemNumber: 1]] 
    attributes: [NSDictionary dictionaryWithObject: style1 forKey: NSParagraphStyleAttributeName]];
  NSAttributedString *str2 = [[NSAttributedString alloc] 
    initWithString: [NSString stringWithFormat: @"%@ list2\n", [list2 markerForItemNumber: 1]] 
    attributes: [NSDictionary dictionaryWithObject: style2 forKey: NSParagraphStyleAttributeName]];

  // Append the attributed strings to the text storage
  NSTextStorage *textStorage = [textView textStorage];
  [textStorage appendAttributedString: str1];
  [textStorage appendAttributedString: str2];
  
  // Attempt to find the list range for one of the two adjacent lists
  // This causes a problem and the call will not return
  NSLog(@"Preparing to find list range...");
  
[...]

Issue #294, opened 2024-09-23
Problems with NSTextList and NSParagraphStyle isEqual: implementations


fjmilens3:
The implementation of `isEqual:` for NSTextList and NSParagraphStyle needs to be reconsidered. As a practical matter it causes problems with list handling in NSTextViews/NSTextStorage. 

(I'd also be happy to submit some PRs on these if nobody else is interested, but at this time, I've just started playing with GNUStep and don't have it building locally. The three issues I'm reporting are all interrelated to some degree in terms of their effects.)

### NSTextList

For NSTextList on GNUStep, we do the following and compare some of the member fields:

https://github.com/gnustep/libs-gui/blob/fb61a0f287f545f4b3dc3aa7c5244572ba016d33/Source/NSTextList.m#L54-L67

On the other hand, Cocoa itself seems to treat each NSTextList as unique, even if they have the same values:

```
NSTextList *list1 = [[NSTextList alloc] initWithMarkerFormat: @"{box}" options:0];
NSTextList *list2 = [[NSTextList alloc] initWithMarkerFormat: @"{box}" options:0];

NSLog(@"Equals: %d", [list1 isEqual: list2]);
```

On Cocoa this prints 0 but on GNUStep it prints 1. I suspect the Cocoa implementation's `isEqual:` only looks to see if they're the exact same object, which makes sense. Two lists might have the same values (marker format, starting value) but they would still be fundamentally different lists. (This also relates to the below topic.)

### NSParagraphStyle

For NSParagraphStyle it currently doesn't include the `textLists` as part of the equality test:

https://github.com/gnustep/libs-gui/blob/fb61a0f287f545f4b3dc3aa7c5244572ba016d33/Source/NSParagraphStyle.m#L597-L625

That appears to be causing a problem where NSParagraphStyle runs in attributed strings get merged together when they shouldn't. In this case it can happen when adjacent paragraph styles differing only by lists end up getting combined when they shouldn't, but it could conceivably happen in other situations as well (text blocks perhaps?).

Note that adding the `textLists` to the comparison here won't have any effect unless the NSTextList comparison itself is fixed to treat each instance as unique.

fjmilens3:
To see what I mean about it causing paragraph styles to be merged together when they shouldn't, see the following:

```
- (void) paragraphMergeExample
{
  // Create two lists (even with different parameters for emphasis)
  NSTextList *list1 = [[NSTextList alloc] initWithMarkerFormat: @"{box}" options: 0];
[...]

Issue #293, opened 2024-09-23
Behavior difference in NSAttributeString itemNumberInTextList:atIndex:


fjmilens3:
Per Apple documentation, NSAttributeString's `itemNumberInTextList:atIndex:` should [return the index of the item at the specified location within the list](https://developer.apple.com/documentation/foundation/nsattributedstring/1527303-itemnumberintextlist?language=objc).

The current GNUStep implementation instead returns the index position _of the list_ within the lists for the current paragraph style. This gives you the nesting level instead of the item number.

https://github.com/gnustep/libs-gui/blob/fb61a0f287f545f4b3dc3aa7c5244572ba016d33/Source/NSAttributedString.m#L1095-L1112

For an example, consider the following list:

- Alpha
- Beta
- Gamma
- Delta
- Epsilon

For a text position anywhere in, say, "Delta," Apple's implementation would return 4 as it's the fourth item in the list. The current GNUstep implementation returns 0 as the list is the zeroth item in the paragraph style's lists array.

I've been playing around with this locally but I'd encourage someone else with access to a Mac to verify the difference as well. Note that if you write a test program of your own with Cocoa on recent Macs, you'll probably want to [opt-out of TextKit 2 behavior](https://indiestack.com/2022/11/opting-out-of-textkit2-in-nstextview/) via `[textView layoutManager]` for an equal comparison to GNUStep. 

If it turns out to be a valid bug, some thoughts based on my current, hacky workaround: Get the entire range of the list from `rangeOfTextList:atIndex:`, iterate over the range looking for newline characters, and basically count paragraphs until it reaches the position we were asked about. You'd also want to check against the lists in each paragraph's style to account for things like nested lists.

fjmilens3:
For those who don't have access to a Mac, see below.

Cocoa example code for staging list data:

```
    // Do this to revert to TextKit 1 instead of TextKit 2
    NSLayoutManager *manager = [_textView layoutManager];
    
    NSTextList *list = [[NSTextList alloc] initWithMarkerFormat:@"{decimal}" options:0];
    
    NSMutableParagraphStyle *style = [[NSParagraphStyle defaultParagraphStyle] mutableCopy];
    
    [style setTextLists: [NSArray arrayWithObject: list]];
    
    NSMutableString *str = [[NSMutableString alloc] init];
    for (int i = 0; i < 10; i++) {
        [str appendFormat: @"%@ item\n", [list markerForItemNumber: i]];
    }
    
[...]

PR #297, opened 2024-09-26
Improvements and fixes to NSTextList-related code

fjmilens3:
Fixes for some NSTextList-related issues (#293, #294, #295). 

Changes include:

- Changes to `isEqual:` behavior for NSTextList and NSParagraphStyle.
- Includes a default empty array value for NSParagraphStyle `_textLists` (similar to `_tabStops`).
- Fixes NSAttributeString's `rangeOfTextList:atIndex:` to avoid lockup conditions for adjacent ranges.
- Fixes NSAttributeString's `itemNumberInTextList:atIndex:` to return the item number for the list.
- Includes test cases for the above.

Notes:

- I only installed GNUStep last week and don't have a full setup for building and working on GNUStep itself yet. I was able to build libs-gui with some opt-outs. If I keep running into problems, is the gnustep-discuss list the best place to ask for help?
- In addition to the test cases included here, I've tested the code by including it in category methods within my existing GNUStep installation. They seem to resolve the problems that led me to report the bugs.
- This is my first PR so if there are any general GNUStep code practices I'm missing, let me know and I'll adjust accordingly.

### DIFF
--- a/Source/NSAttributedString.m
+++ b/Source/NSAttributedString.m
@@ -1110,7 +1110,6 @@ documentAttributes: (NSDictionary **)dict
   
   NSCharacterSet *newlineCharacterSet = [NSCharacterSet newlineCharacterSet];
   
-  BOOL isNested = NO;
   NSUInteger itemNumber = 1;
   NSUInteger index;
   for (index=1; index<length; index++)
@@ -1125,12 +1124,6 @@ documentAttributes: (NSDictionary **)dict
           if ([list isEqual: [textLists lastObject]])
             {
               itemNumber++;
-              isNested = NO;
-            }
-          else if (!isNested && [textLists containsObject: list]) 
-            {
-              itemNumber++;
-              isNested = YES;
             }
           
           if (buffer[index-1] == '\r' && buffer[index] == '\n')
--- a/Source/NSAttributedString.m
+++ b/Source/NSAttributedString.m
@@ -53,6 +53,7 @@
 #import "AppKit/NSFont.h"
 #import "AppKit/NSFontDescriptor.h"
 #import "AppKit/NSFontManager.h"
+#import "AppKit/NSTextList.h"
 // For the colour name spaces
 #import "AppKit/NSGraphics.h"
 #import "AppKit/NSTextTable.h"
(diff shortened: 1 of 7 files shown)


Source/NSAttributedString.m
```diff
@@ -1095,20 +1096,43 @@ - (NSFileWrapper *) fileWrapperFromRange: (NSRange)range
...
 {
-  NSParagraphStyle *style = [self attribute: NSParagraphStyleAttributeName
-                                  atIndex: location
-                                  effectiveRange: NULL];
-  if (style != nil)
+  NSRange listRange = [self rangeOfTextList: list atIndex: location];
+  if (listRange.location == NSNotFound)
     {
-      NSArray *textLists = [style textLists];
+      return 0;
```
fjmilens3, line 1102: This matches the return value in Cocoa (I'd have expected `NSNotFound` but it seems to return 0 instead).

Source/NSAttributedString.m
```diff
@@ -1095,20 +1096,43 @@ - (NSFileWrapper *) fileWrapperFromRange: (NSRange)range
...
+            atIndex: subRange.location + index
+            effectiveRange: NULL];
+          
+          NSArray *textLists = [style textLists];
+          if ([list isEqual: [textLists lastObject]])
+            {
+              itemNumber++;
+            }
+          
+          if (buffer[index-1] == '\r' && buffer[index] == '\n')
```
fjmilens3, line 1128: This and the use of `NSCharacterSet newlineCharacterSet` may be overkill, would we use '\n' internally or do we need to support Windows CR+LF combinations?

Source/NSParagraphStyle.m
```diff
@@ -303,6 +303,8 @@ - (id) init
           [_tabStops addObject: tab];
           RELEASE(tab);
         }
+      
+      ASSIGN(_textLists, [NSArray array]);
```
fjmilens3, line 307: Instances in Cocoa always have an empty array set for the default value, so I included this here and below.

We weren't even saving the textLists in some of the serialization paths, and other fields like text blocks and text tables are ignored as well. I'd suggest deferring any work on that part unless we make the changes as a batch and bump the object version; I'm not that familiar with any of this yet, and I don't want the PR to get out of hand.

fredkiefer (approved):
Sorry for not looking into this PR before. I must have forgotten about it. The changes look good and there are a lot of tests. Thank you!

PR #298, opened 2024-10-02
NSView (autoresize): simpler implementation

trunkmaster:
Change to a simpler and clearer implementation. As a side effect this change fixes Font Panel resize handling.

### DIFF
--- a/Source/NSView.m
+++ b/Source/NSView.m
@@ -1990,61 +1990,33 @@ static void autoresize(CGFloat oldContainerSize,
 		       BOOL maxMarginFlexible)
 {
   const CGFloat change = newContainerSize - oldContainerSize;
-  const CGFloat oldContentSize = *contentSizeInOut;
-  const CGFloat oldContentPosition = *contentPositionInOut;
-  CGFloat flexibleSpace = 0.0;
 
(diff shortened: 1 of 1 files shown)


Source/NSView.m
```diff
@@ -1990,61 +1990,33 @@ static void autoresize(CGFloat oldContainerSize,
...
-	  
-	  if (sizeFlexible)
-	    { 
-	      *contentSizeInOut += changePerOption;
-	    }
-	  if (minMarginFlexible)
-	    {
-	      *contentPositionInOut += changePerOption;
-	    }
+	  *contentSizeInOut += (change / 2);
```
fredkiefer, line 2002: Shouldn't we divide by 3 if min and max are flexible?

trunkmaster:
Thanks for the hint. It seemes that I've made fix to the wrong place. All autoresizing tests fail.

My change introduces different algorithm of autoresizing - it resizes view not proportionally but keeps views aligned to each other without AutoLayout magic. That's why it fixes Font Panel resizing. However it doesn't conform to Cocoa behaviour (checked on macOS and OPENSTEP).

I'll close this PR and make new with Font Panel fix.

PR #299, opened 2024-10-04
Font Panel autosizing fix

trunkmaster:
Fix autosizing of "Family" and "Typeface" browsers using different approach to NSView's default.

### DIFF
--- a/Source/NSFontPanel.m
+++ b/Source/NSFontPanel.m
@@ -84,6 +84,111 @@ static float sizes[] = {4.0, 6.0, 8.0, 9.0, 10.0, 11.0, 12.0, 13.0,
 }
 @end
 
+@interface FPBrowser : NSBrowser
+@end
+@implementation FPBrowser
+
+static void
+autoresize(CGFloat oldContainerSize, CGFloat newContainerSize,
+	   CGFloat *contentPositionInOut, CGFloat *contentSizeInOut,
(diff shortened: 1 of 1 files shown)


fredkiefer:
I am not sure whether all this is needed or if it could be sufficient to set the correct resizing attributes on the existing views. What is it that you want to achieve?

trunkmaster:
Sizing attributes are correct. Look at screenshots below, can you see the difference?

Initial panel size.
VirtualBox_Fedora 39_05_10_2024_02_39_46

Resized without my fix
VirtualBox_Fedora 39_05_10_2024_02_55_03

Resized with my fix
VirtualBox_Fedora 39_05_10_2024_02_43_26

fredkiefer:
Could you please try how the new resizing behaves when there is an accessoryView? This is not very common, we still need to support it.

trunkmaster:
My change doesn't affect accessoryView, only browsers: familyBrowser, typefaceBrowser and sizeBrowser which are inside bottomSplit view.

However I can add and test accessoryView. What results do you expect from me: screenshots or something else?

fredkiefer:
I don't need any proof, it is sufficient if you test it yourself.

fredkiefer (approved):

trunkmaster:
Looks good.
VirtualBox_Fedora 39_09_10_2024_01_15_24
VirtualBox_Fedora 39_09_10_2024_01_17_21

PR #300, opened 2024-10-07
Fix Issue with NSCell flags not getting set properly

johnathan-becker:
This PR fixes an issue where NSCellType (enum) and NSLineBreakMode (struct) were not being properly respected in MSYS2 as bit field members thus making a 32bit union 44 bytes.

### DIFF
--- a/Source/GSCodingFlags.h
+++ b/Source/GSCodingFlags.h
@@ -42,7 +42,7 @@ typedef struct _GSCellFlags {
   unsigned int        disabled:1;
   unsigned int        editable:1;
 
-  NSCellType          type:2;
+  unsigned int        type:2;
   unsigned int        vCentered:1;
   unsigned int        hCentered:1;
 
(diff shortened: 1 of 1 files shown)


gcasa (commented):
I may need to test these to make sure that older .gorm files are not effected by this change as it could alter the bit layout that is read/written.

johnathan-becker:
> I may need to test these to make sure that older .gorm files are not effected by this change as it could alter the bit layout that is read/written.

That is no problem. I do not expect any problems but please feel free to run test. Better safe than sorry.

gcasa:
> > I may need to test these to make sure that older .gorm files are not effected by this change as it could alter the bit layout that is read/written.
> 
> That is no problem. I do not expect any problems but please feel free to run test. Better safe than sorry.

Looks okay.   I ran a few manual tests.   Thanks.

PR #302, opened 2024-10-09
NSApplication: toggle title of menu item "Hide" on user action.

trunkmaster:
Toggle title of menu item "Hide" to "Unhide" and vice versa.
Right-click app menu on hidden app icon should show "Unhide" on menu item and behave respectively.

### DIFF
--- a/Source/NSApplication.m
+++ b/Source/NSApplication.m
@@ -2496,6 +2496,7 @@ image.</p><p>See Also: -applicationIconImage</p>
 	  NSDictionary  	*info;
 	  NSWindow		*win;
 	  NSEnumerator  	*iter;
+          id<NSMenuItem>  	menuItem;
 
 	  [nc postNotificationName: NSApplicationWillHideNotification
 	                    object: self];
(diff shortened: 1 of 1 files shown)


Source/NSApplication.m
```diff
@@ -2590,6 +2591,10 @@ - (void) hide: (id)sender
 	  [self miniaturizeAll: sender];
 	}
     }
+  else
```
fredkiefer, line 2594: Is this part still needed? And if so why? Maybe we need also to change the action assigned to the menu item? Or hav just one method behind hide: and unhide:?

trunkmaster: Do you think it would be clearer to create one more method instead of `else {}` block inside hide:? 
Do you think that mixing hide/unhide code in one method is a good idea?

trunkmaster: @fredkiefer reminder. Probably you've missed my comment.

And again: my patch fixes bug - right-click menu item doesn't unhide aplication. Maybe we should postpone refactoring until you'll come up with right architectural decision?

fredkiefer: No, I rather think we should fix it as good as possible and not just add another workaround that we will regret later on.

First, your solution won't work for non english applications. We need to use _(@"Hide") and (@"Show"), not Unhide, to get it working for other languages. We also should the action of the menu item not the behaviour of the method itself. That method will be called from other contexts as well, even from other applications and if we change its behaviour depending on the application state, this might confuse things.
As to where the updating of the menu item should happen I am unsure. The -hide: and -unhide: methods are surely one option. But what about -unhideWithoutActivation?

trunkmaster: Now I understand your concerns - thanks for explanation.

I think I've found a solution. The idea is to change action (setAction:) and title (setTitle) of menu item. No new methods were added and it should work on menu item click as well as by directly calling -hide: and unhide: methods.

fredkiefer (approved):
I am still unsure whether the name should be Unhide or Show. If you insist on Unhide, then we should change all the places where Show is currently used.

trunkmaster:
"Show" is OK - fixed.

PR #303, opened 2024-10-10
Add Tests for GSCodingFlags

johnathan-becker:
Due to recent changes around GSCodingFlags.h I have added a test that makes sure it works in the future

### DIFF
--- a/Tests/gui/GSCodingFlags/GSCellFlags.m
+++ b/Tests/gui/GSCodingFlags/GSCellFlags.m
@@ -42,9 +42,9 @@ int main()
     mask.flags = (GSCellFlags){0};
     // now make sure values translate to flags
 #if GS_WORDS_BIGENDIAN == 1
-    mask.value = 0b00000001001000000000110100000001;
+    mask.value = 0b00010010000000000001110000001001;
 #else
-    mask.value = 0b00000001001000000000110100000001;
+    mask.value = 0b10010000001110000000000001001000;
 #endif
 
     pass(mask.flags.state = 1, "state is correctly set");
--- a/Tests/gui/GSCodingFlags/GSCellFlags.m
+++ b/Tests/gui/GSCodingFlags/GSCellFlags.m
@@ -36,13 +36,13 @@ int main()
     mask.value = 0b10010000001110000000000001001000;
 #endif
 
-    pass(mask.flags.state = 1, "state is correctly set");
-    pass(mask.flags.selectable = 1, "selectable is correctly set");
-    pass(mask.flags.scrollable = 1, "scrollable is correctly set");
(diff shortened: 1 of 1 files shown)


Tests/gui/GSCodingFlags/GSCellFlags.m
```diff
@@ -0,0 +1,62 @@
...
+    pass(mask.value == 0b00010010000000000001110000001001, "mask.flags translates to mask.value");
+#else
+    pass(mask.value == 0b10010000001110000000000001001000, "mask.flags translates to mask.value");
+#endif
+    // reset mask
+    mask.value = 0;
+    mask.flags = (GSCellFlags){0};
+    // now make sure values translate to flags
+#if GS_WORDS_BIGENDIAN == 1
+    mask.value = 0b00000001001000000000110100000001;
```
fredkiefer, line 45: Both values look wrong. I would expect that they are different and identical to the ones above

Tests/gui/GSCodingFlags/GSCellFlags.m
```diff
@@ -0,0 +1,62 @@
...
+#include "GSCodingFlags.h"
+
+int main()
+{
+    CREATE_AUTORELEASE_POOL(arp);
+    GSCellFlagsUnion mask = { { 0 } };
+
+    START_SET("GSCodingFlags GNUstep CellFlags Union")
+    
+    NS_DURING
```
fredkiefer, line 16: I don't think that we need this check for this test. Nothing that would require a backend is used here.

Tests/gui/GSCodingFlags/GSCellFlags.m
```diff
@@ -0,0 +1,51 @@
...
+    mask.value = 0;
+    mask.flags = (GSCellFlags){0};
+    // now make sure values translate to flags
+#if GS_WORDS_BIGENDIAN == 1
+    mask.value = 0b00000001001000000000110100000001;
+#else
+    mask.value = 0b00000001001000000000110100000001;
+#endif
+
+    pass(mask.flags.state = 1, "state is correctly set");
```
fredkiefer, line 39: You are assigning the value 1 here not testing it. The same is true for the other checks. This explains why you have seven checks but only six bits that are set to one.

johnathan-becker: I see that aplogies. I have been out of a town for a few days but I did just update the test and I ran it locally and it passed, although I do not have access to a big endian system.

fredkiefer: Now the values are plausible, but the checks are still wrong. Could you please fix that? '=' => '=='

fredkiefer (approved):
Now it looks good

PR #307, opened 2024-10-27
Allow themes to override `[NSMenu _isVisible]`

qmfrederik:
The default implementation of `[NSMenu]` uses `_aWindow` and `_bWindow` to determine the visibility of the menu.  Not all themes use these windows to build a menu.  For example, the WinUXTheme will use Win32 APIs to build a menu.

Allow themes to override the value of `[NSMenu _isVisible]` by introducing a `[GSTheme proposedVisibility: (BOOL)visible forMenu: (NSMenu *) menu]` method.  The default implementation simply returns the proposed visibility.

### DIFF
--- a/Headers/Additions/GNUstepGUI/GSTheme.h
+++ b/Headers/Additions/GNUstepGUI/GSTheme.h
@@ -1569,6 +1569,12 @@ withRepeatedImage: (NSImage*)image
 - (void) organizeMenu: (NSMenu *)menu
 	 isHorizontal: (BOOL)horizontal;
 
+/**
+ * Used by the theme to override the proposed menu visibility.  The default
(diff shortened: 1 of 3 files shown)


qmfrederik:
@fredkiefer This is another approach for #304.  The default implementation for `[NSMenu _isVisible]`  is `[_aWindow isVisible] || [_bWindow isVisible]`, but that doesn't make sense when using the WinUXTheme which doesn't use these windows.

Instead, I suggest we allow the theme to override the proposed window visibility.

fredkiefer (approved):
The intention is good, but the code looks a bit ugly. As I don't have a better idea at the moment, let's merge this and think about it later again.

PR #310, opened 2024-10-28
NSTextFieldCell: Decode placeholder string

qmfrederik:
Support decoding the `placeholderString`  value from Xib files. For example, consider the following snippet:

```xml
<textFieldCell [...] placeholderString="Default">
</textFieldCell>
```

In the current implementation, the value of `placeholderString` was not being deserialized.


qmfrederik:
@fredkiefer This looks pretty straightforward, any objections from your side?

fredkiefer (approved):

PR #312, opened 2024-11-01
Make white as default selection highlight color for NSTableView

trunkmaster:
Currently default highlight color is light blue which is not correct for default theme.

### DIFF
--- a/Source/GSThemeDrawing.m
+++ b/Source/GSThemeDrawing.m
@@ -3319,9 +3319,9 @@ static NSDictionary *titleTextAttributes[3] = {nil, nil, nil};
 
     if (selectionColor == nil)
       {
-	selectionColor = [NSColor colorWithCalibratedRed: 0.86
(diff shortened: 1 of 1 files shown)


fredkiefer (approved):
For the default theme we should also add an entry in the default color map we provide.

trunkmaster:
> For the default theme we should also add an entry in the default color map we provide.

Where does that color map reside? I can't find it.

fredkiefer:
That is the problem, we only have a list of colours in the file NSColor.m. There we build up the predefined colours and store them in a new list, or add them to an already existing list. What we should do is distribute a file System.clr with colours. We have such files in themes, but not for the default theme. The reason is, that way it is easier to extend the list, whenever we need to define a new colour. This comment was rather directed at myself that I need to rectify that situation and provide a clean solution. Or perhaps I should pass on this problem to somebody who will actually resolve it?

PR #314, opened 2024-11-04
modify NSComboBox to handle attributed strings

williameveretteggplant:
Currently NSComboBox when provided with an attributed string will display the description of an attributed string rather than the string value. The mac implementation displays the string value. This brings it into alignment with that implementation.

### DIFF
--- a/Source/NSComboBoxCell.m
+++ b/Source/NSComboBoxCell.m
@@ -1775,23 +1775,39 @@ static inline NSRect buttonCellFrameFromRect(NSRect cellRect)
     }
   else
     {
+      id object = nil;
+      
       if (_dataSource == nil)
         {
-	  NSLog(@"%@: No data source currently specified", self);
-	  return nil;
-	}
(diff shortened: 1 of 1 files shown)


Source/NSComboBoxCell.m
```diff
@@ -1775,23 +1775,39 @@ - (NSString *) _stringValueAtIndex: (NSInteger)index
...
-	  return [[_dataSource comboBoxCell: self
-			      objectValueForItemAtIndex: index] description];
-	}
+          object = [_dataSource comboBoxCell: self objectValueForItemAtIndex: index];
+        }
+      
+      if (object)
+      {
+        // Check for attributed string type and return actual string instead..
+        if ([object isKindOfClass :[NSAttributedString class]])
```
fredkiefer, line 1799: Could you please move the colon to the method name, not the argument?

williameveretteggplant: done

fredkiefer (approved):

PR #315, opened 2024-11-07
Fix issue with nsmenu items blocking the GUI

johnathan-becker:
This MR fixes an issue where when you click item an NSMenuView item you have a frozen UI until you double click a second time, ending the loop.

### DIFF
--- a/Source/NSMenuView.m
+++ b/Source/NSMenuView.m
@@ -1832,6 +1832,11 @@ static float menuBarHeight = 0.0;
 					inMode: NSEventTrackingRunLoopMode
 				       dequeue: YES];
 	  type = [event type];
+    if (type == NSLeftMouseUp || type == NSRightMouseUp || type == NSOtherMouseUp)
+      {
+          shouldFinish = YES;
+          break;  // Exit the loop to proceed to StopPeriodicEvents
+      }
 	  if (type == NSAppKitDefined)
 	    {
 	      [[event window] sendEvent: event];


Source/NSMenuView.m
```diff
@@ -1545,6 +1545,7 @@ - (BOOL) _trackWithEvent: (NSEvent*)event
   BOOL justAttachedNewSubmenu = NO;
   BOOL subMenusNeedRemoving = YES;
   BOOL shouldFinish = YES;
+  BOOL shouldStop = NO;
```
fredkiefer, line 1548: Where does this value ever get used? You seem to only set it, but never read it.

Source/NSMenuView.m
```diff
@@ -1826,18 +1827,24 @@ - (BOOL) _trackWithEvent: (NSEvent*)event
...
-					inMode: NSEventTrackingRunLoopMode
-				       dequeue: YES];
-	  type = [event type];
-	  if (type == NSAppKitDefined)
-	    {
-	      [[event window] sendEvent: event];
-	    }
-	}
-      while (type == NSAppKitDefined);
+      {
```
fredkiefer, line 1830: Could you please stick with the current GNUstep indentation. I know that it is annoying to get it right, but we agreed on it a long time ago and now we should keep the code that way.

fredkiefer (changes requested):
Looks interesting, I will need to test it myself.

Source/NSMenuView.m
johnathan-becker: Sorry I have fixed this now.

Source/NSMenuView.m
johnathan-becker: I have removed that value. It was an artifact from debugging apologies.

johnathan-becker:
@fredkiefer I am not sure why the pipelines are failing I am getting the message `It looks like GNUstep backend is not yet installed`

fredkiefer:
Skipping most of the tests is normal in the pipeline. Maybe we should add a step to download and compile back here?
But your error is another one: 
```
gui/NSBezierPath/basic.m:
Failed file:  basic.m aborted without running all tests!

```
But then I have no idea, how your change could affect that.

fredkiefer (approved):

fredkiefer:
I found and fixed the reason why the NSBezierPath test failed. If you rebase it should be fixed for you as well.

johnathan-becker:
@fredkiefer I have rebased and all checks have passed now! Thank you.

Issue #287, opened 2024-08-29
ICU detection


rmottola:
on debian 12, icu doesn't install icu-config

base detects ICU fine:
`checking for ICU... yes
`

however GUI fails to detect it:

```
checking for icu-config... no
Could not find the icu-config script.
Please ensure that it is in your path.
See http://site.icu-project.org/ for help.
no
configure: WARNING: The International Components for Unicode (ICU) development h                              
```

PR #320, opened 2024-11-18
simplify ICU configure, remove icu-config parameter and cascade with …

rmottola:
It was odd to have an extra option which was more a workaround, cascade tests. Import base test and match ICU version to it. Especially PKG_PROG_PKG_CONFIG to have pkg-config available as we do in base.

Should fix https://github.com/gnustep/libs-gui/issues/287


fredkiefer (approved):
There a few strange changes to the generated configure file, which I don't understand. But this is something we cannot control ourselves.

PR #321, opened 2024-11-19
GSColorSliderCell: Move to libs-gui

qmfrederik:
This class is shared between the StandardPicker bundle and the WheelPicker bundles.  Exposing a class with the same name in two separate bundles results in undefined behavior.  Move this class to a shared module (libs-gui in this case) to resolve that.


qmfrederik:
@fredkiefer This fixes an issue when using libs-gui with libobjc2, see https://github.com/gnustep/libobjc2/pull/318 for more details.

PR #322, opened 2024-11-27
Fix issues with non-keyed encoding/decoding of NSPopUpButton

gcasa:
The pull down functionality of the NSPopUpButton appears to be broken when creating a gorm file.   This doesn't seem to be a problem when decoding a XIB file so.  I have narrowed it down to the non keyed encoding logic and incomplete support and set up in Gorm when creating the widget.

### DIFF
--- a/Source/NSPopUpButtonCell.m
+++ b/Source/NSPopUpButtonCell.m
@@ -654,7 +654,6 @@ static NSImage *_pbc_image[5];
       [_menuItem setImage: nil];
     }
 
-  //[super setMenuItem: item];
   ASSIGN(_menuItem, item);
 
   if ([_menuItem image] == nil)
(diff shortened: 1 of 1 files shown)


Source/NSPopUpButtonCell.m
```diff
@@ -609,6 +609,7 @@ - (NSImage *) _currentArrowImage
 
       if (_pbcFlags.preferredEdge == NSMaxYEdge)
         {
+	  NSLog(@"\tDOWN ARROW!!!!!!");
```
fredkiefer, line 612: Looks like a leftover.

gcasa: Still debugging that's why I made it a draft.  This issue is really irking me. 🤨

PR #324, opened 2024-12-16
Don't attempt to load representations if the filename is nil.

rmottola:


rmottola:
checking for a test by @IOOI-SqAR - @gcasa

fredkiefer (approved):

PR #326, opened 2025-01-05
Add method that tells the app/backend that we are using a dark mode theme

gcasa:
The purpose of this PR is to add a method to make the system aware that a theme using "dark mode" is being used.   This will assist in settings for editors and, particularly, with printing.

### DIFF
--- a/Headers/Additions/GNUstepGUI/GSTheme.h
+++ b/Headers/Additions/GNUstepGUI/GSTheme.h
@@ -784,7 +784,7 @@ APPKIT_EXPORT_CLASS
 - (NSString*) license;
 
 /**
- * Returns YES if in dark mode, and NO, if not.  The reasoning behind this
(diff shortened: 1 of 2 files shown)


rmottola:
A discussed via voice, I think we need two properties:
- Theme "is dark". When switched system issues dark mode notification and identifies as darkmode.
- Theme supports both light and dark modes

The latter porperty is for future work, non need to implement it now.

isDarkMode in this context is a little bit ambiguous.

rmottola:
Why was this closed without explanation?

gcasa:
> Why was this closed without explanation?

Because I wasn't sure if this was the right approach.   Are you sure this is the right way to go?

Some of this functionality needs to reside in the backend as well.   The issue is that the theme is affecting how printing is processed.   It is a mystery to me how macOS handles this because some apps print exactly what is in the view and some don't.  So pinning down exactly what is done at the basic level is difficult without us writing our own word processing application which is no small feat.

PR #328, opened 2025-01-24
revert part of TestPlant imports. The _updateFieldEditor method is

rmottola:
called *after* setUpFieldEditorAttributes so it reset attributes set to NSText text object.

### DIFF
--- a/Source/NSTextFieldCell.m
+++ b/Source/NSTextFieldCell.m
@@ -286,14 +286,6 @@
     }
 }
 
-- (void) _updateFieldEditor: (NSText*)textObject
-{
-  [super _updateFieldEditor: textObject];
-  [textObject setDrawsBackground: _textfieldcell_draws_background];
-  [textObject setBackgroundColor: _background_color];
-  [textObject setTextColor: _text_color];
-}
-
 - (BOOL) isOpaque
 {
   if (_textfieldcell_draws_background == NO 


fredkiefer:
I fully agree with you that this change is wrong and shouldn't be there. But on the other side TestPlant had a reason to write something and Greg also had a reason to apply this change. So even though that change was wrong we should try to find out what the underlying issue was that they tried to resolve and maybe address that in a better way?
If we just revert it, we might end up in a commit battle.

fredkiefer (approved):

gcasa:
> I fully agree with you that this change is wrong and shouldn't be there. But on the other side TestPlant had a reason to write something and Greg also had a reason to apply this change. So even though that change was wrong we should try to find out what the underlying issue was that they tried to resolve and maybe address that in a better way?
> 
> If we just revert it, we might end up in a commit battle.

The volume of changes that came on from testplant/keysight was difficult to deal with.  As I remember I questioned this one when I reviewed it, but I decided to accept it only because it seemed minor.

I am okay with this change being reverted and if keysight responds we will deal with it at that point.  At least we will learn more and might be able to help them address the actual issue.

rmottola:
@fredkiefer @gcasa  of course they wanted to "fix" something, but without a testcase it is difficult to reproduce. If they ask to reopenit, we shall ask for examples, more testing or evidence. In our case this revert fixes the edit bug of the Cells in AddressView I checked with you one evening.

PR #330, opened 2025-02-04
protect ivars described as portected

rmottola:
We had a bug on this, but I can't find it!


rmottola:
Compiles for me here on Linux/GCC, but am rebuilding a couple of applications to see if there are side-effects.

fredkiefer:
Looks good to me. Maybe we should delay merging until somebody fixes the pipeline?

fredkiefer (approved):

rmottola:
@fredkiefer CI was fixed by Greg while discussing. I back-ported,checked, and see no harm

PR #332, opened 2025-02-07
Fix display of icons in GWorkspace if the home is a symlink

rmottola:
resolve links, so that e.g /home/xxx /usr/home/xxx match

### DIFF
--- a/Source/NSWorkspace.m
+++ b/Source/NSWorkspace.m
@@ -28,7 +28,7 @@
    If not, see <http://www.gnu.org/licenses/> or write to the 
    Free Software Foundation, 51 Franklin Street, Fifth Floor, 
    Boston, MA 02110-1301, USA.
-*/ 
+*/
 
 #import "config.h"
 
@@ -794,10 +794,10 @@ static NSDictionary		*urlPreferences = nil;
     }
 
   if (sysDir != nil)
-    [folderPathIconDict setObject: @"GSFolder" forKey: sysDir];
+    [folderPathIconDict setObject: @"GSFolder" forKey: [sysDir stringByResolvingSymlinksInPath]];
 
   [folderPathIconDict setObject: @"HomeDirectory"
-    forKey: NSHomeDirectory()];
+			 forKey: [NSHomeDirectory() stringByResolvingSymlinksInPath]];
 
   /* it would be nice to use different root icons... */
   [folderPathIconDict setObject: @"Root_PC" forKey: NSOpenStepRootDirectory()];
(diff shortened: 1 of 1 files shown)


gcasa:
I worked with @rmottola on this and tested it on my machine.  It appears to work properly.

fredkiefer:
Could you please explain what the issue was? Normally the code in GUNstep should be able to handle symlinks, where didn't that work?
Also if it was when reading the image, why didn't you resolve the issue there or at least in the same file on line 1568? Even then the worst I can see is that the same image will get stored with two names. Why is that an issue? At least within NSWorkspace this should not cause any difference. What is GWorkspace doing with this internal data?

rmottola:
@fredkiefer this is a long-standing issue it took me a while to understand.

Image cache has keys, I don't think there would be an issue with having multiple keys.

The scenario is the following. Suppose you have `/home/fred` as your home, but then this is actually a link to somewhere else (e.g. `/usr/home/fred` on FreeBSD by default, or even `/second_disk/home/fred`).

Under your home, you also have further paths, like `Desktop`.

The cache is constructed by NSWorkspace, without interaction with GWorkspace (right so, it is lower-level, shared and also used by e.g. File Panel browsers).

To know the user home `NSHomeDirectory()` is used. It essentially gets it data from /etc/passwd where it is legit to have the `/home/fred` returned.

So your cache keys looks  like this:

- /home/fred
- /home/fred/Documents

These are valid paths if you resolve them, but the "actual" location is different, e.g.:

- /second_disk/home/fred
- /second_disk/home/fred/Documents

When you actually "browse" in GWorkspace or a File Panel, you browse the actual thing. E.g. you see this:

/ -> /second_disk -> - /second_disk/home -> /second_disk/home/fred

The icon gets queried for: `/second_disk/home`  and  `/second_disk/home/fred` and not found in the cache dictionary!

After reasoning and other pitfalls (e.g. if you look at `/home/Documents` it does not appear as a link while _home_ is... so the subpaths needs to be checked).
I thought the best way is to store the cache with the actual paths, it is safest and this is what it is trying to do.

We could have double entries, but for sure I need the full path and at that point it makes sense to always refer to it.

rmottola:
@fredkiefer  to explain your objection at line 1568.

When ` iconForFile: (NSString*)fullPath` it is indeed called with the full path true to its name, so in my example above `/second_disk/home/fred` however the key in the cache is `/home/fred` and I see no way how to get the key from the full path, while my patch attempts to store all images with their full path. I can always "resolve" a path with links to the full path, but not the other way around.

rmottola:
I performed some tests.

I used as test for the use: file browser and gworkspace, accessing the relevant folders both through the proposed home (which starts with the symlink) as well as extra navigating though the long path. I don't know other scenarios where folder icons are retrieved.

- [ ] Linux, /home is just /home
- [ ] NetBSD /home is just /home
- [ ] FreeBSD /home is link to /usr/home
- [ ] Solaris /home is /home

fredkiefer (approved):
Thank you, now I understand the change.

PR #341, opened 2025-03-31
Use NSWorkspace inforForExcetsion to check if current registred HelpV…

rmottola:
…iewer is capable of showing the Help File, otherwise fall-back to NSWorkspace open as if no HelpViewer is set

### DIFF
--- a/Source/NSHelpManager.m
+++ b/Source/NSHelpManager.m
@@ -220,7 +220,7 @@
 
 		  // Let's try to be lenient if Viewer was set instead of Viewer.app
 		  if ([[viewer pathExtension] length] == 0)
-		    viewer = [viewer stringByAppendingPathExtension:@"app"];
+		    viewer = [viewer stringByAppendingPathExtension: @"app"];
 		  appInfo = [apps objectForKey: viewer];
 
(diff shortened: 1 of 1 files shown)


Source/NSHelpManager.m
```diff
@@ -215,27 +215,16 @@ - (void) showHelp: (id)sender
...
-			  NSUInteger j;
-
-			  unixExtensions = [[types objectAtIndex: i] objectForKey:@"NSUnixExtensions"];
-			  for (j = 0; j < [unixExtensions count]; j++)
-			    [extensions addObject: [[unixExtensions objectAtIndex:j] lowercaseString]];
-			}
-		    }
+		  // Let's try to be lenient if Viewer was set instead of Viewer.app
+		  if ([[viewer pathExtension] length] == 0)
+		    viewer = [viewer stringByAppendingPathExtension:@"app"];
```
fredkiefer, line 223: Here one space is missing.

fredkiefer (approved):

Source/NSHelpManager.m
rmottola: fixed!

Issue #344, opened 2025-04-16
Crash in NSView from NSMenuView/NSView dealloc


rmottola:
On Solaris Sparc 64 bit on exit from SystemPreferences I get this crash, completely inside GUI.
Needs display is invoked with a null rect, all comes from deallocing a menu with a nil.
It is not happening in other application, this is very strange.

```
#0  0xfef4043c in -[NSView _setNeedsDisplayInRect:real:] (self=0x360d30, _cmd=<optimized out>, v=0x0)
    at NSView.m:2833
#1  0xfef3ec04 in -[NSView setNeedsDisplayInRect:] (self=0x360d30, 
    _cmd=0xff19aa18 <_OBJC_SELECTOR_TABLE+1536>, invalidRect=...) at NSView.m:2913
#2  0xfef32378 in -[NSView removeFromSuperview] (self=0x2cd360, 
    _cmd=0xff1769e0 <_OBJC_SELECTOR_TABLE+1536>) at NSView.m:939
#3  0xfee74ae8 in -[NSMenuView update] (self=0x360d30, _cmd=0xff176850 <_OBJC_SELECTOR_TABLE+1136>)
    at NSMenuView.m:682
#4  0xfee72d78 in -[NSMenuView setMenu:] (self=0x360d30, _cmd=<optimized out>, menu=<optimized out>)
    at NSMenuView.m:324
#5  0xfee6cb10 in -[NSMenu dealloc] (self=0x2adb30, _cmd=0xfe970838 <_OBJC_SELECTOR_TABLE+24>)
    at NSMenu.m:450
#6  0xfe65a070 in objc_release_fast_np_internal (anObject=0x2adb30) at NSObject.m:593
#7  release_fast (anObject=0x2adb30) at NSObject.m:607
#8  -[NSObject release] (self=0x2adb30, _cmd=0xfe96ddfc <_OBJC_SELECTOR_TABLE+40>) at NSObject.m:2112
#9  0xfe647574 in -[GSNotification dealloc] (self=0x35dae0, _cmd=0xfe970838 <_OBJC_SELECTOR_TABLE+24>)
    at NSNotificationCenter.m:101
#10 0xfe65a070 in objc_release_fast_np_internal (anObject=0x35dae0) at NSObject.m:593
#11 release_fast (anObject=0x35dae0) at NSObject.m:607
[...]

PR #345, opened 2025-04-17
do not crash if passed rect as value is nil #344

rmottola:
for bug #344

### DIFF
--- a/Source/NSView.m
+++ b/Source/NSView.m
@@ -2830,8 +2830,13 @@ in the main thread.
 
 - (void) _setNeedsDisplayInRect_real: (NSValue *)v
 {
-  NSRect invalidRect = [v rectValue];
+  NSRect invalidRect;
   NSView *currentView = _super_view;
+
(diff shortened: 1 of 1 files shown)


fredkiefer (approved):

rmottola:
@rfm  are you satisfied with this fix (which is anyway safe and good, even if it maybe just hides the issue) or do you want to investigate the race condition further?

rfm (approved):
Yes, I have no problem with the fix; it's good defensive code to check for nil when using a method that returns a struct.

More fundamentally, we should probably remove the NSValue atExit code as it exists solely to avoid false positives with some memory leak checkers: I don't think asan or modern valgrind are subject to those false positives, and leak checkers generally support suppression lists to deal with false positives anyway.

The *interesting* thing would be to figure out how/why the NSView code is being called after the process exits.  I suppose that might be due to some poor design of the way an app quits, but most likely it's really not important though.

PR #355, opened 2025-07-13
"Enter" icon in default button only for certain NSInterfaceStyles

probonopd:
Show "Enter" icon in default button only if `NSInterfaceStyle` is  `NSNextStepInterfaceStyle` or `GSWindowMakerInterfaceStyle`.

That icon looks alien in other interface styles such as `NSMacintoshInterfaceStyle`:

image

Reference:
* https://github.com/gershwin-desktop/gershwin-rik-theme/issues/4

### DIFF
--- a/Source/GSHelpManagerPanel.m
+++ b/Source/GSHelpManagerPanel.m
@@ -101,8 +101,13 @@ static GSHelpManagerPanel* _GSsharedGSHelpPanel;
     [button setTitle: NSLocalizedString(@"OK", @"")];
     [button setKeyEquivalent: @"\r"];
     [button setImagePosition: NSImageRight];
-    [button setImage: [NSImage imageNamed: @"common_ret"]];
-    [button setAlternateImage: [NSImage imageNamed: @"common_retH"]];
+  NSInterfaceStyle interfaceStyle = NSInterfaceStyleForKey(@"NSInterfaceStyle", nil);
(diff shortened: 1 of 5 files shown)


Source/GSHelpManagerPanel.m
```diff
@@ -101,8 +101,13 @@ - (id)init
     [button setTitle: NSLocalizedString(@"OK", @"")];
     [button setKeyEquivalent: @"\r"];
     [button setImagePosition: NSImageRight];
-    [button setImage: [NSImage imageNamed: @"common_ret"]];
-    [button setAlternateImage: [NSImage imageNamed: @"common_retH"]];
+  NSInterfaceStyle interfaceStyle = NSInterfaceStyleForKey(@"NSInterfaceStyle", nil);
```
fredkiefer: The indentation here looks wrong

fredkiefer (approved):
Looks good to me, but maybe we should discuss this change on the mailing list before merging.

Source/GSHelpManagerPanel.m
probonopd: Thanks. Fixed.

probonopd:
According to https://mail.gnu.org/archive/html/discuss-gnustep/2025-07/msg00059.html this should be done in the theme rather than here. Hence closing.

PR #357, opened 2025-07-22
Add more documentation

gcasa:


gcasa:
Merging this since it is only documentation, more to come on other branches in the future.

PR #360, opened 2025-09-02
Add missing documentation

gcasa:
Add missing documentation to headers


gcasa:
As this causes no errors in compilation and is purely documentation related, I am merging this.

PR #361, opened 2025-09-04
Add NSFontAssetRequest implementation

gcasa:
This PR provides an implementation of NSFontAssetRequest.

The test for this is here:

https://github.com/gcasa/NSFontAssetRequest_test

### DIFF
--- a/Headers/Additions/GNUstepGUI/GSFontAssetDownloader.h
+++ b/Headers/Additions/GNUstepGUI/GSFontAssetDownloader.h
@@ -126,6 +126,16 @@ GS_EXPORT_CLASS
                                         error: (NSError **)error;
 
 /**
+ * Downloads and installs a font from the specified descriptor with a preferred format.
+ * This variant allows specifying the preferred font format (e.g., "woff2", "woff", "ttf")
+ * when downloading from CSS URLs. For direct font URLs, the format parameter is ignored.
+ * Returns YES if the font was successfully downloaded and installed, NO otherwise.
+ */
+- (BOOL) downloadAndInstallFontWithDescriptor: (NSFontDescriptor *)descriptor
+                               preferredFormat: (NSString *)format
+                                        error: (NSError **)error;
+
+/**
  * Resolves a font URL from a font descriptor.
  * This method can be overridden to implement custom URL resolution
  * strategies, such as querying different font services or using
--- a/Headers/Additions/GNUstepGUI/GSFontAssetDownloader.h
+++ b/Headers/Additions/GNUstepGUI/GSFontAssetDownloader.h
@@ -177,6 +177,17 @@ GS_EXPORT_CLASS
                              error: (NSError **)error;
 
 /**
+ * Downloads a font file from the specified URL with a given font name.
+ * This variant allows specifying the font name for better filename generation.
+ * The downloaded file will be saved with a name based on the font name and
+ * appropriate extension. Returns the path to the downloaded temporary file,
+ * or nil on failure.
+ */
+- (NSString *) downloadFontFromURL: (NSURL *)fontURL
+                          fontName: (NSString *)fontName
+                             error: (NSError **)error;
+
+/**
  * Extracts font URLs from CSS content based on the specified format.
  * This method parses CSS @font-face declarations and extracts URLs
  * that match the given format (e.g., "woff2", "woff", "ttf").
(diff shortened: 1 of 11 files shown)


gcasa:
@fredkiefer I have tested this and it is working.   Please take a look and let me know if it is acceptable.

Headers/Additions/GNUstepGUI/GSFontAssetDownloader.h
```diff
@@ -0,0 +1,287 @@
+/* Definition of class GSFontAssetDownloader
```
fredkiefer, line 1: Does this file have to be public or would it be sufficient to have it in Source?

Headers/AppKit/NSFontAssetRequest.h
```diff
@@ -100,6 +109,18 @@ APPKIT_EXPORT_CLASS
 - (instancetype) initWithFontDescriptors: (NSArray *)fontDescriptors
                                  options: (NSFontAssetRequestOptions)options;
 
+/**
```
fredkiefer: I could not find these methods in the Apple documentation. If they are GNUstep extensions we should at least put them into a separate category. But do we really want to make all this public?

Source/GSFontAssetDownloader.m
```diff
@@ -0,0 +1,951 @@
...
+}
+
+- (BOOL) downloadAndInstallFontWithDescriptor: (NSFontDescriptor *)descriptor
+			       preferredFormat: (NSString *)format
+					error: (NSError **)error
+{
+  if (descriptor == nil)
+    {
+      if (error != NULL)
+	{
```
fredkiefer, line 127: Here the indentation seems wrong.

Headers/Additions/GNUstepGUI/GSFontAssetDownloader.h
gcasa: I put it into GNUstepGUI since the idea is that it can be overridden and replaced if someone wants to use a different downloader or font source.  Right now the default is to use Google because it's kind of a standard for open fonts.

gcasa: If you believe it should internal I will make that change, but that was my original reason.

Headers/AppKit/NSFontAssetRequest.h
gcasa: It's here...

https://developer.apple.com/documentation/appkit/nsfontassetrequest/init(fontdescriptors:options:)?language=objc

Source/GSFontAssetDownloader.m
gcasa: Fixed

Source/GSFontAssetDownloader.m
```diff
@@ -0,0 +1,951 @@
...
+					      styleMask: NSWindowStyleMaskTitled | NSWindowStyleMaskClosable
+						backing: NSBackingStoreBuffered
+						  defer: NO];
+  [_progressPanel setTitle: @"Font Download"];
+  [_progressPanel setLevel: NSModalPanelWindowLevel];
+  [_progressPanel setReleasedWhenClosed: NO];
+
+  // Create and configure the status label
+  NSRect labelFrame = NSMakeRect(20, 70, 360, 20);
+  _statusLabel = [[NSTextField alloc] initWithFrame: labelFrame];
```
fredkiefer, line 873: Here all the RELEASE statements for the different elements are missing.

Headers/AppKit/NSFontAssetRequest.h
fredkiefer: That is strange, I still can't find -setDefaultDownloaderClass: there. Not sure what I am doing wrong.

gcasa: Oh.  That is an addition.  Your comment is on the "initWithFontDescriptors" method or so I thought.  I will put the setSefaultDownloader methods into a separate category.

gcasa: Done

rfm:
I'm away from home for the next week and a half, and don't think I'll be able to look at this until I'm back.

gcasa:
I will make another PR and move the relevant code to the backend.

fredkiefer:
Thank you Greg, I think it would be sufficient to just remove the font installation code here and not to revert the whole PR. Th question then is, how should we structure that code in the backend to reduce the security risks? One position would be to move the decision over to the user. If they really want to download and install a font from a dubious website it is their problem.

gcasa:
> Thank you Greg, I think it would be sufficient to just remove the font installation code here and not to revert the whole PR. Th question then is, how should we structure that code in the backend to reduce the security risks? One position would be to move the decision over to the user. If they really want to download and install a font from a dubious website it is their problem.

I am working to answer these questions right now.   I am not sure.  I think confirmation by the user is sufficient.  The downloader uses google fonts, mainly because I can't think of any other place where I can get the full font list AND then download the fonts for free. :/  I implemented it so that a developer could implement one and use that if needed so the developer isn't locked in.  So the user can't just get a font from anywhere.  Thanks fred.  I hope your vacation was restful. GC

PR #362, opened 2025-09-12
Add implementations for NSAccessibilityCustomRotor/Action and changes to related classes

gcasa:
This PR adds support for NSAccessibilityCustomRotor and NSAccessibilityCustomAction to improve accessibility navigation and interaction for assistive technologies.

In compliance with policy: Parts of this PR, both code and documentation, were generated using AI


gcasa:

rfm (commented):
I'm not familiar with what this stuff is supposed to be doing, but as far as I can see this replaces unimplemented code so can't really be breaking anything even if it doesn't operate correctly.

One thing that  confuses me is the use of
#pragma clang diagnostic ignored "-Warc-performSelector-leaks"

From what I read, that diagnostic should only be produced if you are building in ARC mode, but we don't make the core libraries depend on ARC and it doesn't look like you are even preparing to make this code depend on ARC (for instance, that would require the calls to [super dealloc]; to be replaced with the DEALLOC macro), so I wonder why the pragma to suppress a non-existent warning?  Is it because a compiler bug generates an incorrect warning?

Headers/AppKit/NSAccessibilityCustomAction.h
```diff
@@ -52,9 +52,18 @@ APPKIT_EXPORT_CLASS
                        target: (id)target
                      selector: (SEL)selector;
 
+/* Convenience factory returning an autoreleased custom action that invokes a block. */
```
fredkiefer: Shouldn´t we separate the GNUstep extensions from the standard methods? That way people might get frustrated when writing for GNUstep and then porting to MacOS. It would be easy to have a specific category here.

Source/NSAccessibilityCustomAction.m
```diff
@@ -33,8 +33,12 @@ - (instancetype) initWithName: (NSString *)name
   self = [super init];
   if (self != nil)
     {
-      ASSIGN(_name, name);
-      ASSIGN(_handler, handler);
+      ASSIGNCOPY(_name, name);
+      if (_handler != handler)
```
fredkiefer: Why copy this complicated code? You could just use the setter here.

Source/NSAccessibilityCustomRotor.m
```diff
@@ -25,89 +25,131 @@
...
+  [super dealloc];
 }
-  
+
 @end
 
 // Results...
 @implementation NSAccessibilityCustomRotorItemResult : NSObject
 
 - (instancetype)initWithTargetElement:(id<NSAccessibilityElement>)targetElement
```
fredkiefer: Spaces are missing in this line.

Source/NSAccessibilityCustomRotor.m
```diff
@@ -25,89 +25,131 @@
...
 {
-  return nil;
+  self = [super init];
+  if (self != nil)
+    {
+      _targetElement = targetElement;
+      _targetRange = NSMakeRange(0, 0);
```
fredkiefer: Did you verify this specific value? The old value `NSMakeRange(0,NSNotFound)` seems more plausible to me.

fredkiefer (approved):
I would prefer to see these few changes, but otherwise the code looks fine.

Headers/AppKit/NSAccessibilityCustomAction.h
gcasa: Done

Source/NSAccessibilityElement.m
```diff
@@ -1,30 +1,149 @@
...
+
+- (NSString *) accessibilityLabel
+{
+  return _accessibilityLabel;
+}
+
+- (void) setAccessibilityLabel: (NSString *)label
+{
+  ASSIGN(_accessibilityLabel, label);
+}
```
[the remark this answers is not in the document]
gcasa: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/gnustep/libs-gui/pull/362#discussion_r2366434166)

Source/NSAccessibilityCustomAction.m
gcasa: simplified.

Source/NSAccessibilityCustomRotor.m
gcasa: Resolved

Source/NSAccessibilityCustomRotor.m
gcasa: Resolved

Source/NSAccessibilityElement.m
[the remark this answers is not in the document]
gcasa: Corrected

Headers/AppKit/NSView.h
```diff
@@ -201,6 +203,21 @@ PACKAGE_SCOPE
   NSAppearance* _appearance;
   NSUserInterfaceItemIdentifier _identifier;
 
+  // Accessibility support
```
fredkiefer: Do we really want to blow up every NSView instance with that many extra instance variables? I think it would be better to have a pointer to a separate structure, doesn't have to be an object, but could be. And that would be only allocated when actually used. Richard uses this concept in different places.

gcasa: As you already approved this, do you want me to do that now or in a separate PR?

gcasa: These changes are made, they seem to look okay to me.  Let me know if you're okay with them.  I added this after approval as your suggestion came afterwards.

Source/NSAccessibilityCustomAction.m
```diff
@@ -100,5 +104,43 @@ - (void) setSelector: (SEL)selector
...
+- (BOOL) perform
+{
+  if (_handler != NULL)
+    {
+      CALL_BLOCK(_handler, YES);
+      return YES;
+    }
+  if (_target != nil && _selector != NULL && [_target respondsToSelector: _selector])
+    {
+      // Suppress potential leak warning for performSelector (intentional dynamic invocation)
```
fredkiefer: What does this comment mean? I see no suppression here.

Source/NSAccessibilityCustomAction.m
```diff
@@ -100,5 +104,43 @@ - (void) setSelector: (SEL)selector
...
+      return YES;
+    }
+  return NO;
+}
+
+- (NSString *) description
+{
+  return [NSString stringWithFormat: @"<%@: %p name=%@ hasHandler=%@ target=%@ selector=%@>",
+          NSStringFromClass([self class]), self, _name,
+          _handler?@"YES":@"NO", _target, NSStringFromSelector(_selector)];
```
fredkiefer: Spaces are missing

Source/NSButton.m
```diff
@@ -48,6 +51,311 @@ @interface NSButtonCell (_NSButton_Private_)
...
+  // Check for radio button first
+  if ([cell respondsToSelector: @selector(_isRadio)] && [cell _isRadio])
+    {
+      return NSAccessibilityRadioButtonRole;
+    }
+  
+  // Check button behavior to determine if it's stateful (checkbox-like)
+  // Buttons that maintain state (can be on/off) are typically checkboxes
+  if ([self allowsMixedState] || 
+      ([self state] != NSControlStateValueOff && [self respondsToSelector: @selector(setState:)]))
```
fredkiefer: Why the `respondsToSelector:` check? All NSButton instances respond to this method.

Source/NSAccessibilityCustomAction.m
gcasa: I had removed the #pragma that was there originally, but I forgot to remove the comment.

Source/NSAccessibilityCustomAction.m
gcasa: Fixed

Source/NSButton.m
gcasa: Removed.

Source/NSColorWell.m
```diff
@@ -502,3 +505,195 @@ - (id) target
...
+- (id) accessibilityValue
+{
+  NSColor *color = [self color];
+  if (color)
+    {
+      // Use component methods directly for better compatibility
+      CGFloat red, green, blue, alpha;
+      
+      // Try to get RGB components using getRed:green:blue:alpha:
+      @try
```
fredkiefer: I am not sure here, but as far as I remember we still prefer to have all try/catch code written with the old macros. We could change that though, but I would prefer an official decision.

gcasa: We need to stick to NS_DURING and the ObjC1.0 conventions.   I will take this out.

gcasa: My apologies, force of habit.  I am so accustomed to using try/catch in my work that I sometimes do it by habit.

gcasa: Done

Source/NSColorWell.m
```diff
@@ -502,3 +505,195 @@ - (id) target
...
+        {
+          [color getRed: &red green: &green blue: &blue alpha: &alpha];
+          return [NSString stringWithFormat: @"RGB(%.0f, %.0f, %.0f)", 
+                  red * 255.0, green * 255.0, blue * 255.0];
+        }
+      @catch (NSException *exception)
+        {
+          // Fallback to individual component methods if available
```
fredkiefer: This is rather strange code. For the cases where the method call above fails, these components are also not defined. What should be done is to convert the colour to a well defined colour space and get the components there.

Source/NSView.m
```diff
@@ -769,13 +773,26 @@ - (void) dealloc
   TEST_RELEASE(_cursor_rects);
   TEST_RELEASE(_tracking_rects);
   TEST_RELEASE(_shadow);
+  RELEASE(_accessibilityData);
```
fredkiefer: Could be a TEST_RELEASE

Source/NSColorWell.m
gcasa: Okay, I will do that.

gcasa: I simplified the solution.   It essentially retrieves the localized name if a named color space is used.   If they don't it falls back to the description.  This is how the function is described on macOS.

Source/NSView.m
gcasa: Done.

Source/NSView.m
```diff
@@ -5547,6 +5564,296 @@ - (void) _recursiveSetUpKeyViewLoopWithNextKeyView: (NSView *)nextKeyView
...
+  return YES;
+}
+
+- (void) setAccessibilityElement: (BOOL) isElement
+{
+  // This could be stored in instance variable if needed
+  // For now, calculated dynamically based on children
+}
+
+- (NSRect) accessibilityFrame
```
fredkiefer, line 5786: Why is this in screen coordinates whereas accessibilityActivationPoint is in parent view coordinates?
Also why has the implementation of this method been duplicated in many subclasses?

gcasa: Because I felt as though, since this is for accessibility, it would be more useful for someone who is blind to hear absolute vs relative coordinates.

fredkiefer: But that still leaves my question unanswered. Some of these methods are in one coordinate system, some in the other. Sometimes the same method has another coordinate system when implemented in a different class. At least for me this is very confusing. We should define in advance for each method what to use and implement it consistently.
And you did not address the second question.

gcasa:
The test for the NSColor stuff is here https://github.com/gcasa/NSColor_test

PR #364, opened 2025-09-21
Add skeletal implementation of NSGestureRecognizer

gcasa:
Implement Gesture recognition logic

### DIFF
--- a/Headers/AppKit/NSEvent.h
+++ b/Headers/AppKit/NSEvent.h
@@ -452,6 +452,8 @@ APPKIT_EXPORT_CLASS
           CGFloat deltaY;
           /** Z-axis scroll delta for 3D scroll devices */
           CGFloat deltaZ;
+          /** Magnification factor for gesture events */
+          CGFloat magnification;
         } mouse;
       /** Keyboard event data including characters and key codes */
       struct
@@ -588,6 +590,7 @@ APPKIT_EXPORT_CLASS
  * deltaX: Horizontal scroll delta for scroll events
  * deltaY: Vertical scroll delta for scroll events
  * deltaZ: Z-axis scroll delta for 3D scroll devices
+ * magnificationValue: Magnification factor for gesture events
  * Returns: A new NSEvent object with extended mouse information
  */
 + (NSEvent*) mouseEventWithType: (NSEventType)type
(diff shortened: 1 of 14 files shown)


gcasa:

Source/NSEvent.m
```diff
@@ -853,21 +871,21 @@ - (id) initWithCoder: (NSCoder*)aDecoder
         case NSLeftMouseDragged:
         case NSOtherMouseDragged:
         case NSRightMouseDragged:
-          [aDecoder decodeValuesOfObjCTypes: "iififff",
+          [aDecoder decodeValuesOfObjCTypes: "iififfff",
```
fredkiefer, line 874: You should not change the encoding/decoding without changing the version number and checking on the version when decoding.

Headers/AppKit/NSEvent.h
```diff
@@ -602,7 +605,8 @@ APPKIT_EXPORT_CLASS
                    buttonNumber: (NSInteger)buttonNum
                          deltaX: (CGFloat)deltaX
                          deltaY: (CGFloat)deltaY
-                         deltaZ: (CGFloat)deltaZ;
+                         deltaZ: (CGFloat)deltaZ
+                  magnification: (CGFloat)magnificationValue;
```
fredkiefer, line 609: If I remember correctly we actually use this method in the MS Windows backend. Changing it here without a change there will break things. The best way to process would be to add a new method here with the additional argument, add that argument to all the backend code that uses the old method and after the next GNUstep release, remove the old method without that argument.

Source/NSGestureRecognizer.m
```diff
@@ -9,40 +9,458 @@
...
    License along with this library; see the file COPYING.LIB.
-   If not, see <http://www.gnu.org/licenses/> or write to the 
-   Free Software Foundation, 51 Franklin Street, Fifth Floor, 
+   If not, see <http://www.gnu.org/licenses/> or write to the
+   Free Software Foundation, 51 Franklin Street, Fifth Floor,
    Boston, MA 02110-1301, USA.
 */
 
 #import <AppKit/NSGestureRecognizer.h>
+#import <AppKit/NSEvent.h>
```
fredkiefer, line 31: Normally we would include the AppKit headers here with "" instead of <>. This allows local headers to be found before installed ones and could make a difference when headers change.

Source/NSGestureRecognizer.m
```diff
@@ -9,40 +9,458 @@
...
+{
+  return _state;
+}
+
+- (void)_setState:(NSGestureRecognizerState)state
+{
+  if (_state == state)
+    return;
+
+  NSGestureRecognizerState oldState = _state;
```
fredkiefer, line 120: Why store the old state when nothing gets done with it?

PR #365, opened 2025-09-28
Fix the GUI interface version (restore code which was, presumably accidentally, removed a while back)

rfm:
Re-instate the gui interface version setting to specify the version for linking the library in order to maintain ABI compatibility. Without this, apps linked to one version of the GUI library will try to use more recent (but incompatible) versions.


fredkiefer (approved):
Feel free to go ahead

PR #366, opened 2025-09-29
Re-enable font asset request on this branch for testing / refactor

gcasa:
Also move font installation to backend, remove hard-coded font paths.


gcasa:
@fredkiefer I have moved the installer to the backend, so it can be overridden.

PR #371, opened 2025-10-11
Update to include function to open an alert panel relative to a window

gcasa:
Add new function to bring up a panel relative to a window

### DIFF
--- a/Source/NSAlert.m
+++ b/Source/NSAlert.m
@@ -1809,7 +1809,7 @@ NSRunAlertPanelRelativeToWindow(NSString *title,
     }
   else
     {
-      panel = getSomeSheet(&standardAlertPanel,
+      panel = getSomeSheet(&informationalAlertPanel,
 			   defaultTitle,
 			   title,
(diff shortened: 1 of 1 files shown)


Source/NSAlert.m
```diff
@@ -1770,6 +1770,60 @@ void NSBeginInformationalAlertSheet(NSString *title,
...
+                           defaultButton, alternateButton, otherButton);
+      result = [panel runModal];
+      NSReleaseAlertPanel(panel);
+      return result;
+    }
+
+  if (defaultButton == nil)
+    defaultButton = @"OK";
+
+  panel = getSomeSheet(&standardAlertPanel, defaultTitle, title, message,
```
fredkiefer: For `getSomeSheet` we should use a separate instance, otherwise normal panels could also be displayed as sheets, if this method gets called first. But this is already wrong in a few other places in this file.

gcasa: Done, using a different instance for the sheet.

PR #375, opened 2025-12-14
Fixes for openURL/Open URL service

rfm:
This fixes/improves various issues with the openURL mechanism.
We allow such services to accept data of either NSURLPboardType or NSStringPboardType, and provide both on the pasteboard.
We allow NSProvideService() to use any service matching the types of the supplied pasteboard.
We make the service name used be more permissive (either 'Open URL' or 'openURL').
Also, added some diagnostics to make_services and fixed it to check whether an app that dynamically registered services still exists.  This fixes the problem where old dynamic service entries for an app which used to exist override new static entries for apps which do exist.

### DIFF
--- a/Source/GSServicesManager.m
+++ b/Source/GSServicesManager.m
@@ -90,26 +90,6 @@ static NSString		*providerName = nil;
 
 static NSDictionary *serviceFromAnyLocalizedTitle(NSString *title);
 
-/* Check every NSPort entry in the NSServices array in our info plist.
- */
-static BOOL
-isServicePort(NSString *name)
-{
-  NSDictionary	*info = [[NSBundle mainBundle] infoDictionary];
-  NSEnumerator	*svcs = [[info objectForKey: @"NSServices"] objectEnumerator];
-
-  while ((info = [svcs nextObject]) != nil)
-    {
-      NSString	*portName = [info objectForKey: @"NSPortName"];
-
-      if ([name isEqual: portName])
-	{
-	  return YES;
-	}
-    }
-  return NO;
-}
-
 /**
  * Unregisters the service provider registered on the named port.<br />
  * Applications should use [NSApplication-setServicesProvider:] with a nil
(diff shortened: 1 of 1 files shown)


rfm:
That's a very fair question, and the truth is that it isn't really necessary. But ...

The -runUntilDate: method is perversely unintuitive in that it doesn't always run until the date specified. It actually runs until that date as long as there is something waiting for an event in the run loop.  So the timer is added to the run loop to ensure that it runs for at least a tenth of a second.

Currently, we always have a housekeeping timer in the main run loop, and also have a pipe in the main loop of each thread (used to send messages between threads), so in this situation we don't really need to add the timer.  But the presence of those things are implementation details of the base library and could in theory change.  So adding the timer ourselves avoids a dependency on something that's not dependable.
The other reason it's not really necessary is that the worst that could happen is we get a busy loop, using 100% of a CPU core as we repeatedly enter the run loop and leave again.  Unlikely to be a big deal.

rfm:
I simplified the make_services code, and added some more improvements for opening URLs:
Avoid generating alert panels if trying to use the 'Open URL' service fails, because we have two more options to try.
Try to launch a non-gnustep browser using the NSWebBrowser user default (new code)
Try to open the URL as a file of unknown type (already present).
Also, change the -openURL: method to deactivate the app if any of those methods succeeds (previously this was done only if the first attempt to open the URL: succeeds).

rfm:
Sorry about all the additions after my initial fixes ... I think that openURL (and the service part in particular) is working pretty smoothly now, but it would be good if people could try it out and see if further tweaks are warrented before merging.

ChangeLog
```diff
@@ -1,3 +1,25 @@
...
+	be 1 second rather than 30.
+	* Source/NSPasteboard.m: Make reading/writing string data from/to a
+	pasteboard work for URLs.
+	* Source/NSWorkspace.m: When looking for an Open URL service, use the
+	correct name and use both string and URL types so that services can
+	work whether they are expecting strings or more specific URL data.
+	If we can't open a URL using a service, try to use the NSWebBrowser
+	user default to launch a non-gnustep browser before resorting to treat
+	a URL as an unknown file type.  Don't raise alert panels on service
+	failure since we have a couple more options to try oafter that.
```
fredkiefer: oafter => after

PR #383, opened 2026-01-20
restructure layout line new paragraph

rmottola:
Attempt to restructure code a little

(in progress)

### DIFF
--- a/Source/GSHorizontalTypesetter.m
+++ b/Source/GSHorizontalTypesetter.m
@@ -711,7 +711,7 @@ restart: ;
       line_frags_num = 0;
       rect = [curTextContainer lineFragmentRectForProposedRect: remain
                                                 sweepDirection: NSLineSweepRight
-                                             movementDirection: line_frags_num?NSLineDoesntMove:NSLineMovesDown
+                                             movementDirection: NSLineMovesDown
                                                  remainingRect: &remain];
       while (!NSIsEmptyRect(rect))
         {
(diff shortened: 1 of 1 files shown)


Source/GSHorizontalTypesetter.m
```diff
@@ -694,50 +694,56 @@ -(int) layoutLineNewParagraph: (BOOL)newParagraph
...
-	  /* Try to make sure each container contains at least one line frag
-	  rect by shrinking our line height. */
-	  line_height = [curTextContainer containerSize].height;
-	  max_line_height = line_height;
-	  goto restart;
-	}
-      return 1;
-    }
+                                                sweepDirection: NSLineSweepRight
+                                             movementDirection: line_frags_num?NSLineDoesntMove:NSLineMovesDown
```
fredkiefer: Here `line_frags_num` is always 0 so the passed on value is NSLineMovesDown.

Source/GSHorizontalTypesetter.m
```diff
@@ -694,50 +694,56 @@ -(int) layoutLineNewParagraph: (BOOL)newParagraph
...
+          if (line_frags_num > line_frags_size)
+            {
+              line_frags_size += 2;
+              line_frags = realloc(line_frags, sizeof(line_frag_t) * line_frags_size);
+            }
+          line_frags[line_frags_num - 1].rect = rect;
 
+          rect = [curTextContainer lineFragmentRectForProposedRect: remain
+                                                    sweepDirection: NSLineSweepRight
+                                                 movementDirection: line_frags_num?NSLineDoesntMove:NSLineMovesDown
```
fredkiefer: Here `line_frags_num` is always non-zero, so we pass the value NSLineDoesntMove.
If we really want to make two calls, then we should at least split them up correctly.

Source/GSHorizontalTypesetter.m
rmottola: I understand. Now that the code has unrolled the calls, we can resolve the conditions. I hope the last commit does what you intend.

PR #386, opened 2026-01-27
Fix viewbased cell refresh

gcasa:
Refresh the tableview when the window is resized to fix the issue with viewbased cell refresh.


gcasa:
I am going to go ahead and merge this to main since it is critical (and relatively simple).

PR #389, opened 2026-01-28
refactor  WANT_LINE_HEIGHT(h)  macro with static inline function

rmottola:
Try to improve code-flow readability and see if late it can be further improved.

Functionally should be a no-change.

PS. git merge rebase or to maintain  comments instead of squash?


rmottola:
@fredkiefer if you have a better idea for maxLineHeight, tell me. I was thinking in somehow setting it locally to the function since it is static, but rarely do that so don' t know how the most efficient way to do ti. I tried not to add statements in this expensive code.

Issue #354, opened 2025-07-12
NSOutlineView does no support variable row height


gcasa:
I need to create outline view with variable row height, but NSOutlineView does not call delegate's method outlineView:heightOfRowByItem:

Is there another way to do it?

PR #390, opened 2026-02-03
Use variable height callbacks for NSOutlineView

gcasa:
This is to address bug #354

### DIFF
--- a/Tests/gui/NSOutlineView/README.md
+++ b/Tests/gui/NSOutlineView/README.md
@@ -1,0 +1,55 @@
+# NSOutlineView Variable Row Heights Test
+
+This test validates the new delegate methods for NSOutlineView:
+
+## Delegate Methods Tested
+
+1. **`outlineView:heightOfRowByItem:`**
+   - Returns custom height for each item
+   - Enables variable row heights in outline views
+   - Root items: 30pt height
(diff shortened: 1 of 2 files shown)


Source/NSTableView.m
```diff
@@ -6986,6 +7029,121 @@ - (NSInteger) _numRows
...
+
+	for (i = 0; i < rowIndex && i < _numberOfRows; i++)
+		{
+			y += [self _rowHeightForRow: i];
+		}
+
+	return y;
+}
+
+- (NSInteger) _rowAtPointUsingVariableHeights: (NSPoint)aPoint
```
fredkiefer, line 7088: Why have this code as a separate method and not in the -rowAPoint: method?

Source/NSOutlineView.m
```diff
@@ -2080,6 +2081,55 @@ - (NSInteger) _numRows
...
+          return height;
+        }
+    }
+
+  return [self rowHeight];
+}
+
+- (void) _sendDoubleActionForColumn: (NSInteger)columnIndex
+{
+  if ([_delegate respondsToSelector:
```
fredkiefer, line 2113: This implementation is a bit of a surprise. In which case should resizing be the standard action behind the double action?

PR #391, opened 2026-02-04
Clarify Variable names, cleanup code

rmottola:
Working on the code with @IOOI-SqAR  Lars we found it quite confusing to read, so before fixing, I propose some further refactor.

- fix variable aliasing
- clarify if variables are index counters or actual object pointers (some comparisons were unreadable)
- remove goto which was actually very minimal

Current code seems equivalent both in intent and in test cases I used (Ink test files, Help Viewer.

Single commits for each operation of refactor. I'd rebase them as they are. Or add further for suggestions.


IOOI-SqAR (commented):
LGTM!

PR #392, opened 2026-02-05
Improve line-break and attribute handling.

### LINKED ISSUE #382: layoutmanager - new line issue
HelpViewer - when using its internal parser which is portable to Cocoa - does not display the same way on Cocoa as on GNUstep. First issue found is a new-line issue.

debugging HelpViewer and Note boxes. Notes on debugging session of 12th Jan

In the code, there is an extra hack to append a "new line" before appending a Text Attachment. It is needed only on GNUstep, where now an extra-new-line is shown, on Cocoa it is not (looks better on GNUstep). However it is a hack. Removing it makes things the same on Mac, GNUstep follows, but then wrongly formats the line. One can clearly see that the previous line is affected by the following Box.
A first guess is that for some reason the layout manager does not take in account that the notes box is put on a new line (because it doesn't fit) and reformats the current line on its height wrongly

rmottola:
Improve line-break and attribute handling. Fix that if the last Glyph or Attachment do not fit on the current line, the attribute sof the current line should remain unchanged.
- sets a local variable didLastGlyphFit to handle Wrapping
- checks extra that the Attachment is not the first element in line, or overlappings happen However, word-wrapping still retains the original bug and is here not handled

### DIFF
--- a/Headers/Additions/GNUstepGUI/GSHorizontalTypesetter.h
+++ b/Headers/Additions/GNUstepGUI/GSHorizontalTypesetter.h
@@ -64,7 +64,7 @@ APPKIT_EXPORT_CLASS
   NSFont *currentFont;
   NSRange fontRange; /* glyphs */
 
-  struct GSHorizontalTypesetterGlyphCacheEntry *glyphCache;
+  struct GSHorizontalTypesetterGlyphCacheStruct *glyphCache;
   /*
     cacheBase: index of first glyph in cache within the text container
     cacheSize: capacity of cache
@@ -74,7 +74,7 @@ APPKIT_EXPORT_CLASS
   BOOL atEnd;
 
 
-  struct GSHorizontalTypesetterLineFragment *lineFragments;
+  struct GSHorizontalTypesetterLineFragmentStruct *lineFragments;
   int lineFragmentCount, lineFragmentCapacity;
 }
 
--- a/Headers/Additions/GNUstepGUI/GSHorizontalTypesetter.h
+++ b/Headers/Additions/GNUstepGUI/GSHorizontalTypesetter.h
@@ -15,13 +15,13 @@
 
    This library is distributed in the hope that it will be useful,
    but WITHOUT ANY WARRANTY; without even the implied warranty of
-   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.	 See the GNU
+   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
    Lesser General Public License for more details.
 
    You should have received a copy of the GNU Lesser General Public
    License along with this library; see the file COPYING.LIB.
-   If not, see <http://www.gnu.org/licenses/> or write to the 
-   Free Software Foundation, 51 Franklin Street, Fifth Floor, 
+   If not, see <http://www.gnu.org/licenses/> or write to the
+   Free Software Foundation, 51 Franklin Street, Fifth Floor,
    Boston, MA 02110-1301, USA.
 */
 
@@ -40,42 +40,42 @@ APPKIT_EXPORT_CLASS
 {
   NSLock *lock;
 
-  GSLayoutManager *curLayoutManager;
-  NSTextContainer *curTextContainer;
-  NSTextStorage *curTextStorage;
+  GSLayoutManager *currentLayoutManager;
+  NSTextContainer *currentTextContainer;
+  NSTextStorage *currentTextStorage;
 
-  unsigned int curGlyphIndex;
-  NSPoint curPoint;
+  unsigned int currentGlyphIndex;
+  NSPoint currentPoint;
 
 
-  NSParagraphStyle *curParagraphStyle;
+  NSParagraphStyle *currentParagraphStyle;
   NSRange paragraphRange; /* characters */
 
-  NSDictionary *curAttributes;
+  NSDictionary *currentAttributes;
   NSRange attributeRange; /* characters */
   struct
     {
-      BOOL explicit_kern;
+      BOOL explicitKern;
       float kern;
-      float baseline_offset;
+      float baselineOffset;
       int superscript;
     } attributes;
 
-  NSFont *curFont;
+  NSFont *currentFont;
   NSRange fontRange; /* glyphs */
 
-  struct GSHorizontalTypesetter_glyph_cache_s *cache;
+  struct GSHorizontalTypesetterGlyphCacheEntry *glyphCache;
   /*
-    cache_base: index of first glyph in cache within the text container
-    cache_size: capacity of cache
-    cache_length: how much of the cache is filled
+    cacheBase: index of first glyph in cache within the text container
+    cacheSize: capacity of cache
+    cacheLength: how much of the cache is filled
    */
-  unsigned int cache_base, cache_size, cache_length;
-  BOOL at_end;
+  unsigned int cacheBase, cacheSize, cacheLength;
+  BOOL atEnd;
 
 
-  struct GSHorizontalTypesetter_line_frag_s *line_frags;
-  int line_frags_num, line_frags_size;
+  struct GSHorizontalTypesetterLineFragment *lineFragments;
+  int lineFragmentCount, lineFragmentCapacity;
 }
 
 +(GSHorizontalTypesetter *) sharedInstance;
@@ -83,4 +83,3 @@ APPKIT_EXPORT_CLASS
 @end
 
 #endif
-
--- a/Source/GSHorizontalTypesetter.m
+++ b/Source/GSHorizontalTypesetter.m
@@ -170,23 +170,23 @@ typedef struct GSHorizontalTypesetterGlyphCacheStruct GlyphCacheEntry;
     attributes.superscript = 0;
 }
 
--(void) _moveCacheToGlyph: (unsigned int)glyph
+-(void) _moveCacheToGlyph: (unsigned int)glyphIndex
 {
   BOOL valid;
 
-  if (cacheBase <= glyph && cacheBase + cacheLength > glyph)
+  if (cacheBase <= glyphIndex && cacheBase + cacheLength > glyphIndex)
     {
-      int delta = glyph - cacheBase;
+      int delta = glyphIndex - cacheBase;
       cacheLength -= delta;
       memmove(glyphCache, &glyphCache[delta], sizeof(GlyphCacheEntry) * cacheLength);
-      cacheBase = glyph;
+      cacheBase = glyphIndex;
       return;
     }
 
-  cacheBase = glyph;
+  cacheBase = glyphIndex;
   cacheLength = 0;
 
-  [currentLayoutManager glyphAtIndex: glyph
+  [currentLayoutManager glyphAtIndex: glyphIndex
                         isValidIndex: &valid];
 
   if (valid)
@@ -194,7 +194,7 @@ typedef struct GSHorizontalTypesetterGlyphCacheStruct GlyphCacheEntry;
       unsigned int charIndex;
 
       atEnd = NO;
-      charIndex = [currentLayoutManager characterIndexForGlyphAtIndex: glyph];
+      charIndex = [currentLayoutManager characterIndexForGlyphAtIndex: glyphIndex];
       [self _cacheAttributesAtCharacterIndex: charIndex];
 
       paragraphRange = NSMakeRange(charIndex, [currentTextStorage length] - charIndex);
@@ -207,7 +207,7 @@ typedef struct GSHorizontalTypesetterGlyphCacheStruct GlyphCacheEntry;
           currentParagraphStyle = [NSParagraphStyle defaultParagraphStyle];
         }
 
-      currentFont = [currentLayoutManager effectiveFontForGlyphAtIndex: glyph
+      currentFont = [currentLayoutManager effectiveFontForGlyphAtIndex: glyphIndex
                                                                  range: &fontRange];
     }
   else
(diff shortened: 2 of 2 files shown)


Source/GSHorizontalTypesetter.m
```diff
@@ -834,15 +836,26 @@ -(int) layoutLineNewParagraph: (BOOL)newParagraph
...
-
-	    if (f_ascender > ascender)
-	      ascender = f_ascender;
-	    if (f_descender > descender)
-	      descender = f_descender;
-
-            if (wantNewLineHeight(new_height, &line_height, max_line_height))
-              goto restart;
+            /* does the glyph fit ?*/
+            if (p.x + g->size.width > lf->rect.size.width)
```
fredkiefer: We could simplify all this block by setting first `didLastGlyphFit = p.x + g->size.width <= lf->rect.size.width` and checking that value in the `if`. That way we only have one `if` branch per case.

Source/GSHorizontalTypesetter.m
```diff
@@ -1010,23 +1035,35 @@ the first tab would move (exactly) to the next tab
 	    compared to everything else here, and has it's origin in p.
 	    (Makes sense from the cell's pov, though.) */
 
-	    if (-NSMinY(r) > descender)
-	      descender = -NSMinY(r);
+            /* does the attachment fit (and it is not the first element in line) ?*/
+            if ((i > firstGlyphIndex) && (p.x + NSMaxX(r) > lf->rect.size.width))
```
fredkiefer: I am not sure, whether we need this first condition in all the other cases as well. It is hard to construct a use case where this actually would happen, the first character not fitting into a line, but if it is possible, we would now handle it wrong.

Source/GSHorizontalTypesetter.m
```diff
@@ -1052,7 +1089,7 @@ the first tab would move (exactly) to the next tab
 	  }
 
 	/* Did the glyph fit in the line frag rect? */
-	if (p.x > lf->rect.size.width)
+	if (!didLastGlyphFit || p.x > lf->rect.size.width)
```
fredkiefer: Do we still need the second expression, or does `!didLastGlyphFit` include this condition in all cases?

Source/GSHorizontalTypesetter.m
rmottola: You suppose having an extremely big character? I did try that and it did work. However I noticed there is some asymmetry in the code of Glyphs vs Attachments. Glyphs call wantNewLineHeight twice, Attachments once.

Source/GSHorizontalTypesetter.m
rmottola: The ultimate goal would be to make didLastGlyph fit the only. Compared to my first attempts, I was able to reduce the scope of didLastGlyphFit of one loop level and set it to YES and NO everytime for safety.
I can try and see, but I would like to reason well with you if it is enough.

Source/GSHorizontalTypesetter.m
rmottola: Do we have regressions test to run besides me testing apps and making testfiles with Ink?

Source/GSHorizontalTypesetter.m
```diff
@@ -834,15 +836,26 @@ -(int) layoutLineNewParagraph: (BOOL)newParagraph
...
-              goto restart;
+            /* does the glyph fit ?*/
+            if (p.x + g->size.width > lf->rect.size.width)
+              {
+                // did not fit
+                didLastGlyphFit = NO;
+              }
+            else
+              {
+                // glyh did fit
```
IOOI-SqAR: Nitpick: typo in comment

Source/GSHorizontalTypesetter.m
IOOI-SqAR: Regression tests would indeed be nice to have. Although I have no idea on how to test this.

Source/GSHorizontalTypesetter.m
rmottola: Implemented.

Source/GSHorizontalTypesetter.m
rmottola: I extended the first-char condition in the Glyph test above, id does no harm.

Source/GSHorizontalTypesetter.m
rmottola: We shouldn't need that, I removed it after implementing other changes and everything seems to work as before.

Headers/Additions/GNUstepGUI/GSHorizontalTypesetter.h
```diff
@@ -40,47 +40,46 @@ APPKIT_EXPORT_CLASS
...
+    cacheLength: how much of the cache is filled
    */
-  unsigned int cache_base, cache_size, cache_length;
-  BOOL at_end;
+  unsigned int cacheBase, cacheSize, cacheLength;
+  BOOL atEnd;
 
 
-  struct GSHorizontalTypesetter_line_frag_s *line_frags;
-  int line_frags_num, line_frags_size;
```
rmottola, line 78: personally, I prefer Size to Capacity - but @fredkiefer  has the word!

Source/GSHorizontalTypesetter.m
```diff
@@ -99,333 +99,343 @@ +(GSHorizontalTypesetter *) sharedInstance
...
   /* These fields are filled in during layout: */
   BOOL nominal;
-  NSPoint pos;    /* relative to the line's baseline */
+  NSPoint position;    /* relative to the line's baseline */
   NSSize size;    /* height is used only for attachments */
-  BOOL dont_show, outside_line_frag;
+  BOOL dontShow, outsideLineFragment;
 };
-typedef struct GSHorizontalTypesetter_glyph_cache_s glyph_cache_t;
+typedef struct GSHorizontalTypesetterGlyphCacheEntry GlyphCacheEntry;
```
rmottola: Personally I liked the distinction between structure and type. I would have done:

typedef struct GSHorizontalTypesetterGlyphCacheStruct GlyphCacheEntry;

again, @fredkiefer

Source/GSHorizontalTypesetter.m
```diff
@@ -99,333 +99,343 @@ +(GSHorizontalTypesetter *) sharedInstance
...
-  curAttributes = nil;
+  currentAttributes = nil;
   attributeRange = NSMakeRange(0, 0);
-  curFont = nil;
+  currentFont = nil;
   fontRange = NSMakeRange(0, 0);
 }
 
--(void) _cacheAttributes: (unsigned int)char_index
+-(void) _cacheAttributesAtCharacterIndex: (unsigned int)characterIndex
```
rmottola, line 139: why not just _cacheAttributesAtIndex ? @fredkiefer

Source/GSHorizontalTypesetter.m
```diff
@@ -99,333 +99,343 @@ +(GSHorizontalTypesetter *) sharedInstance
...
-      i = [curLayoutManager characterIndexForGlyphAtIndex: glyph];
-      [self _cacheAttributes: i];
-
-      paragraphRange = NSMakeRange(i, [curTextStorage length] - i);
-      curParagraphStyle = [curTextStorage attribute: NSParagraphStyleAttributeName
-					atIndex: i
-					longestEffectiveRange: &paragraphRange
-					inRange: paragraphRange];
-      if (curParagraphStyle == nil)
+      unsigned int characterIdx;
```
rmottola: personally I think we are going too much in verbose... is this AI ?
i is fine as iteraror and tool variable. charIndex... I don't know, taste @fredkiefer

Source/GSHorizontalTypesetter.m
```diff
@@ -99,333 +99,343 @@ +(GSHorizontalTypesetter *) sharedInstance
...
-      at_end = YES;
+      atEnd = YES;
     }
 }
 
--(void) _cacheGlyphs: (unsigned int)new_length
+-(void) _cacheGlyphsUpToLength: (unsigned int)newLength
 {
-  glyph_cache_t *g;
+  GlyphCacheEntry *glyphEntry;
```
rmottola, line 221: also for me here glyphEntry is verbose and adds not much.. compared to g or glyph... @fredkiefer

Source/GSHorizontalTypesetter.m
```diff
@@ -99,333 +99,343 @@ +(GSHorizontalTypesetter *) sharedInstance
...
+   including glyphIndex will have been cached.
 */
--(unsigned int) breakLineByWordWrappingBefore: (unsigned int)gi
+-(unsigned int) breakLineByWordWrappingBefore: (unsigned int)glyphIndex
 {
-  glyph_cache_t *g;
-  unichar ch;
-  NSString *str = [curTextStorage string];
+  GlyphCacheEntry *glyphEntry;
+  unichar character;
```
rmottola, line 283: please, again, ch is perfectly understandable as character, isn't it @fredkiefer your opinion? I would suggest uChar

Source/GSHorizontalTypesetter.m
```diff
@@ -99,333 +99,343 @@ +(GSHorizontalTypesetter *) sharedInstance
...
+      start = lineFragment->lastGlyphIndex;
+      lineFragment->lastUsed = lineFragment->rect.size.width;
     }
 }
 
--(void) rightAlignLine: (line_frag_t *)lf : (int)num_line_frags
+-(void) rightAlignLine: (LineFragment *)lineFragment : (int)numLineFragments
 {
-  unsigned int i;
+  unsigned int index;
```
rmottola, line 402: again here and in other cases, if i is just an iterator, it is just fine... no need to expand it? @fredkiefer

rmottola:
unfortunately the big refactor of @IOOI-SqAR is not equivalent and introduces regressions - first char of new paragraph is not drawn or is put out of place.

Source/GSHorizontalTypesetter.m
```diff
@@ -99,330 +99,343 @@ +(GSHorizontalTypesetter *) sharedInstance
...
-    attributes.superscript = [n intValue];
+  numberValue = [currentAttributes objectForKey: NSSuperscriptAttributeName];
+  if (numberValue)
+    attributes.superscript = [numberValue intValue];
   else
     attributes.superscript = 0;
 }
 
--(void) _cacheMoveTo: (unsigned int)glyph
+-(void) _moveCacheToGlyph: (unsigned int)glyph
```
fredkiefer: glyph should be called glyphIndex as that is what it is.

IOOI-SqAR: done in 5fdf07ac8dc6936f19b0b50a33a5cde1d0e10f0a

Source/GSHorizontalTypesetter.m
```diff
@@ -584,71 +598,580 @@ static inline BOOL wantNewLineHeight(CGFloat h, CGFloat *lineHeight, CGFloat max
...
+          glyphEntry->position.y = position.y - cellFrame.origin.y;
+
+          position.x = glyphEntry->position.x + glyphEntry->size.width;
+
+          /* An attachment is always in a point range of its own. */
+          glyphEntry->nominal = NO;
+        }
+      else
+        {
+          /* TODO: this is a major bottleneck */
```
fredkiefer, line 901: In this case we don't compute `doesGlyphFitInLine`, which seems wrong to me. Also the old comment about the actual correct, but slow code is lost.

rmottola: But the old code was commented out, only two lines were proper and we cleaned it up.

True, we don't calculate doesGlyphFitInLine here (but never did either). But what condition is it? the standard glyph (not attachment, not control character) isn't handled at line 716?

My understanding is

Line 715 : For all glyphs, calculate doesGlyphFitInLine
Line 757 : For ControlGlyphs do special stuff, then continue loop
Line 851 : For Attachments, do some more stuff
Line 903 : for "non-attachments" adjust glyph position, but what can it be? non-control and non-attachment, a standard glyph treated in 715. Why recalculate doesGlyphFitInLine ?

As a test, I put in a doesGlyphFitInLine calculation and from an NSLog noticed that it usually fits but not always. But does this override the previous calculation of line 716? don't know. Maybe position.x is recalculated. Recalculating it doesn't seem to do neither good nor harm.

fredkiefer: You are correct and I should have known all that, but it has been too long since we last worked on this code.
Still the value of `positions.x` will be wrong if the glyph won't fit into the line. This value gets used to set the `lastused` value of the line fragment. I am not sure whether this is important. Most likely we should only change that, when actually adding the glyph to the line fragment.
As for the commented out code, this is what should actually happen here. But as Alexander wrote it is too slow to do, we would need an optimization here. That code was here to remind us of this short coming.

rmottola: @fredkiefer oh, got you. Didn't understand you meant the commented out-code as bottleneck, now I understand also better the comment. My fault. I put back the code back in, updated it to current variable names, tested it, found no real difference, commented it out again and committed it so it remains for future reference.

rmottola: as your observation about `position.x` it sounds correct. Would this reflect your idea?

 ```
         /* does the attachment fit (and it is not the first element in line) ?*/
          doesGlyphFitInLine = !((index > firstGlyphIndex) && (position.x + NSMaxX(cellFrame) > lineFragment->rect.size.width));
          if (doesGlyphFitInLine)
            {
              lastPosition = glyphEntry->position = position;
              /* Only the width is used. */
              position.x += glyphEntry->size.width;
            }
```

Or could we trust the previously calculated doesGlyphFitInLine?

This without the bottleneck code, for that one I wonder if it should be all skipped since it reference only the lastGlyph that "should" have fitted.

fredkiefer (approved):

PR #397, opened 2026-02-17
Add a markdown for AI policy

gcasa:
Hey @fredkiefer, per our discussion, I have written up a markdown that I think encapsulates our discussion on Saturday.  Please review it and let me know if you can think of any revisions.

In compliance with this policy: AI was used in the creation of this document.

### DIFF
--- a/POLICY_AI.md
+++ b/POLICY_AI.md
@@ -26,28 +26,78 @@ AI tools may be used as assistive drafting aids. Contributors remain
 fully responsible for the accuracy, safety, quality, and licensing
 compliance of anything they submit.
 
-## 4. Mandatory Thorough Review
+## 4. Mandatory Independent Verification and Enhanced Review
 
-Any AI-assisted contribution **must be reviewed thoroughly** before it
-is merged or otherwise accepted.
+AI-assisted contributions are subject to heightened scrutiny and must
+undergo rigorous, independent human validation prior to acceptance.
 
-Thorough review includes, at minimum:
+### 4.1 Independent Human Verification
 
--   **Correctness** -- The content behaves as intended and aligns with
-    GNUstep conventions and architectural requirements.
--   **Security & Safety** -- No unsafe patterns, hidden network calls,
-    credential leakage, injection risks, or insecure defaults.
--   **Maintainability** -- Code style, naming, structure, and comments
-    meet project standards and are understandable to maintainers.
--   **Testing** -- Appropriate tests are added or updated; existing
-    tests pass. Compilation alone is not sufficient validation.
--   **Documentation Accuracy** -- Claims and instructions are verified
-    against current GNUstep behavior and supported platforms.
--   **Licensing Awareness** -- The contributor ensures that no
-    copyrighted or license-incompatible material is introduced.
+The contributor must personally:
 
-Reviewers may apply additional scrutiny where changes are large,
-complex, security-sensitive, or difficult to reason about.
+-   Read and understand every line of AI-assisted code or documentation.
+-   Be able to explain the logic, control flow, and design decisions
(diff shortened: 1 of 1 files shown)


gcasa:
And, yes, this policy was partly generated by AI as I asked for suggestions about things it should cover.

gcasa:
@rfm the idea of a policy was discussed at the meeting on Saturday with @fredkiefer.     I came up with this after.  Take a look and review.   If we agree on it, please put a copy in the base library as well.

POLICY_AI.md
```diff
@@ -0,0 +1,120 @@
...
+
+## 3. Guiding Principle
+
+AI tools may be used as assistive drafting aids. Contributors remain
+fully responsible for the accuracy, safety, quality, and licensing
+compliance of anything they submit.
+
+## 4. Mandatory Thorough Review
+
+Any AI-assisted contribution **must be reviewed thoroughly** before it
```
fredkiefer: This is the one point where I would like to see a more strict policy. 
"The submission must the thoroughly reviewed by the submitter before opening a PR."
It is already our standard policy to have a code review before merging, so the current formulation would change nothing.

gcasa: What do you think it should look like.  I am going to take a crack at making it better.

gcasa: I made it stricter... please review.

fredkiefer: To me the new version doesn't look more strict. It rather seems to me that when you were talking about the reviewer you formulated a lot of requirements, now that you are talking about the submitter there are less requirements. This seems wrong to me. In the case of AI generated code we rather should have two independent reviews, one by the submitter and a second one by the reviewer.

gcasa: The issue is that I am having trouble coming up with rules that don't apply IN GENERAL.   I can only think "DON'T COMMIT MESSY AI CODE".

fredkiefer: Actually that one sentence would be good enough for me.

rmottola: > Actually that one sentence would be good enough for me.

For me too.

gcasa: LOL, okay.  I will boil it down to that.

gcasa: I boiled it down.  I left some of the other language in there for clarity, but it has been simplified.

POLICY_AI.md
```diff
@@ -0,0 +1,115 @@
...
+## 3. Guiding Principle
+
+AI tools may be used as assistive drafting aids. Contributors remain
+fully responsible for the accuracy, safety, quality, and licensing
+compliance with anything they submit.
+
+## 4. Mandatory Independent Verification and Enhanced Review
+
+While we welcome AI-assisted contributions, it is important to remember
+that AI can and does make mistakes based on its training.  Because of this
```
fredkiefer, line 33: Actually the problem is not just the insufficient training material. It is the general working of LLM, they predict probability not correctness. It is surprising how good some of the generated code still is. This mostly proves how repetitive coding really is :-(

fredkiefer (approved):
Fine by me now. Most likely we should have this file not only here but in the other core modules as well. For that reason it would be good to have a response by @rfm .

gcasa:
@fredkiefer I would love to have a discussion about AI with you offline.   Much of what we are seeing are "emergent properties".   The truth is that even experts don't understand how or why they do what they do.   It's dangerous that they are, in essense, GROWN, not coded.   I specialized in AI in college and I never imagined anything like what we have today.    They are more than just predictive because they include a neural network to manage the associations between concepts.  Anyway, we can save that for a later time...

gcasa:
Sent as an email to the list...

Issue #336, opened 2025-02-22
Application Menus and NSPopUpButtons don't stay open anymore after releasing mouse button


buzzdeee:
after upgrading to gnustep-gui 0.32.0 I observed that:

When clicking/opening menus, they close immediately when I release the mouse button.
This is true for Application Menus, as well as NSPopUpButtons, which wasn't the case before.
Downgrading to prior release, when I click a menu item in the Application menu, 
or an NSPopUpButton then it stays visible, even I release and only goes away, when I click something in the menu.

In case it matters, I use NSMenuInterfaceStyle NSMacintoshInterfaceStyle.
The Noteworthy changes for gnustep-gui-0.32.0 mention a number of NSMenu related changes...
Is that maybe even expected?

pkgdemon:
I can confirm reverting this resolves the issue with NSMacintoshInterfaceStyle https://github.com/gnustep/libs-gui/commit/82717eefec63e7d9251d780e41862858ca06c748 and it only seems to affect that style.

pkgdemon:
@fredkiefer @gcasa Any thoughts here?  Does the project see this as a regression, or more of a someone needs to make a patch to make NSMacintoshInterfaceStyle not follow this routine?  I believe this may affect NSWindows95InterfaceStyle as well for what it is worth or recall someone mentioning that it does I will confirm.  So for a user on a laptop for example with a trackpad, this gets very hard to use the menus because they rely on a button being depressed to display the menu, when the button is released the menu can no longer display.

PR #399, opened 2026-02-20
remove shouldFinish to allow outer loop to handle cleanup

johnathan-becker:
Remove shouldFinish assignment on mouse-up in inner event loop                                                                                                            

In the inner do/while loop that drains NSAppKitDefined events after a mouse-up is received, there was a redundant shouldFinish = YES assignment immediately before a  break. We didn't need this because shouldFinish is already set to yes in the outer loop whenever a mouse-up event is seen and setting it in the inner loop actually caused the outer loop exit prematurely before it does its cleanup.

### DIFF
--- a/Source/NSMenuView.m
+++ b/Source/NSMenuView.m
@@ -1834,7 +1834,6 @@ static float menuBarHeight = 0.0;
 	  type = [event type];
     if (type == NSLeftMouseUp || type == NSRightMouseUp || type == NSOtherMouseUp)
       {
-          shouldFinish = YES;
           break;  // Exit the loop to proceed to StopPeriodicEvents
       }
 	  if (type == NSAppKitDefined)


rmottola:
> I have no issue with removing this specific line, but wouldn't it be better, if we removed the whole `if` statement? I would prefer to clean up this whole thing once and for all and not have some left over, that doesn't hurt as much as the current code, but is still not a clean solution.

After our discussion @fredkiefer , I performed some tests.

Status Quo: current code has horizontal menu broken (mac-style): click on the menu does not leave the menu open, an annoying bug.

Removing `shouldFinish = YES` is enough to have mac-style menus working again.
Of course removing the whole

```
    if (type == NSLeftMouseUp || type == NSRightMouseUp || type == NSOtherMouseUp)
      {
          shouldFinish = YES;
          break;  // Exit the loop to proceed to StopPeriodicEvents
      }
```

Fixes horizontal menus for me too, so the whole if should be removed.

The whole if-statement is not placebo, because the outer loop continues while it is a MouseEvent or it should finish, while shouldFinish is set YES if a MouseEvent is found - this is cause of the menu-click breakage.

rmottola:
I proposed an equivalent fix which is more comprehensive in #336

PR #401, opened 2026-03-05
Fix for view shifting over in NSCollectionView

gcasa:
This fix should address the issue where the NSCollectionView is slightly shifted to the right.   It addresses this by recalculating the origin and resetting it properly.   I am not sure if there is an earlier miscalculation.  For now this fixes the issue.

Addresses https://github.com/gnustep/libs-gui/issues/402


fredkiefer (approved):
The change looks strange, but if it helps fixing the issue, go ahead.

PR #403, opened 2026-03-07
Fix issue with refresh of view based NSTableView

gcasa:
The issue that I am trying to fix is that when the scrollbar is used the number of rows in the NSTableView shows ONE row and not all.  I am not sure why this is happening.

### DIFF
--- a/Source/NSTableView.m
+++ b/Source/NSTableView.m
@@ -2146,8 +2146,8 @@ static void computeNewSelection
   /* Remove any existing resize observers for this table view before it
      moves to a new window or is detached. */
   [nc removeObserver: self
-		name: NSWindowDidResizeNotification
-	      object: nil];
+		 name: NSWindowDidResizeNotification
+	       object: nil];
 
   [super viewDidMoveToWindow];
 
@@ -3868,7 +3868,7 @@ if (currentRow >= 0 && currentRow < _numberOfRows) \
 		      [NSEvent stopPeriodicEvents];
 		      startedPeriodicEvents = NO;
 		    }
-		  mouseLocationView.x = _bounds.origin.x;
+		  mouseLocationView.x = 0.0;
 		  oldRow = currentRow;
 		  currentRow = [self rowAtPoint: mouseLocationView];
 
@@ -3942,7 +3942,7 @@ if (currentRow >= 0 && currentRow < _numberOfRows) \
 		      startedPeriodicEvents = NO;
 		    }
 
-		  mouseLocationView.x = _bounds.origin.x;
+		  mouseLocationView.x = 0.0;
 		  oldRow = currentRow;
 		  currentRow = [self rowAtPoint: mouseLocationView];
 		  if (oldRow != currentRow)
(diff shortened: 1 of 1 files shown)


Source/NSTableView.m
```diff
@@ -2415,6 +2420,10 @@ - (void) setDataSource: (id)anObject
       _dataSource_editable = YES;
     }
 
+  // Set refresh actions...
```
fredkiefer: This looks like a left over.

Source/NSTableView.m
```diff
@@ -5454,25 +5463,34 @@ - (void) setDelegate: (id)anObject
...
+	  // Changed from cell-based to view-based, add observer
+	  [nc addObserver: self
+		 selector: @selector(_windowDidResize:)
+		     name: NSWindowDidResizeNotification
+		   object: [self window]];
+
+	  [nc addObserver: self
+		 selector: @selector(_windowDidResize:)
+		     name: NSViewBoundsDidChangeNotification
+		   object: nil];
```
fredkiefer: If we watch this for all views, won't this result in way to many notifications?

fredkiefer:
I really would prefer if you could do formatting changes as a separate commit. This can be in the same PR. That way it would be simpler to look through the remaining changes.

gcasa:
> I really would prefer if you could do formatting changes as a separate commit. This can be in the same PR. That way it would be simpler to look through the remaining changes.

Okay.  Understood.  I will do so going forward.

gcasa:
I am going to see if there are other solutions to this problem before marking it ready.

Source/NSTableView.m
gcasa: Yes, that's the issue I have with this change.  It works, but it's messy.  It tells me that the cause is something else.  I am looking into what the root cause is.   I will clearly state the issue in the above description.

gcasa: Fix, it's now restricted to the clipView containing the tableView.

Source/NSTableView.m
gcasa: removed

gcasa:
@fredkiefer This is ready for review, I don't think I can simplify it more.

fredkiefer (commented):
As I wrote in my mail. I am not happy with this PR in its current state. But if you really want to merge it, go ahead

gcasa:
> As I wrote in my mail. I am not happy with this PR in its current state. But if you really want to merge it, go ahead

I haven't been able to find one, but I think you're right.  I am going to give it another week to try to find a better solution.

fredkiefer (approved):
I still think there should be a better way to do this. But if you want, go ahead and merge this.

gcasa:
> I still think there should be a better way to do this. But if you want, go ahead and merge this.

I agree.  I don't like how this is done.  I'm convinced it is a view resizing issue. I suspect one of the subviews is not resizing properly and hiding the rows.  I'm going to revisit this likely tomorrow.

gcasa:
PR#437 seems to fix this issue more cleanly.  https://github.com/gnustep/libs-gui/pull/437

PR #422, opened 2026-05-04
Update codec for ubuntu 22.04

gcasa:
This PR is basically a quick fix for an issue found on Ubuntu when building with avcodec and libao.


gcasa:
I am merging this as a hot fix.

fredkiefer:
I don't understand what was so urgent about this change that it could not even wait for an hour. You will know why you did this and I am not asking.
What I still would like to know why you added a config step into the make file. Of course we could do all the configuration again during compilation but we prefer not to. So why break this here?

gcasa:
A blocking issue like this seemed urgent enough.  I will streamline the process and fix the configuration.  Please let me know when you're available to review.

gcasa:
> I don't understand what was so urgent about this change that it could not even wait for an hour. You will know why you did this and I am not asking. What I still would like to know why you added a config step into the make file. Of course we could do all the configuration again during compilation but we prefer not to. So why break this here?



> I don't understand what was so urgent about this change that it could not even wait for an hour. You will know why you did this and I am not asking. What I still would like to know why you added a config step into the make file. Of course we could do all the configuration again during compilation but we prefer not to. So why break this here?

I believe what you're referring to with respect to configuration in the makefile is this...

```C
ifeq ($(BUILD_MOVIE), yes)
libgnustep-gui_OBJC_FILES += GSMovieView.m
libgnustep-gui_OBJC_FILES += GSAudioPlayer.m
libgnustep-gui_OBJC_FILES += GSAVUtils.m
endif
```

This has been in since I did the initial change for NSMovieView.

Issue #435, opened 2026-06-03
AddressManager crashes - infinite loop


rmottola:
Issue discovered by @IOOI-SqAR and it has been confirmed that it affects only libobjc2/clang builds, not gcc builds.
AddressManager never receives `applicationDidFinishLaunching` never gets invoked while `awakeFromNib` completes correctly.
No relevant changes in AddressManager itself.

rmottola:
I have certified that this is a **regression**. HEAD fails.

Works: 07a667b9f5147b70d7e350aa38517de6f31581cc
Broken: f535d7ab6b75bdfd00866ed2ddb62bd4a28d3fcb

Works: b05760475cb05a52e8f8ee481b24f85b0b694ecb

rmottola:
Bisecting yielded the issue to this merge: https://github.com/gnustep/libs-gui/commit/f535d7ab6b75bdfd00866ed2ddb62bd4a28d3fcb caused by https://github.com/gnustep/libs-gui/pull/386 of @gcasa 

I don't know how to bisect or revert merged commits, I get issues.

gcasa:
> Bisecting yielded the issue to this merge: [f535d7a](https://github.com/gnustep/libs-gui/commit/f535d7ab6b75bdfd00866ed2ddb62bd4a28d3fcb) caused by [#386](https://github.com/gnustep/libs-gui/pull/386) of [@gcasa](https://github.com/gcasa)
> 
> I don't know how to bisect or revert merged commits, I get issues.

Write a bug against this in GUI.  Assign it to me and I will address it.  The fix proposed by @IOOI-SqAR is not the proper fix.

IOOI-SqAR:
> > Bisecting yielded the issue to this merge: [f535d7a](https://github.com/gnustep/libs-gui/commit/f535d7ab6b75bdfd00866ed2ddb62bd4a28d3fcb) caused by [#386](https://github.com/gnustep/libs-gui/pull/386) of [@gcasa](https://github.com/gcasa)
> > I don't know how to bisect or revert merged commits, I get issues.
> 
> Write a bug against this in GUI. Assign it to me and I will address it. The fix proposed by [@IOOI-SqAR](https://github.com/IOOI-SqAR) is not the proper fix.

Hey @gcasa,

what fix are you talking about? I never proposed anything here …

Kind regards,

Lars

gcasa:
You never wrote a bug or detailed the actual issue.  I have no proposal until I can recreate the issue

[...]

PR #436, opened 2026-06-04
conditional removal of NSWindowDidResizeNotification to only _viewBased, attempt to fix crash

rmottola:
crude fix for https://github.com/gnustep/libs-gui/issues/435  open for discussion

I suppose this is only a workaround for #435 because AddressManager is not using view-based Tables, but what if? Why is removing a notification which perhaps is was not set such a big issue?

### DIFF
--- a/Source/NSTableView.m
+++ b/Source/NSTableView.m
@@ -2119,9 +2119,12 @@ static void computeNewSelection
 	  object: self];
     }
   // Remove window resize observer
-  [nc removeObserver: self
-                name: NSWindowDidResizeNotification
-              object: nil];
+  if (_viewBased)
+    {
+      [nc removeObserver: self
+		    name: NSWindowDidResizeNotification
+		  object: nil];
+    }
   TEST_RELEASE (_autosaveName);
   if (_numberOfColumns > 0)
     {


IOOI-SqAR (commented):
IDK if I like this, since it seems to cover up an issue with autorelease pools in base

gcasa:
This change feels wrong to me.   It seems like there are a few issues with view based cells currently.    It seems here you've uncovered another.   If there is an problem resizing, then this work around seems improper.

rmottola:
I agree it is a "cover up" change. I think it shouldn't be an issue removing that notification, it shouldn't crash and not that bad as it is doing now.
This is "food for thought".

I think we have an issue in this table code that triggers a bug in the ARP pool in case of NG libobjc2 runtime, there I ask help with @rfm see this bug as companion of https://github.com/gnustep/libs-base/pull/616

rmottola:
> This change feels wrong to me. It seems like there are a few issues with view based cells currently. It seems here you've uncovered another. If there is an problem resizing, then this work around seems improper.

The issue happens hard on startup. Use libobcj2 and will just crash.

I agree it is a cover up of potential other issues, but I also argue it is not "wrong": that NSWindowDidResizeNotification is in the rest of the code only added with view-based tables, why remove it unconditionally? Seems asymmetric to me.

A good test would be with a view-based table example. AddressManager has still a classic cell based table.

rmottola:
Besides covering up, I still thing it is good to remove notification one had set, like a reverse-order clean-up.
If certain things are set only for view-based tables, their removal should be done only for them, don't you think? Also this would help not tainting classic tables, e.g. @gcasa . It should be "harmless" to remove the notification.

So I still thing the change is "good", but I'd wait to fix the deeper issue before.

gcasa:
I have provided a more comprehensive fix for this in https://github.com/gnustep/libs-gui/pull/437

fredkiefer:
@gcasa I disagree your PR #437 tries to fix unrelated issues in your code. As @rfm explained it really is preferable that we only remove notifications we ask for and this is what this PR does.

fredkiefer (approved):

gcasa:
> @gcasa I disagree your PR #437 tries to fix unrelated issues in your code. As @rfm explained it really is preferable that we only remove notifications we ask for and this is what this PR does.

Ok. Understood.

PR #437, opened 2026-06-07
New fix for viewBased cell reloading issue

### LINKED ISSUE #435: AddressManager crashes - infinite loop

gcasa:
The purpose of these changes is to address the issue described with AddressManager (infinite loop) as well as an issue I faced where the table was not refreshing properly.   I am in the process of testing this, but I thought I would create a PR for review.

As it turns out, the issue here and the infinite loop are unrelated, but this does seem to fix the issue with redrawing of view-based cells.  I had assumed that the issue was in the code that I had written for this.  Particularly in layout and initialization.  I have cleaned up that code here.

### DIFF
--- a/Source/NSTableView.m
+++ b/Source/NSTableView.m
@@ -173,6 +173,7 @@ typedef struct _tableViewFlags
 - (CGFloat*) _columnOrigins;
 - (NSView*)  _renderedViewForPath: (NSIndexPath*)path;
 - (void) _setRenderedView: (NSView*)view forPath: (NSIndexPath*)path;
+- (void) _layoutViewBasedRows;
 @end
 
 @interface NSTableView (SelectionHelper)
@@ -2164,7 +2165,9 @@ static void computeNewSelection
 {
   if (_viewBased)
     {
-      [self reloadData];
+      [self tile];
+      [self _layoutViewBasedRows];
+      [self setNeedsDisplay: YES];
     }
 }
 
@@ -4642,6 +4645,10 @@ This method is deprecated, use -columnIndexesInRect:. */
       // TODO width?
     }
   [super setFrame: tmpRect];
+  if (_viewBased)
+    {
+      [self _layoutViewBasedRows];
+    }
 }
 
 - (void) setFrameSize: (NSSize)frameSize
@@ -4664,6 +4671,10 @@ This method is deprecated, use -columnIndexesInRect:. */
       // TODO width?
     }
   [super setFrameSize: tmpSize];
+  if (_viewBased)
+    {
+      [self _layoutViewBasedRows];
+    }
 }
 
 - (void) viewWillMoveToSuperview:(NSView *)newSuper
@@ -5022,6 +5033,11 @@ This method is deprecated, use -columnIndexesInRect:. */
       [_headerView setNeedsDisplay: YES];
       [_cornerView setNeedsDisplay: YES];
     }
+
+  if (_viewBased)
+    {
+      [self _layoutViewBasedRows];
+    }
 }
 
 /*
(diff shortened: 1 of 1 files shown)


gcasa:
This fix appears to fix the issue with resizing... I will test with AddressManager to see if it fixes the infinite loop issue....

Original window:

Untitled

After resize:

Untitled 2

rfm:
While the GUI code may have issues,  they aren't the bug causing the autorelease pool loop, and the GUI code is doing nothing wrong it its use of autorelease here as far as I can see.

gcasa:
> While the GUI code may have issues, they aren't the bug causing the autorelease pool loop, and the GUI code is doing nothing wrong it its use of autorelease here as far as I can see.

What do you believe the issue is then?   The fix that was submitted can't be correct.

gcasa:
I think this change fixes the issue with refresh more cleanly than #403

gcasa:
@fredkiefer Please review this and let me know what you think.

rfm:
>> What do you believe the issue is then? The fix that was submitted can't be correct.
If you mean the pull request created for the base library, no that was not correct at all.
I did describe the actual problem there, but the whole trail is complicated so I left out some context.

A GUI object (iirc an NSTableView) was, int its -dealloc method, removing itself as an observer for a notification which it was not registered.  That's inefficient but should be harmless.

However, the particular code path followed for that case in NSNotificationCenter returned from a method without releasing the pool allocated earlier in the method.  This was also inefficient and bad practice, but should be harmless (we have to handle that situation when exceptions occur anyway).

The above occurred during deallocation of an autorelease pool, when libobj2 was popping a pool from within the -emptyPool method.

The -emptyPool implementation failed to call  the method to clean up child autorelease pools, so the child leaked  in the  notification observer removal remained attached to the pool being emptied.  This is the first part of the bug, since the child pool should also have been emptied at this point (to release its contents).

That 'empty' pool was then pushed into the cache for re-use later.

Then we needed to create a new pool, and the alloc method returned the cached pool, but in the -init method the code to link that cached pool into the list of pools assumed that the child pointer was nil and didn't need to be updated.

Then we needed to create another new pool, and in this case the -init method took the child (from the previous pool) and linked it as its own child without adjusting the parent pointer of the child.  This created the  corrupt/inconsistent loop in one direction of the doubly linked list;  the second part of the bug.

Later still, the most recently created pool was drained, and when -_emptyChild was called it found a linked list with a cycle in it, so it looped forever.  That's not a bug in itself, just a symptom of the earlier bugs, demonstrating that the linked list had been corrupted.

Any one of many small changes could alter this chain of events and hide the issue;
avoiding the removal of an unregistered notification observer would do it
avoiding the leak of a pool in NSNotificationCenter (which I've done now) would do it
avoiding the linked list traversal would do it

rfm:
PS. about removing an object as a notification observer ...
Best practice is to do that in -dealloc, but only if the object was first registered (otherwise it's a waste of cpu)
Second best, is to remove it even if it was never registered
But actually, with a recent copy of the base library, you don't need to bother at all as it should no longer crash when it encounters a deallocated observer (because base now uses zeroing weak pointers and checks for nulls in that code).

gcasa:
@rfm I have added you as a reviewer here due to your comments.  Thank you for your input.

gcasa:
@fredkiefer Any feedback on this?  I think using tile and layout when the window is resized is better instead of a full reload.  This focuses on the actual problem instead of a shotgun approach (which is what the other approach was doing).  LMK.

fredkiefer (approved):

PR #438, opened 2026-06-08
Table Cleanup:

rmottola:
* remove observers before dealloc'ing objects
* remove notifications with the conditions that actually add them

### DIFF
--- a/Source/NSTableView.m
+++ b/Source/NSTableView.m
@@ -2099,6 +2099,25 @@ static void computeNewSelection
 {
   [self abortEditing];
 
+  if (_autosaveTableColumns == YES)
+    {
+      [nc removeObserver: self
+                    name: NSTableViewColumnDidResizeNotification
+                  object: self];
+    }
+  // Remove window resize observer
+  if (_viewBased)
+    {
+      [nc removeObserver: self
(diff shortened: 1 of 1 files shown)


Source/NSTableView.m
```diff
@@ -2145,11 +2148,12 @@ - (BOOL) isFlipped
...
-  /* Remove any existing resize observers for this table view before it
-     moves to a new window or is detached. */
-  [nc removeObserver: self
-                 name: NSWindowDidResizeNotification
-               object: nil];
+  if ([self window] != nil && _viewBased)
+    {
+      [nc removeObserver: self
+                    name: NSWindowDidResizeNotification
+                  object: [self window]];
```
fredkiefer, line 2155: This is wrong. When we did move to a window, we no longer reference the window for which we did register as an observer. You could add this code to the `viewWillMoveToWindow` or just leave the old code, maybe with a check for `_viewBased`.

rmottola: @fredkiefer took me a second to understand, but you are right - I had to subclass viewWillMoveToWindow.

fredkiefer (approved):

PR #439, opened 2026-06-17
Improve tearoff support

rmottola:


rmottola:
This branch is WIP to have the corresponding branch in back working

Source/GSHorizontalTypesetter.m
```diff
@@ -613,30 +613,30 @@ - (BOOL)_baseLayoutBlockNewParagraph:(BOOL *)newParagraph
                         andDescender:(CGFloat *)descender
           returningLineFragmentIndex:(int *)lineFragmentIndex
                  returningGlyphIndex:(unsigned int*)glyphIndex
-                   returningPosition:(NSPoint *)position
 {
-  *glyphIndex = 0;
+
```
fredkiefer: This file probably does not belong into this PR :-)

fredkiefer:
I think this is a very nice idea, but it contradicts the Apple documentation where they state that this two values are synonymous. If we still want to make this change we really need to be careful:
- are these constants used consistently in gui
- is the window level stored in any NIB files
- check all the different backends not just x11 in back for usages

And if Apple uses as documented the same value, who is willing to decide that we break compatibility?

Source/GSHorizontalTypesetter.m
rmottola: Right, but I branched off a branch and don't know how to fix it. I suppose I need to create a new branch. ? Or can I remove somehow "commits"..

rmottola:
> I think this is a very nice idea, but it contradicts the Apple documentation where they state that this two values are synonymous. If we still want to make this change we really need to be careful:
> 
>     * are these constants used consistently in gui
> 
>     * is the window level stored in any NIB files
> 
>     * check all the different backends not just x11 in back for usages
> 
> 
> And if Apple uses as documented the same value, who is willing to decide that we break compatibility?

Well, Apple does not support X11 since the time of OpenStep on solaris... MacOS does not have tear-off menus since Rhapsody. Thus we have more needs than apple!

We need this information and in the code "tear off" is actually used in distinction of menu, but then it ends being the same thing currently.

I looked around the code, cannot be sure of course this is all what is needed, but it is needed for the corresponding changes in libs-back

PR #440, opened 2026-06-27
NSBitmapImageRep: fix heap buffer overflow decoding sub-8-bit palette TIFFs

DTW-Thalion:
## Summary

A `PHOTOMETRIC_PALETTE` TIFF with `bitsPerSample < 8` causes a heap buffer overflow when decoded through `+[NSBitmapImageRep imageRepWithData:]` / `-initWithData:`.

`NSTiffRead` expands a palette image to a 24-bit RGB image, writing **three bytes per pixel**:

```c
case PHOTOMETRIC_PALETTE:
  for (row = 0; row < info->height; ++row) {
    ...
    for (col = 0; col < info->width; col++) {
      *outP++ = map->red[*inP]   / 256;
      *outP++ = map->green[*inP] / 256;
      *outP++ = map->blue[*inP]  / 256;
      inP++;
    }
  }
```

But the destination bitmap in `-_initFromTIFFImage:number:` is allocated using the file's `bitsPerSample` (only `samplesPerPixel` was forced to 3):

```c
case PHOTOMETRIC_PALETTE:
  space = NSDeviceRGBColorSpace;
  info->samplesPerPixel = 3;   /* bitsPerSample left as-is */
  break;
```

`initWithBitmapDataPlanes:` then sizes the buffer as `ceil(width * bitsPerSample * 3 / 8) * height`. For `bitsPerSample == 8` this equals the `3 * width * height` bytes that `NSTiffRead` writes, which is why ordinary palette TIFFs work. For `bitsPerSample == 4` the buffer is only ~half the required size (an ~2x over-write); for `bitsPerSample == 1` it is ~1/8 (an ~8x over-write). The image data is attacker-controlled, so this is reachable by opening a crafted file.

## Fix

* `-_initFromTIFFImage:number:` rejects palette images whose `bitsPerSample` is not 8. The palette reader only ever handled 8-bit indices (it reads one byte per pixel as the index), so non-8-bit palette images were never decoded correctly in any case.
* `-_initBitmapFromTIFF:` now honours the return value of `-_initFromTIFFImage:number:`. It previously ignored it and returned `self` even on the existing error paths, which (after `RELEASE(self)`) handed the caller a deallocated instance instead of `nil`.

## Test

`Tests/gui/NSBitmapImageRep/tiffPaletteOverflow.m` embeds two small libtiff-generated fixtures: a malicious 1024x1024 4-bit palette TIFF (LZW-compressed, so it is tiny on disk but drives a multi-megabyte over-write) and a valid 8x8 8-bit palette TIFF.

* **Negative control (unpatched):** the test aborts with a crash while decoding the 4-bit fixture.
* **Positive control (patched):** 2 of 2 assertions pass — the malicious image is rejected (`nil`, no overflow) and the valid 8-bit palette TIFF still decodes (8x8, three samples per pixel).


fredkiefer (approved):
Thank you for this patch!

PR #441, opened 2026-06-27
NSBitmapImageRep: fix heap buffer overflow importing malformed ICNS files

DTW-Thalion:
## Summary

Opening a crafted `.icns` file through `+[NSBitmapImageRep imageRepWithData:]` / `-initWithData:` causes a heap buffer overflow.

The in-tree ICNS reader (used when libicns is not available — the common configuration) reads the icns family resource size straight from the file and uses it for `malloc()` and a `memcpy()` with no validation, in `icns_import_family_data`:

```c
el_size = NSSwapBigIntToHost(element->elementSize);   /* from the file */
family = malloc(el_size);
...
memcpy((char*)(family->elements),
       (char*)element + ICNS_HEADER_SIZE,
       el_size - ICNS_HEADER_SIZE);
```

`ICNS_HEADER_SIZE` is 8. If the file's resource size field is **less than 8**, `el_size - ICNS_HEADER_SIZE` underflows (unsigned) to a huge length, so a file beginning with `"icns"` followed by a resource size of, say, `4` writes far past a tiny `malloc(4)` and reads far past the input buffer. A resource size **larger than the file** likewise makes the `memcpy` read past the end of the input.

The resource size is the four bytes immediately after the `"icns"` magic, so this is trivially reachable by opening a malicious file.

## Fix

`icns_import_family_data` now rejects an `el_size` that is smaller than the header or larger than the data actually available from the element, before the `malloc`/`memcpy`.

## Test

`Tests/gui/NSBitmapImageRep/icnsImportOverflow.m` uses two hand-built fixtures: an `"icns"` file with a resource size of 4 (the overflow trigger), and a valid 32x32 `il32` icon as a positive control.

* **Negative control (unpatched):** the test aborts with heap corruption while importing the malformed fixture.
* **Positive control (patched):** 2 of 2 assertions pass — the malformed file is rejected (`nil`, no overflow) and the valid 32x32 icon still decodes.


fredkiefer (approved):
Thank you again

PR #442, opened 2026-06-27
NSBitmapImageRep: fix out-of-bounds read decoding truncated ICNS mask/RLE data

DTW-Thalion:
## Summary

Decoding a crafted `.icns` file through `+[NSBitmapImageRep imageRepWithData:]` / `-initWithData:` reads past the end of the heap buffer holding the icon family.

In the in-tree ICNS reader (used when libicns is not available — the common configuration), `icns_get_image32_with_mask_from_family` decodes the icon and then copies the 8-bit mask:

```c
// mask copy: reads imageHeight*imageWidth bytes with no check against `end`
for (j = 0; j < iconImage->imageHeight * iconImage->imageWidth; j++)
  iconImage->imageData[samplesPerPixel * j + 3] = *b++;
```

The loop copies `imageHeight * imageWidth` bytes (e.g. 1024 for a 32×32 `l8mk`, 16384 for a 128×128 `t8mk`) regardless of how many bytes the mask element actually contains. A crafted file whose mask element is shorter than the icon dimensions makes `*b++` read past the element data and past the end of the malloc'd icon family buffer. The RLE icon decode has the same problem in its run loops (the compressed `val` read and the uncompressed run copy advance `b` past `end`).

The over-read leaks adjacent heap memory into the decoded image's alpha (or colour) channel, and can crash if it crosses an unmapped page. It is reachable simply by opening a malicious file.

## Fix

Bound each of the three reads (`icns_get_image32_with_mask_from_family`) by the end of the element data:
* the RLE compressed-run value read,
* the RLE uncompressed-run copy, and
* the 8-bit mask copy.

## Test

`Tests/gui/NSBitmapImageRep/icnsMaskBounds.m` decodes a 32×32 `il32` icon with a solid colour followed by a truncated (empty) `l8mk` mask, and checks that it decodes without crashing and still returns its colour data.

The over-read is silent on a normal heap, so the negative control is an AddressSanitizer replica of the mask loop:

```
==ERROR: AddressSanitizer: heap-buffer-overflow ... READ of size 1
  ... in main   (unpatched logic)
done alpha0=190           (patched: (b < end) bound, clean)
```

Validated on Ubuntu 24.04 with clang + libobjc2. The `gnustep-tests` regression test passes 2/2 with the fix; the ASan replica reports the out-of-bounds read on the unpatched logic and is clean once the bound is added.


fredkiefer (approved):
Looks good

PR #443, opened 2026-06-27
NSBitmapImageRep: avoid infinite loop on a zero-size ICNS element (DoS)

DTW-Thalion:
## Summary

Opening a crafted `.icns` file through `-[NSBitmapImageRep initWithData:]` or `+[NSBitmapImageRep imageRepsWithData:]` can hang the process in an infinite loop.

In the in-tree ICNS reader (used when libicns is not available — the common configuration), both `_initBitmapFromICNS:` and `_imageRepsWithICNSData:` walk the icon family like this:

```c
while (((dataOffset + 8) < iconFamily->resourceSize))
  {
    icns_element_t element;
    memcpy(&element, (data + dataOffset), 8);
    ...
    dataOffset += element.elementSize;   /* never advances when 0 */
  }
```

If an element's size field is zero, `dataOffset` never advances and the loop spins forever. The element size comes straight from the file, so an `"icns"` family containing a single element with a zero size field (plus a little padding so the loop is entered) hangs the reader.

## Fix

Both walks now stop when they encounter an element with a size of zero.

## Test

`Tests/gui/NSBitmapImageRep/icnsZeroElement.m` feeds such a file to both `-initWithData:` and `+imageRepsWithData:` and checks that each returns.

* **Negative control (unpatched):** the test hangs and has to be killed by a timeout.
* **Positive control (patched):** both entry points return immediately and the 2 assertions pass.


fredkiefer:
I was too fast in merging the first three PR. Now the test suite fails. Could you please look into that? I will only have time to look later today.

fredkiefer (approved):
Again a great fix

DTW-Thalion:
Thanks, and sorry for the breakage. I traced it: the CI runners build with `libicns-dev` installed, so `HAVE_LIBICNS` is defined and decoding goes through the **system libicns** library rather than the built-in reader the #441/#442 fixtures target. libicns rejects those minimal hand-built icons (returns `nil`), which is what failed the positive-control assertions — the security fixes themselves are inside `#if !HAVE_LIBICNS` and aren't even compiled in that configuration.

I've opened #444 with a tests-only fix (the crafted/truncated inputs just need to decode memory-safely, so the tests now accept a `nil` result and otherwise check the decoded image). Validated with `gnustep-tests` both with and without libicns — 7/7 in both. Once #444 is in, a re-run of this PR's CI should be green. This PR's own change is unaffected.

fredkiefer (approved):
As we have a new PR to resolve the test issues, I will merge this one, even though the tests fail.

PR #447, opened 2026-06-28
NSParagraphStyle: heap-allocate decoded tab-stop arrays (stack overflow)

DTW-Thalion:
Fixes a stack overflow in `-[NSParagraphStyle initWithCoder:]` (`NSParagraphStyle.m`).

The non-keyed decoding path read a tab-stop count from the (untrusted) archive and then allocated two stack VLAs sized by it:

```objc
[aCoder decodeValueOfObjCType: @encode(NSUInteger) at: &count];
...
float locations[count];
NSTextTabType types[count];
```

The VLAs are allocated *before* the array data is read, so a large `count` value overflows the stack immediately. Reachable via `[NSUnarchiver unarchiveObjectWithData:]` on a crafted non-keyed archive.

**Reproduced:** a fake non-keyed coder returning `count = 4000000` SIGSEGVs the unpatched decoder. Fix: heap-allocate `locations`/`types` (with a NULL check) and free them afterwards.

### Validation (gnustep-tests, Ubuntu 24.04 + clang 18)
`Tests/gui/NSParagraphStyle/paragraphStyleCoderTabStops.m`:
- a fake non-keyed coder feeding a 4,000,000 tab-stop count aborts the unpatched decoder (negative control) and is handled safely after the fix;
- a real paragraph style with tab stops still round-trips through an `NSArchiver`/`NSUnarchiver` (non-keyed) archive, preserving its tab stops.


fredkiefer (approved):
I don't like the test with the fake decode We should rather craft a broken archive and test against that. But I see that this is rather hard to build.

DTW-Thalion:
I will try to handcraft a broken archive and come back with a replacement test using that.

PR #449, opened 2026-06-30
Add code for nibarchive reading

gcasa:
The NIBArchive file format is different from the XML nib format.  This patch should allow for support of it.  This is based on https://github.com/michaelwright235/nibarchive.


gcasa:
AI DISCLOSURE: I have used AI in the code review process to identify style and other issues with the code to help ensure better quality commits and also to help with documentation.

gcasa:
No nib archive is going to be created without a xib.  While it's an interesting exercise this pr is not needed

PR #452, opened 2026-07-01
Fix NSDataLink so that the tests pass

gcasa:
Fix some issues with NSDataLink the test should now pass.   Also make sure the class works as specified.

### DIFF
--- a/Source/NSDataLink.m
+++ b/Source/NSDataLink.m
@@ -27,9 +27,11 @@
 */ 
 
 #include "config.h"
+#import <Foundation/NSArray.h>
 #import <Foundation/NSFileManager.h>
 #import <Foundation/NSArchiver.h>
 #import <Foundation/NSData.h>
+#import <Foundation/NSDate.h>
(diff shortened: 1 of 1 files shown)


Source/NSDataLink.m
```diff
@@ -490,23 +561,31 @@ - (id) initWithCoder: (NSCoder*)aCoder
 	  [aCoder decodeValueOfObjCType: @encode(int) at: &_linkNumber];
 	  [aCoder decodeValueOfObjCType: @encode(int) at: &_disposition];
 	  [aCoder decodeValueOfObjCType: @encode(int) at: &_updateMode];
-	  [aCoder decodeValueOfObjCType: @encode(id)  at: &_lastUpdateTime];
+	  [aCoder decodeValueOfObjCType: @encode(id)  at: &obj];
```
fredkiefer, line 564: I think this call already retains the object. You could switch to `decodeObject:`, where there is an additional autorelease.

rfm:
Already superceded by improved fixes.

PR #454, opened 2026-07-04
tests: add NSColor tests

DTW-Thalion:
Adds `Tests/gui/NSColor` (there were no NSColor tests), covering:

- the RGB, HSB, CMYK and white component accessors, and device-colour component clamping to [0,1]
- the RGB→HSB and HSB→RGB conversions at the primary hues, and an RGB→HSB→RGB round trip
- CMYK→RGB conversion (cyan, black) and CMYK component access
- `colorWithAlphaComponent:` (replacement and clamping)
- `blendedColorWithFraction:ofColor:` (midpoint and the 0/1 endpoints)
- `colorUsingColorSpaceName:` (calibrated↔device RGB, and same-space returning the receiver)
- `isEqual:` and `colorSpaceName`

Test-only; 35 assertions, all passing.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,12 @@
+2026-07-03 Todd White <todd.white@thalion.global>
+
+	* Tests/gui/NSColor/basic.m:
+	* Tests/gui/NSColor/TestInfo:
+	Add tests for NSColor: the RGB, HSB, CMYK and white component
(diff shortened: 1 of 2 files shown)


fredkiefer:
I expect that some of the conversion tests will fail on Apple, as they actually depend on colour device context which we don't implement correctly.  But this is not the fault of the tests.

PR #455, opened 2026-07-04
NSGradient: fix inverted interpolatedColorAtLocation:

DTW-Thalion:
`-[NSGradient interpolatedColorAtLocation:]` measured the blend fraction from the *upper* stop of a segment rather than the lower one:

```objc
float fraction = (_locations[i] - location) / (_locations[i] - _locations[i - 1]);
return [c1 blendedColorWithFraction: fraction ofColor: c2];   // c1 = lower stop, c2 = upper
```

Since `blendedColorWithFraction:0` is the receiver (`c1`) and `:1` is the other colour (`c2`), at `location == locations[i-1]` the fraction is 1 and it returns `c2` — the wrong stop. The result is mirrored within each segment; it happened to be correct only exactly at the stops and at the midpoint of a two-stop gradient. For a red→blue gradient, `interpolatedColorAtLocation: 0.25` returned three-quarters blue instead of three-quarters red.

Measuring the fraction from the lower stop fixes it:

```objc
float fraction = (location - _locations[i - 1]) / (_locations[i] - _locations[i - 1]);
```

Adds `Tests/gui/NSGradient` (there were no NSGradient tests) covering construction, colour stops, and the location interpolation; the interpolation assertions fail before this change and pass after. 21 assertions.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,15 @@
+2026-07-03 Todd White <todd.white@thalion.global>
+
+	* Source/NSGradient.m (-[NSGradient interpolatedColorAtLocation:]):
+	Measure the blend fraction from the lower stop rather than the upper
+	one.  As written the interpolated colour was mirrored within each
+	segment -- correct only at the stops and the exact midpoint -- so a
+	location nearer one stop returned the colour of the other.
+	* Tests/gui/NSGradient/basic.m:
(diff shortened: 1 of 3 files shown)


fredkiefer (approved):
Looks correct, thank you!

PR #456, opened 2026-07-04
tests: add NSAffineTransform tests

DTW-Thalion:
Adds `Tests/gui/NSAffineTransform` (there were no tests for these), covering the AppKit/GNUstep additions to NSAffineTransform:

- `rotationAngle` and `isRotated` (identity, 90°, 30°, and a pure scale)
- `setFrameRotation:` setting an absolute angle
- `setMatrix:`/`getMatrix:` round-trip, cross-checked against `transformPoint:`
- `deltaPointInMatrixSpace:` (scaled but not translated)
- `concatenateWithMatrix:` composition
- `boundingRectFor:result:` for the identity, a translation, and a 90° rotation (extents swapped)
- `makeIdentityMatrix` and `takeMatrixFromTransform:`

Test-only; 16 assertions, all passing.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,13 @@
+2026-07-03 Todd White <todd.white@thalion.global>
+
+	* Tests/gui/NSAffineTransform/basic.m:
+	* Tests/gui/NSAffineTransform/TestInfo:
+	Add tests for the AppKit/GNUstep additions to NSAffineTransform:
+	rotationAngle, isRotated, setFrameRotation:, setMatrix:/getMatrix:,
+	deltaPointInMatrixSpace:, concatenateWithMatrix:, boundingRectFor:result:
+	(including a rotated rect), makeIdentityMatrix and
+	takeMatrixFromTransform:.
+
 2026-06-27 DTW-Thalion <todd.white@thalion.global>
 
 	* Source/GSTable.m (-[GSTable initWithCoder:]): Validate the row and
--- a/Tests/gui/NSAffineTransform/basic.m
+++ b/Tests/gui/NSAffineTransform/basic.m
@@ -1,0 +1,134 @@
+/* Coverage for the AppKit/GNUstep additions to NSAffineTransform:
+ * rotationAngle, isRotated, boundingRectFor:result:, the delta transform,
+ * concatenateWithMatrix:, setMatrix:/getMatrix: and the frame helpers.
+ */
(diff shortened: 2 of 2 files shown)


DTW-Thalion:
Pushed a fix for the GCC CI failure. The cause wasn't the test logic but the runtime: several of these NSAffineTransform additions don't work under the traditional GNU runtime. The ones taking a `const float[6]` selector (`setMatrix:`/`getMatrix:`/`concatenateWithMatrix:`) abort the process with `unknown type ...` when sent, and `setFrameRotation:` comes back as unrecognized. Both clang/libobjc2 jobs are fine.

I've guarded the set behind `__OBJC2__` so it runs fully on the modern runtime and is skipped on the traditional one, which gets the test green without papering over anything in the additions themselves.

Flagging the underlying runtime behaviour separately in case it's worth a look on the library side — happy to open an issue with the details if that's useful.

fredkiefer:
The failures from gcc are indeed very strange. Although all of these are deprecated methods, I really would like to understand why they are failing. The easiest way to do so is to merge your changes and enable gcc locally for debugging the issue.

DTW-Thalion:
Agreed it's worth understanding — I dug in a bit before pushing the guard, in case this saves you setup time.

The abort is specifically the `const float[6]` selectors (`setMatrix:`/`getMatrix:`/`concatenateWithMatrix:`): under the CI's gcc runtime the message send prints `unknown type rf]16` and aborts — the runtime's type parser choking on the array-argument encoding. `setFrameRotation:` (a plain `CGFloat`) comes back as `does not recognize` in the same run, which I couldn't fully account for.

The telling part is that it looks version-dependent. A ~30-line standalone repro — a class with `- (void) setMatrix: (const float[6])m` sent via `gcc … -lobjc`, no GNUstep — aborts against the older libobjc the CI installs (libobjc-10) but runs clean against a current libobjc (14.2 here), so newer runtimes appear to handle the array encoding. That fits only the pinned-gcc job tripping.

Happy to paste the repro here or open an issue with the details — whichever is more useful. Either way the guard keeps the set exercised on the modern runtime.

fredkiefer:
Thank you for explaining this. Most likely now my local investigation will be futile as I only have the most current gcc installed. As a way forward I suggest we document the limitation in the test, update the gcc/libobjc version use in CI and re-enable this test. But this can happen later, after merging your changes.

PR #458, opened 2026-07-04
NSTextList: start alphabetic markers at the first letter

DTW-Thalion:
`-[NSTextList markerForItemNumber:]` added the whole item number to `'a'`/`'A'` for the alphabetic formats:

```objc
[s replaceOccurrencesOfString: @"{lower-alpha}"
   withString: [NSString stringWithFormat: @"%c", item + 'a'] ...
```

Item numbers are 1-based — the starting item number defaults to 1, and `{decimal}` maps item 1 to `"1"` — so the first item of an alphabetic list came out as `"b"`/`"B"` instead of `"a"`/`"A"` (item 2 → `"c"`, etc.). This affected `{lower-alpha}`, `{lower-latin}`, `{upper-alpha}` and `{upper-latin}`.

Subtracting one maps item 1 to the first letter, consistent with `{decimal}`:

```objc
[NSString stringWithFormat: @"%c", item - 1 + 'a']
```

Adds `Tests/gui/NSTextList` (there were no NSTextList tests) covering the accessors and `markerForItemNumber:` for the numeric, alphabetic and glyph formats; the alphabetic assertions fail before this change and pass after. 20 assertions.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,15 @@
+2026-07-03 Todd White <todd.white@thalion.global>
+
+	* Source/NSTextList.m (-[NSTextList markerForItemNumber:]): Start the
+	alphabetic markers at the first letter.  Item numbers are 1-based (the
+	starting item number defaults to 1, and {decimal} maps item 1 to "1"),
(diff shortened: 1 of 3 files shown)


fredkiefer (approved):
Thank you

PR #459, opened 2026-07-04
tests: add NSTextTab tests

DTW-Thalion:
Adds `Tests/gui/NSTextTab` (there were no NSTextTab tests), covering:

- `initWithType:location:` mapping each tab stop type (left, right, center, decimal) to its alignment
- `initWithTextAlignment:location:options:` mapping each alignment back to a tab stop type, including the decimal tab produced from a right alignment carrying `NSTabColumnTerminatorsAttributeName`
- the `location`, `tabStopType`, `alignment` and `options` accessors
- `compare:` ordering by location (ascending, descending, same, self)
- `isEqual:`/`hash` (by type and location)
- `copy` preserving location, type, alignment and options

Test-only; 23 assertions, all passing.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,11 @@
+2026-07-03 Todd White <todd.white@thalion.global>
+
+	* Tests/gui/NSTextTab/basic.m:
+	* Tests/gui/NSTextTab/TestInfo:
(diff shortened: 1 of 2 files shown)


fredkiefer (approved):
Thank you

PR #463, opened 2026-07-12
tests: cover NSGradient custom stop locations

DTW-Thalion:
basic.m exercises only evenly spaced stops with the first at 0 and the last
at 1. This adds unevenly spaced stops via initWithColorsAndLocations: and
initWithColors:atLocations:colorSpace:, interpolation across segments of
unequal length, and clamping to the first or last colour when a location
falls outside a stop range that does not span 0 to 1.


fredkiefer (approved):
I like that you are sticking with spelling "colour" that Richard and I prefer :-)

DTW-Thalion:
Shhhh, I'm Canadian.

gcasa:
Yes, people from "Canadia" do spell things weird. ;) (kidding)

DTW-Thalion:
A lifetime spent translating between HM's English and American ... woe is me.

PR #464, opened 2026-07-12
tests: cover the NSButtonCell state and value coupling

DTW-Thalion:
A button cell has no separate value: its int, object and string value are
derived from the cell state, and setting any of them sets the state. This
covers setIntValue reducing to a state (non-zero to on, so 5 reads back 1),
the state driving int/object/float/string value, mixed state carried
through the value as -1, and setObjectValue mapping nil, numbers and empty
strings to the state. The existing NSButtonCell test only covers encoding.

### DIFF
--- a/Tests/gui/NSButtonCell/stateValue.m
+++ b/Tests/gui/NSButtonCell/stateValue.m
@@ -1,0 +1,91 @@
+/* Tests the coupling between an NSButtonCell's value and its state.  A button
+ * cell has no separate value: its int/object/string value is derived from the
+ * cell state, and setting any of them sets the state.  These are plain value
+ * operations on a cell.
+ */
+#include "Testing.h"
+
+#include <Foundation/NSAutoreleasePool.h>
+#include <Foundation/NSValue.h>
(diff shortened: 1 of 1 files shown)


Tests/gui/NSButtonCell/stateValue.m
```diff
@@ -0,0 +1,91 @@
...
+  /* The string value follows the state, and setting it sets the state from
+   * whether the string is empty. */
+  [cell setState: NSOnState];
+  pass([[cell stringValue] isEqualToString: @"1"],
+       "the string value of an on cell is 1");
+  [cell setStringValue: @"x"];
+  pass([cell state] == NSOnState, "a non-empty string value sets the on state");
+  [cell setStringValue: @""];
+  pass([cell state] == NSOffState, "an empty string value sets the off state");
+
```
fredkiefer, line 86: Here a RELEASE of the cell is missing

DTW-Thalion: Fixed, the cell is now autoreleased.

fredkiefer (approved):

PR #465, opened 2026-07-12
tests: cover the NSSliderCell tick mark values

DTW-Thalion:
minMax.m covers value clamping to the minimum and maximum. This covers the
tick mark values, which it does not: the tick mark count,
tickMarkValueAtIndex: spacing and its out-of-range exception, and
closestTickMarkValueToValue: snapping a value to the nearest tick (rounding
halves towards positive infinity and clamping out-of-range values to the
end ticks), plus the no-tick and single-tick cases.

### DIFF
--- a/Tests/gui/NSSliderCell/tickMarks.m
+++ b/Tests/gui/NSSliderCell/tickMarks.m
@@ -1,0 +1,117 @@
+/* Tests the NSSliderCell tick mark value logic, which minMax.m does not cover:
+ * the tick mark count, tickMarkValueAtIndex:, and closestTickMarkValueToValue:
+ * (which snaps a value to the nearest tick and clamps out-of-range values to
+ * the end ticks).  These are plain value operations on a cell.
+ */
+#include "Testing.h"
+
+#include <math.h>
+#include <Foundation/NSAutoreleasePool.h>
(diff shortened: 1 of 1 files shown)


Tests/gui/NSSliderCell/tickMarks.m
```diff
@@ -0,0 +1,117 @@
...
+    [c setMaxValue: 1];
+    [c setNumberOfTickMarks: 5];
+    pass(eq([c closestTickMarkValueToValue: 0.3], 0.25),
+         "a value near a quarter tick snaps to it");
+    pass(eq([c closestTickMarkValueToValue: 0.4], 0.5),
+         "a value nearer the half tick snaps to the half");
+  }
+
+  /* The restrict-to-ticks flag round-trips. */
+  [cell setAllowsTickMarkValuesOnly: YES];
```
fredkiefer: Both slider cells are never released. Also this bit of code could be moved up a bit to keep the two separate d.

DTW-Thalion: Both cells are now autoreleased, and the flag round-trip is moved up so the second cell stands on its own.

fredkiefer (approved):

PR #467, opened 2026-07-12
tests: cover the NSAttributedString attribute mutators

DTW-Thalion:
Adds the NSMutableAttributedString AppKit mutators: superscriptRange: and
subscriptRange: raising and lowering the superscript level (and cancelling
to zero), unscriptRange: removing it, their range exception, and
setAlignment:range: setting the paragraph alignment over a range while
leaving text outside the range without a paragraph style.

### DIFF
--- a/Tests/gui/NSAttributedString/mutation.m
+++ b/Tests/gui/NSAttributedString/mutation.m
@@ -1,0 +1,138 @@
+/* Tests the NSMutableAttributedString AppKit attribute mutators:
+ * superscriptRange:, subscriptRange: and unscriptRange: (which raise and lower
+ * the superscript level and remove it), and setAlignment:range: (which sets the
+ * alignment on the paragraph style over a range).  These are plain value
+ * operations.
+ */
+#include "Testing.h"
+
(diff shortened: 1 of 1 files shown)


Tests/gui/NSAttributedString/mutation.m
```diff
@@ -0,0 +1,137 @@
...
+#include <AppKit/NSApplication.h>
+#include <AppKit/NSAttributedString.h>
+#include <AppKit/NSParagraphStyle.h>
+#include <AppKit/NSText.h>
+#include <AppKit/NSStringDrawing.h>
+
+static NSMutableAttributedString *
+str(void)
+{
+  return [[NSMutableAttributedString alloc] initWithString: @"hello world"];
```
fredkiefer: You need to autorelease the string here, as you never release it in the test.

DTW-Thalion: Good catch - will fix.

DTW-Thalion: Done.

fredkiefer (approved):

PR #471, opened 2026-07-12
tests: cover NSParagraphStyle tab stop handling

DTW-Thalion:
The existing tab stop test covers archiving and a decode regression; this
covers the behaviour: the default twelve tab stops every 28 points and the
zero default interval, setTabStops: sorting the stops by location,
addTabStop: keeping them in order wherever a new stop belongs,
removeTabStop:, and setDefaultTabInterval:.

### DIFF
--- a/Tests/gui/NSParagraphStyle/tabStops.m
+++ b/Tests/gui/NSParagraphStyle/tabStops.m
@@ -1,0 +1,102 @@
+/* Tests the NSParagraphStyle tab stop handling that the coder test does not:
+ * the default tab stops and interval, addTabStop: keeping
+ * the stops in location order, removeTabStop:, and setDefaultTabInterval:.
+ * These are plain value operations.
+ */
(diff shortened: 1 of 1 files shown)


Tests/gui/NSParagraphStyle/tabStops.m
```diff
@@ -0,0 +1,102 @@
...
+locAt(NSParagraphStyle *p, NSUInteger i)
+{
+  return [[[p tabStops] objectAtIndex: i] location];
+}
+
+static NSTextTab *
+tab(CGFloat loc)
+{
+  return [[[NSTextTab alloc] initWithType: NSLeftTabStopType
+                                 location: loc] autorelease];
```
fredkiefer, line 30: It would be great to use the AUTORELEASE macro here.

fredkiefer (approved):

PR #472, opened 2026-07-12
tests: add NSStepperCell tests

DTW-Thalion:
NSStepperCell had no test coverage.  This adds tests for its value-logic
surface: the -init defaults, the increment/autorepeat/valueWraps
accessors, the clamping done by -setMaxValue:/-setMinValue:/-setObjectValue:
(including the fallback to the minimum for a non-numeric or nil object and
when the minimum exceeds the maximum), and the keyed and non-keyed
NSCoding round trips.  No source changes; all assertions pass.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,13 @@
+2026-07-12 Todd White <todd.white@thalion.global>
+
+	* Tests/gui/NSStepperCell/valueLogic.m:
+	* Tests/gui/NSStepperCell/TestInfo:
+	Add tests for NSStepperCell: the defaults, the increment, autorepeat
+	and valueWraps accessors, the value clamping in setMaxValue:,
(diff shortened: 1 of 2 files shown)


valueLogic.d
```diff
@@ -0,0 +1,3 @@
+valueLogic.o: Tests/gui/NSStepperCell/valueLogic.m \
```
fredkiefer: Could you please explain this file?

DTW-Thalion: Sorry - that appears to be an artifact from my dev-env - let me be sure, then fix.

DTW-Thalion: It was a make dependency file left over from building the test, committed by mistake. Removed.

fredkiefer (approved):

PR #473, opened 2026-07-12
tests: add NSPrintInfo tests

DTW-Thalion:
NSPrintInfo only had the sharedPrintInfo smoke test.  This adds tests for
the dictionary-backed accessors: the defaults from -initWithDictionary:,
the paper name/size relationship, the way -setPaperSize: and
-setOrientation: keep the size and orientation consistent, the margin,
pagination, centering and jobDisposition accessors, the
-initWithDictionary: overlay and -copy independence.  Paper dimensions
are checked against +sizeForPaperName: rather than hard-coded, since they
come from the default printer.  No source changes; all assertions pass.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,12 @@
+2026-07-12 Todd White <todd.white@thalion.global>
+
+	* Tests/gui/NSPrintInfo/dictionary.m:
+	Add tests for NSPrintInfo: the defaults from initWithDictionary:, the
+	paper name/size relationship, the way setPaperSize: and setOrientation:
+	keep the size and orientation consistent, the margin, pagination,
+	centering and jobDisposition accessors, the initWithDictionary: overlay
+	and copy independence.
+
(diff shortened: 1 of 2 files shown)


dictionary.d
```diff
@@ -0,0 +1,3 @@
+dictionary.o: Tests/gui/NSPrintInfo/dictionary.m \
```
fredkiefer: Again, what is this file for?

DTW-Thalion: Same as the other one: a make dependency file left over from the build. Removed.

fredkiefer: There are more of these files in the other PRs. Could you please clean them up as well?

fredkiefer (approved):

dictionary.d
DTW-Thalion: Yes, on it. A mistake in a macro I just made in JetBrains, my regex was off.

PR #476, opened 2026-07-12
NSSecureTextFieldCell: echo bullets by default

DTW-Thalion:
A freshly created NSSecureTextFieldCell left echosBullets at NO because no
initialiser set the ivar.  A cell used on its own therefore drew nothing
where it should have drawn bullets; only NSSecureTextField hid this by
setting the flag itself.

Default echosBullets to YES in -initTextCell:, matching NSSecureTextField
and Apple.  Verified on a macOS runner: a standalone
[[NSSecureTextFieldCell alloc] init] reports echosBullets == YES there.

Adds a test for the default.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,15 @@
 2026-07-12 Todd White <todd.white@thalion.global>
 
+	* Source/NSSecureTextField.m (-[NSSecureTextFieldCell initTextCell:]):
+	Default echosBullets to YES so a freshly created cell echoes bullets,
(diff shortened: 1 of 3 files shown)


fredkiefer (approved):

fredkiefer:
Not sure whether we also need to change the implementation of the -initWithCoder: method. And what about these methods in the NSSecureTextField class? Should we remove these?
I merge anyway and we think about that later.

PR #477, opened 2026-07-12
NSButtonCell: string value follows the state

DTW-Thalion:
`-[NSButtonCell stringValue]` returned `_cell.state ? @"1" : @""`, so an
off cell gave the empty string and a mixed cell gave `"1"`.  A button
cell's value is its state, so the string value should be the string form
of the state.  Return that, giving `"1"` / `"0"` / `"-1"` for on / off /
mixed.  Verified on a macOS runner: NSButtonCell reports `"1"`, `"0"` and
`"-1"` for the three states.

Adds a test for the string value in each state.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,11 @@
+2026-07-12 Todd White <todd.white@thalion.global>
+
+	* Source/NSButtonCell.m (-[NSButtonCell stringValue]): Return the
+	string form of the state ("1", "0" or "-1").  An off cell gave "" and a
+	mixed cell "1"; both now match the state as OS X does.
+	* Tests/gui/NSButtonCell/stringValue.m:
+	New test for the string value in each state.
+
 2026-07-03 Todd White <todd.white@thalion.global>
 
 	* Tests/gui/NSTextTab/basic.m:
--- a/Source/NSButtonCell.m
+++ b/Source/NSButtonCell.m
@@ -853,7 +853,7 @@
 
 - (NSString *) stringValue
 {
-  return _cell.state ? @"1" : @"";
+  return [NSString stringWithFormat: @"%ld", (long)_cell.state];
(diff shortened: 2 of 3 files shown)


fredkiefer:
Could you please test this on MacOS? It sounds plausible to me, but we rather should make sure.

DTW-Thalion:
> Could you please test this on MacOS? It sounds plausible to me, but we rather should make sure.

I ran it on the MacOS runner (GitHub container) - but not actual hardware, will need someone else to do that if you want to test further.

DTW-Thalion:
I am running the test harnesses on the MacOS runner using their AppKit, then comparing expected outputs as part of putting together these tests - sometimes GNUStep is a bit more strict or there are some differing defaults that are GNUStep specific, but I use that as the basis of most of the bug fixes.

gcasa:
> I am running the test harnesses on the MacOS runner using their AppKit, then comparing expected outputs as part of putting together these tests - sometimes GNUStep is a bit more strict or there are some differing defaults that are GNUStep specific, but I use that as the basis of most of the bug fixes.

You beat me to it.   I just refactored your test inside of a quick test app.  The tests pass on macOS.  The code is here, it's derived from your test.  https://github.com/gcasa/NSButtonStringValue_test

GC

DTW-Thalion:
> > I am running the test harnesses on the MacOS runner using their AppKit, then comparing expected outputs as part of putting together these tests - sometimes GNUStep is a bit more strict or there are some differing defaults that are GNUStep specific, but I use that as the basis of most of the bug fixes.
> 
> You beat me to it. I just refactored your test inside of a quick test app. The tests pass on macOS. The code is here, it's derived from your test. https://github.com/gcasa/NSButtonStringValue_test
> 
> GC

Thank you sir!

PR #478, opened 2026-07-12
NSSliderCell: break tick ties towards the lower tick

DTW-Thalion:
`-[NSSliderCell closestTickMarkValueToValue:]` snapped exact halves to
the upper tick (2.5 -> 3 with integer ticks over 0..10).  OS X breaks
ties towards the lower tick (2.5 -> 2, 7.5 -> 7).  Snap ties down to
match, leaving `GSRoundTowardsInfinity` (shared with other classes)
untouched.  Verified on a macOS runner.

Adds a test for the tie rounding.

### DIFF
--- a/Source/NSSliderCell.m
+++ b/Source/NSSliderCell.m
@@ -860,7 +860,9 @@ double _doubleValueForMousePoint (NSPoint point, NSRect knobRect,
 
   d = _maxValue - _minValue;
   f = ((aValue - _minValue)  * (effectiveTicks - 1)) / d;
-  f = ((GSRoundTowardsInfinity(f) * d) / (effectiveTicks - 1)) + _minValue;
+  /* Snap to the nearest tick, breaking ties towards the lower tick to
+     match OS X. */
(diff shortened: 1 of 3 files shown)


Source/NSSliderCell.m
```diff
@@ -860,7 +860,9 @@ - (double) closestTickMarkValueToValue: (double)aValue
 
   d = _maxValue - _minValue;
   f = ((aValue - _minValue)  * (effectiveTicks - 1)) / d;
-  f = ((GSRoundTowardsInfinity(f) * d) / (effectiveTicks - 1)) + _minValue;
+  /* Snap to the nearest tick, breaking ties towards the lower tick to
+     match OS X. */
+  f = ((-GSRoundTowardsInfinity(-f) * d) / (effectiveTicks - 1)) + _minValue;
```
fredkiefer, line 865: Why didn't you use GSRoundTowardsNegativeInfinity? I think that function would be easier to understand.

fredkiefer (approved):

PR #487, opened 2026-07-12
NSPopUpButtonCell: keep the correct selection when removing an item

DTW-Thalion:
`-[NSPopUpButtonCell removeItemAtIndex:]` mishandled the selection in two
ways:

1. When the selected item was removed, the popup was left with no
selection.  OS X selects the first remaining item (and clears the
selection only when the list becomes empty).

2. When a different item *before* the selection was removed, the removal
posts a notification that re-syncs the selection from the menu view's
highlighted index — which does not track the removed item — so the
selection slid to a different item.  OS X keeps the selection on the same
item (its index just shifts down).

Select the first item when the selected item is removed, and otherwise
re-select the item that was selected.  Verified on a macOS runner across
all four cases (remove selected middle/last, remove an earlier item,
remove the only item).

Adds a test for the selection after removing an item.

### DIFF
--- a/Source/NSPopUpButtonCell.m
+++ b/Source/NSPopUpButtonCell.m
@@ -460,12 +460,33 @@ static NSImage *_pbc_image[5];
  */
 - (void) removeItemAtIndex: (NSInteger)index
 {
-  if (index == [self indexOfSelectedItem])
+  BOOL removingSelected = (index == [self indexOfSelectedItem]);
+  id <NSMenuItem> keepSelected = removingSelected ? nil : _selectedItem;
+
+  if (removingSelected)
     {
       [self selectItem: nil];
     }
 
   [_menu removeItemAtIndex: index];
+
+  if (removingSelected)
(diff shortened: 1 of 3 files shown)


Source/NSPopUpButtonCell.m
```diff
@@ -460,12 +460,33 @@ - (void) removeItemWithTitle: (NSString *)title
...
+      /* When the selected item is removed, select the first remaining item
+         rather than leaving the popup with no selection, as OS X does.  */
+      if ([_menu numberOfItems] > 0)
+        {
+          [self selectItemAtIndex: 0];
+        }
+    }
+  else if (keepSelected != nil)
+    {
+      /* Removing an item posts a notification that re-syncs the selection
```
fredkiefer, line 484: Most likely it is caused by a bug in the menu implementation that we need this code here. But for now it is OK.

fredkiefer (approved):

PR #489, opened 2026-07-12
NSTableColumn: keep the resizable flag and resizing mask consistent

DTW-Thalion:
`-[NSTableColumn isResizable]` and the resizing mask were independent:
`setResizingMask:` did not affect `isResizable`, and `setResizable:` did
not touch the mask (the existing code had a TODO noting this).  OS X keeps
the two consistent -- `isResizable` reflects whether the mask allows any
resizing, and `setResizable:` sets the mask.

Make `setResizable:` set the mask to the auto/user resizing combination
(YES) or to no resizing (NO), and `setResizingMask:` update the resizable
flag from whether the mask allows any resizing.  `isResizable` stays the
value serialised by the coder, which both setters now keep in step with
the mask.  Verified on a macOS runner across the four combinations.

Adds a test for the resizable/mask relationship.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,16 @@
 2026-07-12 Todd White <todd.white@thalion.global>
 
+	* Source/NSTableColumn.m (-[NSTableColumn setResizable:], -[NSTableColumn
+	setResizingMask:]): Keep the resizable flag and the resizing mask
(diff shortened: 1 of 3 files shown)


fredkiefer (approved):
Quentin would be happy to know that this was finally changed. I don't remember why he didn't do it 17 years ago.

PR #492, opened 2026-07-12
NSMatrix: no cell class when a prototype is used

DTW-Thalion:
`-[NSMatrix cellClass]` returned the prototype cell's class when the
matrix was built with a prototype.  A matrix's cell class and its
prototype are alternatives, and OS X returns nil from `cellClass` while a
prototype is set.  Return `Nil` in that case.  The stored class is left
untouched (it is still what `setPrototype:` records for the archiver and
is unused for cell creation while a prototype is set), so the coder and
cell creation are unaffected.  Verified on a macOS runner: a
prototype-built matrix reports no cell class.

Adds a test for the cell class / prototype relationship.

### DIFF
--- a/Tests/gui/NSMatrix/cellClassPrototype.m
+++ b/Tests/gui/NSMatrix/cellClassPrototype.m
@@ -1,0 +1,66 @@
+/* A matrix that uses a cell prototype has no cell class, as on OS X:
+   -cellClass returns Nil while a prototype is set, and returns the class
+   again once a cell class is set (which clears the prototype). */
+#include "Testing.h"
+
+#include <Foundation/NSAutoreleasePool.h>
+#include <Foundation/NSGeometry.h>
+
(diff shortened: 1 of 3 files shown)


Tests/gui/NSMatrix/cellClassPrototype.m
```diff
@@ -0,0 +1,66 @@
...
+                                             mode: NSListModeMatrix
+                                        prototype: proto
+                                     numberOfRows: 1
+                                  numberOfColumns: 1]);
+
+    pass([m prototype] != nil, "the prototype is set");
+    pass([m cellClass] == Nil, "a matrix with a prototype has no cell class");
+  }
+
+  /* Setting a cell class clears the prototype and reports the class. */
```
fredkiefer, line 44: As we never set the prototype, this does not clear it.

fredkiefer (approved):

PR #500, opened 2026-07-12
NSTrackingArea: store the userInfo and use the AppKit option values

DTW-Thalion:
Two AppKit-alignment fixes in NSTrackingArea, found by adding coverage for it.

initWithRect:options:owner:userInfo: did `_userInfo = RETAIN(_userInfo)`, retaining the still-nil ivar instead of the userInfo argument, so -userInfo always returned nil even though the dictionary was passed in.

The NSTrackingAreaOptions constants were numbered contiguously from 0x08 (ActiveWhenFirstResponder=0x08, ActiveInKeyWindow=0x10, ... EnabledDuringMouseDrag=0x200). AppKit leaves 0x08 unused: the active options start at 0x10 and run 0x10/0x20/0x40/0x80, then 0x100/0x200/0x400. Verified against Cocoa on macOS. The only in-tree use is the NSTrackingAssumeInside check in initWithRect:, which reads the named constant and stays correct.

The new test covers the option values, the rect/options/owner/userInfo accessors, a nil userInfo and copy.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,17 @@
 2026-07-12 Todd White <todd.white@thalion.global>
 
+	* Headers/AppKit/NSTrackingArea.h: Give the NSTrackingAreaOptions
+	constants their AppKit values; the active and later options were off
+	by the unused 0x08 bit.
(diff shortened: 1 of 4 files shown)


fredkiefer (approved):
Nice catch!

PR #504, opened 2026-07-12
NSTextBlock: give NSTextBlockDimension its AppKit values

DTW-Thalion:
NSTextBlockDimension numbered the height dimensions one short. AppKit leaves the value 3 unused: NSTextBlockHeight is 4, NSTextBlockMinimumHeight 5 and NSTextBlockMaximumHeight 6. Verified against Cocoa on macOS. The backing _value/_valueType arrays are sized from NSTextBlockMaximumHeight, so they grow to seven and the existing named-constant indexing keeps working.

Stacked on #503 (adds an assertion to its NSTextTable test); merge that first. Will rebase to an incremental diff once it lands.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,12 @@
 2026-07-12 Todd White <todd.white@thalion.global>
 
+	* Headers/AppKit/NSTextTable.h: Give the NSTextBlockDimension constants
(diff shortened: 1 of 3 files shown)


fredkiefer (approved):
i don't understand why I don't get a conflict here.

PR #508, opened 2026-07-13
NSColorSpace: return ICC profile data for the generic RGB space (prototype)

DTW-Thalion:
Prototype. NSColorSpace stored ICC data only when one was passed to initWithICCProfileData:; the standard colour spaces returned nil, where AppKit returns a profile for the generic spaces.

This adds the ICC sRGB v2 reference profile (sRGB2014.icc) as a gui library resource and loads it for +genericRGBColorSpace, so -ICCProfileData returns a valid profile. The profile is the International Color Consortium reference profile from https://registry.color.org/rgb-registry/srgbprofiles, redistributed unmodified under the ICC licence ("may be copied, distributed, embedded, made, used, and sold without restriction"); its provenance and licence are recorded in Resources/ColorSpaces/README. No Apple profile data is used (for obvious reasons).

Scope: only the generic RGB space is wired up, and this is deliberate. I looked at extending it to the generic gray and CMYK spaces and found no cleanly-licensed ICC-registry source for either:
- The ICC registry has no generic grayscale profile.
- Its CMYK profiles are all third-party (APTEC, ECI, Idealliance) with no stated permissive licence, so they are not safe to embed here. The only ICC-owned CMYK-adjacent profile, PRMG, is a reference-gamut exchange profile rather than a CMYK output space, so it is the wrong fit for genericCMYK.

So gray and CMYK are left returning nil for now. They could be added later either from a suitably-licensed profile or by synthesising minimal profiles, which is a maintainer call. The device spaces are also left without a profile, matching AppKit.

Whether sRGB is the right mapping for the generic RGB space, and whether to take on colour management more generally (e.g. via lcms2), is a design decision for the GNUStep team, but this at least puts the fraamework in place using a fully permissive licensed file.

Stacked on #507 (adds an assertion to its NSColorSpace test); merge that first.


rmottola:
Interesting.

Would these profiles used somewhere? Usually they are considered when loading images, displaying, printing. To my knowledge, currently internally GNUstep doesn't use them

How can I "test" this implementation? loading images with specific profiles?

I wonder if this partial implementation could cause harm.

DTW-Thalion:
> Interesting.
> 
> Would these profiles used somewhere? Usually they are considered when loading images, displaying, printing. To my knowledge, currently internally GNUstep doesn't use them
> 
> How can I "test" this implementation? loading images with specific profiles?
> 
> I wonder if this partial implementation could cause harm.

Unless you were initialising the NSColorSpace specifically with initWithICCProfileData:; it would have no effect on you. MacOS supports ICC Color Spaces for graph designers and other production arts uses. I added this for AppKit API compatibility and wanted to put in an actual implementation (not just a stub), so I grabbed the open source ICC data for RGB as a default.

Presumably you could take ICC data from any Mac machine (if you run a Mac) and use that as your profile data or if you had some ICC data from some other source.  Where this is a GNU project I stuck with the most permissive data directly from the ICC. but can make an effort to provide more profiles (open source).

AS for testing it - yes, if you have ICC color profiles available, call the method with a ICC source and you should eb able to see it work - if not, I have some bug fixes to do ! :)

fredkiefer (approved):

PR #517, opened 2026-07-15
NSLevelIndicatorCell: add the AppKit levelIndicatorStyle getter

DTW-Thalion:
AppKit's getter for the style is `-levelIndicatorStyle`; GNUstep only had the non-AppKit `-style`. Add `-levelIndicatorStyle` (keeping `-style`).

### DIFF
--- a/Headers/AppKit/NSLevelIndicatorCell.h
+++ b/Headers/AppKit/NSLevelIndicatorCell.h
@@ -72,6 +72,7 @@ APPKIT_EXPORT_CLASS
 
 
 - (NSLevelIndicatorStyle)style;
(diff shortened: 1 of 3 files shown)


fredkiefer (approved):
I am not sure, whether this method was called that way originally. Probably it is best to keep it, but flag it as a GNUstep extension in the header file.

DTW-Thalion:
So all of the tests are being run and compared to the outputs from a MacOS oracle runner. I am mimicking the behavior of the Mac (as best I can), so unless it is obvious that GNUStep is architecturally designed differently, the goal is to be MacOS API surface equivalent.

PR #519, opened 2026-07-15
NSDatePickerCell: clamp the date value to minDate and maxDate

DTW-Thalion:
NSDatePickerCell stored the date value unchanged; AppKit clamps it to `[minDate, maxDate]` when it is set and re-clamps it when the bounds later move past it. Clamp in `-setDateValue:` and re-clamp in `-setMinDate:`/`-setMaxDate:`.

### DIFF
--- a/Source/NSDatePickerCell.m
+++ b/Source/NSDatePickerCell.m
@@ -30,6 +30,7 @@
 */
 
 #import <Foundation/NSString.h>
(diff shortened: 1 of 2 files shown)


fredkiefer (approved):
Good catch, maybe we also need to ensure that minDate <= maxDate?

PR #525, opened 2026-07-15
NSToolbarItem: return an empty paletteLabel instead of nil by default

DTW-Thalion:
`-paletteLabel` returned nil for a fresh item; AppKit returns the empty string, the same way `-label` already does. Return `@""` when no palette label has been set.

### DIFF
--- a/Source/NSToolbarItem.m
+++ b/Source/NSToolbarItem.m
@@ -1309,7 +1309,14 @@ NSString *GSMovableToolbarItemPboardType = @"GSMovableToolbarItemPboardType";
 
 - (NSString *) paletteLabel
 {
-  return _paletteLabel;
+  if (nil != _paletteLabel)
+    {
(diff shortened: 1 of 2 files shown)


Source/NSToolbarItem.m
```diff
@@ -1309,7 +1309,14 @@ - (NSSize) minSize
 
 - (NSString *) paletteLabel
 {
-  return _paletteLabel;
+  if (nil != _paletteLabel)
```
fredkiefer, line 1312: I would prefer that we set the value in the initXXX method, but as this class has a very strange behaviour, lets stick with that for now.

fredkiefer (approved):

Source/NSToolbarItem.m
DTW-Thalion: Will fix.

PR #528, opened 2026-07-15
NSTabViewItem: start the label off as the empty string

DTW-Thalion:
A new item's label was nil, where AppKit's is the empty string. The initialiser
sets it now, rather than the getter answering for it.

Setting it in the initialiser is not only tidier here, it is the behaviour
AppKit has. Checked on a macOS runner:

    NSTabViewItem   default: empty   after setLabel:nil: nil
    NSToolbarItem   default: empty   after setLabel:nil: empty

So a tab view item's label can go back to nil once it has been set, and only
starts out empty. Answering for it in the getter would have kept the empty
string forever, which is not what AppKit does. The test covers the round-trip
and the nil.

-initWithIdentifier: is the initialiser plain -init and the keyed decode both
go through, so an item made any of those ways starts the same.

### DIFF
--- a/Source/NSTabViewItem.m
+++ b/Source/NSTabViewItem.m
@@ -121,7 +121,14 @@
 
 - (NSString *) label
 {
-  return _label;
+  if (nil != _label)
+    {
+      return _label;
+    }
+  else
+    {
+      return @"";
+    }
 }
 
 - (NSSize) sizeOfLabel: (BOOL)shouldTruncateLabel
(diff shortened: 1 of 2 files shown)


Source/NSTabViewItem.m
```diff
@@ -121,7 +121,14 @@ - (void) setLabel: (NSString*)label
 
 - (NSString *) label
 {
-  return _label;
+  if (nil != _label)
```
fredkiefer, line 124: Again, it would be better to set the value in the init method and here that would be simple.

fredkiefer (approved):

PR #542, opened 2026-07-15
Fix: NSPathControl -pathItems returns an empty array

DTW-Thalion:
On a freshly created NSPathControl with no path items, `-pathItems` returns nil under GNUstep; AppKit returns an empty array. This returns an empty array in that case.


Source/NSPathControl.m
```diff
@@ -228,6 +228,10 @@ - (NSPathControlItem *) clickedPathItem
 
 - (NSArray *) pathItems
 {
+  if (_pathItems == nil)
```
fredkiefer: Most likely it would be better to set this value in the initializer.

DTW-Thalion: Done - moved the setup into the designated initializer -initWithFrame: (which -init funnels through), so the items are populated on that path too, and dropped the getter fallback. This shares the init consolidation with #543; I'll rebase whichever lands second.

fredkiefer: It looks like something got lost here.

DTW-Thalion: Done. The initializer now allocates the empty array and -dealloc releases it.

fredkiefer (approved):

PR #543, opened 2026-07-15
Fix: NSPathControl -isEditable / -setEditable:

DTW-Thalion:
AppKit provides `-[NSPathControl isEditable]` (default YES) and `-setEditable:`; GNUstep had neither. This adds the accessors with a backing ivar defaulting to YES.

### DIFF
--- a/Source/NSPathControl.m
+++ b/Source/NSPathControl.m
@@ -79,15 +79,16 @@ static Class pathCellClass;
                                       assumeInside: YES];
 }
 
-- (instancetype) init
+- (instancetype) initWithFrame: (NSRect)frameRect
 {
-  self = [super init];
+  self = [super initWithFrame: frameRect];
   if (self != nil)
     {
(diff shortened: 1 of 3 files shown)


Source/NSPathControl.m
```diff
@@ -79,6 +79,16 @@ - (void) resetCursorRects
                                       assumeInside: YES];
 }
 
+- (instancetype) initWithFrame: (NSRect)frameRect
```
fredkiefer, line 82: Now this class has two independent initializer. Yours is actually the correct one, the other one -init should not be here at all. Best you move all the value setting over to your method and remove the old one.

DTW-Thalion: Done - consolidated the defaults into -initWithFrame: (which -init funnels through) and removed the separate -init override.

fredkiefer (approved):

PR #546, opened 2026-07-15
Fix: NSPathCell -pathComponentCells returns an empty array

DTW-Thalion:
On a freshly created NSPathCell with no components, `-pathComponentCells` returns nil under GNUstep; AppKit returns an empty array. This returns an empty array in that case.

### DIFF
--- a/Source/NSPathCell.m
+++ b/Source/NSPathCell.m
@@ -169,6 +169,10 @@ static Class pathComponentCellClass;
 
 - (NSArray *) pathComponentCells
 {
+  if (_pathComponentCells == nil)
+    {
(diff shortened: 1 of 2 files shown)


Source/NSPathCell.m
```diff
@@ -169,6 +169,10 @@ - (NSPathComponentCell *) clickedPathComponentCell
 
 - (NSArray *) pathComponentCells
 {
+  if (_pathComponentCells == nil)
```
fredkiefer, line 172: Again this belongs rather into the missing -init method.

fredkiefer (approved):

PR #549, opened 2026-07-15
Fix: NSLayoutAnchor -constraintEqualToAnchor:constant: honours the constant

DTW-Thalion:
`-[NSLayoutAnchor constraintEqualToAnchor:constant:]` passed a hardcoded `0.0` as the constant instead of the `c` argument, so the constant was silently dropped. Its `constraintGreaterThanOrEqualToAnchor:constant:` and `constraintLessThanOrEqualToAnchor:constant:` siblings already pass `c` through. This passes `c` through as well.

Verified against AppKit on macOS (a constraint built from an anchor with a constant carries that constant). Includes a test that fails before the change and passes after.

### DIFF
--- a/Source/NSLayoutAnchor.m
+++ b/Source/NSLayoutAnchor.m
@@ -72,7 +72,7 @@
                                          toItem: [anchor item]
                                       attribute: NSLayoutAttributeLeft
                                      multiplier: 1.0
-                                       constant: 0.0];
+                                       constant: c];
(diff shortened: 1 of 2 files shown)


DTW-Thalion:
Subsumed by #550, which makes NSLayoutAnchor attribute-aware and corrects this constant along the way. Closing in favour of that PR.

PR #550, opened 2026-07-16
Fix: vend layout anchors from NSView

DTW-Thalion:
NSView did not implement the NSLayoutAnchor factory properties (`leadingAnchor`, `widthAnchor`, `topAnchor` and the rest), so anchor-based Auto Layout was unavailable even though the constraint solver (`GSAutoLayoutEngine`, `addConstraint:`) was already present. NSLayoutAnchor itself carried no notion of which attribute an anchor represents and hardcoded `NSLayoutAttributeLeft` in every constraint it built.

This change:
- gives NSLayoutAnchor an attribute and an internal initialiser binding an anchor to an item and attribute;
- builds constraints using the anchor's own attribute and the other anchor's attribute instead of a hardcoded `Left`;
- makes `NSLayoutDimension`'s `constraintEqualToConstant:` (and the greater-/less-than variants) produce a constraint with no second item and `NotAnAttribute`, matching AppKit;
- adds the twelve anchor accessors to NSView (leading/trailing/left/right/centerX, top/bottom/centerY/firstBaseline/lastBaseline, width/height).

Verified against AppKit on macOS: for a view, `widthAnchor constraintEqualToConstant:50` and `leadingAnchor constraintEqualToAnchor:other.leadingAnchor constant:8` now produce constraints with the same firstItem/attribute, secondItem/attribute, relation, multiplier and constant as AppKit. Includes a test. Existing NSView tests still pass.

Note: this also corrects `constraintEqualToAnchor:constant:` dropping its constant, which is the subject of the smaller standalone PR #549; either can land first.

### DIFF
--- a/Source/NSLayoutAnchor.m
+++ b/Source/NSLayoutAnchor.m
@@ -28,16 +28,36 @@
 
 #import "AppKit/NSLayoutAnchor.h"
 #import "AppKit/NSLayoutConstraint.h"
+#import "GSAutoLayoutAnchorPrivate.h"
 
 @implementation NSLayoutAnchor
+
+- (instancetype) initWithItem: (id)item attribute: (NSLayoutAttribute)attribute
+{
+  self = [super init];
+  if (self != nil)
+    {
+      _name = nil;
+      ASSIGN(_item, item);
+      _hasAmbiguousLayout = NO;
+      _constraintsAffectingLayout = [[NSMutableArray alloc] init];
+      _attribute = attribute;
+    }
+  return self;
+}
+
+- (NSLayoutAttribute) attribute
+{
+  return _attribute;
+}
 
 - (NSLayoutConstraint *) constraintEqualToAnchor: (NSLayoutAnchor *)anchor
 {
   return [NSLayoutConstraint constraintWithItem: _item
-                                      attribute: NSLayoutAttributeLeft
+                                      attribute: _attribute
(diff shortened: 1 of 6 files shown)


Source/NSLayoutAnchor.m
```diff
@@ -28,71 +28,91 @@
 
 #import "AppKit/NSLayoutAnchor.h"
 #import "AppKit/NSLayoutConstraint.h"
+#import "GSAutoLayoutAnchorPrivate.h"
 
 @implementation NSLayoutAnchor
 
+- (instancetype) initWithItem: (id)item attribute: (NSLayoutAttribute)attribute
```
fredkiefer, line 35: All Objective C classes should have exactly one dedicated initializer. So either this method has to call -init, or -init should call this one (more likely). And all the default values get set there.

fredkiefer (changes requested):
I really like this change, but we should try to be strict about initializers.

DTW-Thalion:
Perfectly fair - I'll loop through and clean these up.

Source/NSLayoutAnchor.m
DTW-Thalion: Done - -init now funnels through the designated initialiser -initWithItem:attribute:, so all the defaults are set in one place.

fredkiefer (approved):
Looks great now!

PR #553, opened 2026-07-16
Fix: vend layout anchors from NSLayoutGuide

DTW-Thalion:
Prerequisite: #550 

NSLayoutGuide's anchor accessors (`leadingAnchor`, `widthAnchor`, `topAnchor` and the rest) returned nil because the anchors were never created. This vends an anchor bound to the guide and the matching attribute for each accessor, so `guide.widthAnchor constraintEqualToConstant:` and anchor-to-anchor constraints build against the guide.

Verified against AppKit on macOS: `guide.widthAnchor constraintEqualToConstant:10` yields a constraint with the guide as firstItem, firstAttribute Width and no second item; `guide.leadingAnchor constraintEqualToAnchor:other.leadingAnchor` carries both guides and the Leading attribute. Includes a test.

**Depends on #550** — it uses the `initWithItem:attribute:` anchor infrastructure added there, so this branch is stacked on it and the diff currently shows #550's commits too. Merge #550 first; this then reduces to just the NSLayoutGuide change.

### DIFF
--- a/Source/NSLayoutGuide.m
+++ b/Source/NSLayoutGuide.m
@@ -23,6 +23,8 @@
 */
 
 #import "AppKit/NSLayoutGuide.h"
+#import "AppKit/NSLayoutAnchor.h"
+#import "GSAutoLayoutAnchorPrivate.h"
 
 @implementation NSLayoutGuide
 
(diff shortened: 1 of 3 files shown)


fredkiefer:
Could you please rebase here?

Source/NSLayoutGuide.m
```diff
@@ -53,52 +55,72 @@ - (void) setIdentifier: (NSUserInterfaceItemIdentifier)identifier
 
 - (NSLayoutXAxisAnchor *) leadingAnchor
 {
-  return _leadingAnchor;
```
fredkiefer, line 56: Do we need these instance variables at all? You removed the only usages.

DTW-Thalion:
Rebased onto current master.

Source/NSLayoutGuide.m
DTW-Thalion: Done. Removed the ten anchor ivars; the accessors vend fresh anchors and nothing else referenced them. Rebased onto master.

fredkiefer (approved):

PR #563, opened 2026-07-16
Fix: NSBrowserCell -copyWithZone: retains the copy's alternate image

DTW-Thalion:
`-[NSBrowserCell copyWithZone:]` retained the **receiver's** `_alternateImage` (`_alternateImage = TEST_RETAIN(_alternateImage)`) rather than the copy's, so the new cell never took its own reference to the alternate image and the receiver ended up retained one extra time. This retains `c->_alternateImage` instead.

Transparency note: because the superclass copy uses `NSCopyObject` (which already copies the `_alternateImage` pointer into the copy), the two extra/missing retains happen to cancel out for the common create-copy-release sequence, so this is not observable as a crash or leak today. The change corrects the assignment so the copy owns its reference regardless of how the superclass copy is implemented. The added test guards that a copy keeps its leaf/loaded flags and alternate image.

### DIFF
--- a/Source/NSBrowserCell.m
+++ b/Source/NSBrowserCell.m
@@ -146,9 +146,7 @@ static NSFont *_leafFont;
 {
   NSBrowserCell	*c = [super copyWithZone: zone];
 
-  _alternateImage = TEST_RETAIN (_alternateImage);
-  //c->_browsercell_is_leaf = _browsercell_is_leaf;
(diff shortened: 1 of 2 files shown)


fredkiefer (approved):
You are correct, this doesn't change anything but it makes the code more explicit. The old code looked like a bug, although it worked correctly. This change should save some time in the future when people (or AI) try to understand the code.

PR #567, opened 2026-07-16
Fix: NSToolbarItem -copyWithZone: copies into the receiver's class

DTW-Thalion:
-copyWithZone: allocated an NSToolbarItem explicitly instead of the receiver's
own class, so copying a subclass returned a plain toolbar item.

The visible effect is that copying an NSToolbarItemGroup raises: the group's own
-copyWithZone: calls super and then sends -setSubitems: to the result, which a
plain NSToolbarItem does not implement.

    NSToolbarItemGroup *group;

    group = [[NSToolbarItemGroup alloc] initWithItemIdentifier: @"group"];
    [group setSubitems: [NSArray arrayWithObject: item]];
    [group copy];
    // NSInvalidArgumentException: NSToolbarItem(instance) does not
    // recognize setSubitems:

NSMenuToolbarItem does not override -copyWithZone:, so its copies came back as
plain items without raising, dropping the menu and the indicator flag.

AppKit copies both subclasses into their own class, checked on a macOS runner.

The test covers the group and the menu item, and keeps the plain item case as a
guard. Without the change 5 of its assertions fail; with it the NSToolbarItem
tests pass 26/26.

### DIFF
--- a/Source/NSToolbarItem.m
+++ b/Source/NSToolbarItem.m
@@ -1609,7 +1609,7 @@ NSString *GSMovableToolbarItemPboardType = @"GSMovableToolbarItemPboardType";
 // NSCopying protocol
 - (id) copyWithZone: (NSZone *)zone 
 {
-  NSToolbarItem *new = [[NSToolbarItem allocWithZone: zone] 
+  NSToolbarItem *new = [[[self class] allocWithZone: zone]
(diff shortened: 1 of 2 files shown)


fredkiefer (approved):
Good catch!

PR #568, opened 2026-07-16
Fix: NSToolbarItemGroup -subitems returns an empty array

DTW-Thalion:
-subitems returned nil for a group that has no subitems. AppKit returns an
empty array, checked on a macOS runner, both for a new group and after the
subitems are set back to nil.

This follows the same shape as the -pathItems and -pathComponentCells getters.

Without the change 2 of the test's assertions fail; with it the
NSToolbarItemGroup tests pass 5/5.

### DIFF
--- a/Source/NSToolbarItemGroup.m
+++ b/Source/NSToolbarItemGroup.m
@@ -40,6 +40,10 @@
 
 - (NSArray *) subitems
 {
+  if (_subitems == nil)
+    {
+      return [NSArray array];
+    }
   return _subitems;
 }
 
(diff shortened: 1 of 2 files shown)


Source/NSToolbarItemGroup.m
```diff
@@ -40,6 +40,10 @@ - (void) setSubitems: (NSArray *)items
 
 - (NSArray *) subitems
 {
+  if (_subitems == nil)
```
fredkiefer, line 43: Again I would prefer to have the array assigned in the -init method. On the other hand, we should most likely return a copy of the array here, so your code may be closer to MacOS.

fredkiefer (approved):

PR #569, opened 2026-07-16
Fix: a new NSMenuToolbarItem shows its indicator

DTW-Thalion:
showsIndicator was NO for a new menu toolbar item. AppKit has it YES, checked
on a macOS runner.

The initialiser already installed the NSMenuToolbarItem indicator image, so the
flag disagreed with the image it is meant to track: -showsIndicator answered NO
while the indicator was on screen, and -setShowsIndicator: NO then had nothing
to remove.

-setShowsIndicator: YES installs that same image, so the initialiser can set the
flag through it and keep the two in step.

Without the change 1 of the test's assertions fails; with it the
NSMenuToolbarItem tests pass 4/4.

### DIFF
--- a/Source/NSMenuToolbarItem.m
+++ b/Source/NSMenuToolbarItem.m
@@ -35,7 +35,7 @@
   self = [super initWithItemIdentifier: identifier];
   if (self != nil)
     {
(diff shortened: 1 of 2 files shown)


fredkiefer (approved):
With these two values in sync we could think about further simplification of the code.

PR #570, opened 2026-07-16
Fix: NSTextAlternatives copies the strings it is given

DTW-Thalion:
The primaryString and alternativeStrings properties are declared copy, but the
initialiser retained its arguments, so mutating the caller's objects afterwards
changed what the object reported:

    NSMutableString *s = [NSMutableString stringWithString: @"abc"];
    NSTextAlternatives *alt;

    alt = [[NSTextAlternatives alloc] initWithPrimaryString: s
                                        alternativeStrings: a];
    [s appendString: @"DEF"];
    [alt primaryString];   // abcDEF, AppKit reports abc

The getters returned [ivar copy]. Neither name implies ownership, so the +1
object they handed back leaked on every call. With the initialiser taking the
copy the getters can return their storage, which also matches AppKit returning
the same object on every call.

Both were checked on a macOS runner: AppKit snapshots the strings at init, its
getters are stable across calls, and for an immutable argument the stored object
is the argument itself, since copying it returns self.

Without the change 4 of the test's assertions fail; with it the
NSTextAlternatives tests pass 8/8. The separate coverage tests stay green either
way.

### DIFF
--- a/Source/NSTextAlternatives.m
+++ b/Source/NSTextAlternatives.m
@@ -39,8 +39,8 @@ NSString *NSTextAlternativesSelectedAlternativeStringNotification =
 {
   if ((self = [super init]))
     {
-      _primaryString = RETAIN(primaryString);
-      _alternativeStrings = RETAIN(alternativeStrings);
+      _primaryString = [primaryString copy];
+      _alternativeStrings = [alternativeStrings copy];
     }
 
   return self;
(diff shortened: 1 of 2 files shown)


Source/NSTextAlternatives.m
```diff
@@ -39,21 +39,21 @@ - (id)initWithPrimaryString:(NSString *)primaryString
 {
   if ((self = [super init]))
     {
-      _primaryString = RETAIN(primaryString);
-      _alternativeStrings = RETAIN(alternativeStrings);
+      _primaryString = [primaryString copy];
```
fredkiefer, line 42: We prefer to use the macro ASSIGNCOPY.

fredkiefer (approved):

PR #572, opened 2026-07-16
Fix: implement the NSToolbarItemGroup selection API

DTW-Thalion:
NSToolbarItemGroup carried a subitems array and nothing else, so an application
could group items but not track which of them is chosen. This adds the selection
API: the two enumerations, the selection mode, the control representation, the
selected index and the per-index selection.

The behaviour follows AppKit, checked on a macOS runner:

- the enumeration values: SelectOne 0, SelectAny 1, Momentary 2, and Automatic
  0, Expanded 1, Collapsed 2
- a new group selects one subitem, has the automatic control representation and
  a selected index of -1
- selecting one: selecting a subitem deselects the previous one, and
  deselecting is ignored, so the group is never left with nothing selected
- selecting any: subitems select and deselect independently, and selectedIndex
  is the one selected last
- momentary: -setSelected:atIndex: holds no selection, while -setSelectedIndex:
  records the index without selecting anything
- -setSelectedIndex: -1 leaves the selection alone; an index out of range raises
  NSRangeException
- changing the selection mode leaves an existing selection alone
- replacing the subitems clears the selection

Two things worth flagging.

AppKit only tracks selection on groups built by its convenience constructors;
on a group built with -initWithItemIdentifier: and -setSubitems: the same calls
do nothing at all. Those constructors exist to build the segmented control that
represents the group, which is also where the titles they take end up. That
representation does not exist here, so the constructors are not part of this
change and the selection is tracked for any group with subitems rather than
reproducing AppKit's silence.

The toolbar does not lay out group subitems yet, so this is state that the
group keeps rather than something the user can see. It is the model half of the
class, and a renderer needs it either way.

26 tests, all passing.

### DIFF
--- a/Headers/AppKit/NSToolbarItemGroup.h
+++ b/Headers/AppKit/NSToolbarItemGroup.h
@@ -36,15 +36,51 @@
 #if OS_API_VERSION(MAC_OS_X_VERSION_10_5, GS_API_LATEST)
 
 @class NSArray;
+@class NSMutableIndexSet;
+
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_15, GS_API_LATEST)
+enum {
+  NSToolbarItemGroupSelectionModeSelectOne = 0,
+  NSToolbarItemGroupSelectionModeSelectAny = 1,
+  NSToolbarItemGroupSelectionModeMomentary = 2
+};
+typedef NSInteger NSToolbarItemGroupSelectionMode;
+
+enum {
+  NSToolbarItemGroupControlRepresentationAutomatic = 0,
+  NSToolbarItemGroupControlRepresentationExpanded = 1,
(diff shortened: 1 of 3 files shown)


Headers/AppKit/NSToolbarItemGroup.h
```diff
@@ -36,16 +36,52 @@
...
+typedef NSInteger NSToolbarItemGroupControlRepresentation;
+#endif
 
 APPKIT_EXPORT_CLASS
 @interface NSToolbarItemGroup : NSToolbarItem
 {
 	NSArray *_subitems;
+	NSMutableIndexSet *_selectedIndexes;
+	NSInteger _selectionMode;
+	NSInteger _controlRepresentation;
```
fredkiefer, line 63: Why not `NSToolbarItemGroupControlRepresentation`?

fredkiefer (approved):

PR #586, opened 2026-07-16
Fix: NSDockTile lets go of its ivars in dealloc

DTW-Thalion:
NSDockTile lets go of its content view, badge label and images in -release
rather than in -dealloc, so it does so on every release, not when it is
deallocated. A tile that is retained and released again, which is all it takes,
is left holding four dangling pointers while still alive.

The second such cycle then over-releases them. With the test below on an
unmodified library:

    Failed test: a balancing release does not let go of the content view
    *** -[NSImage release]: message sent to deallocated instance 0x...
    timeout: the monitored command dumped core

-dealloc is where this belongs: the ivars are let go of once, when the tile
goes.

The test retains and releases a tile and checks it still holds its content view
and badge label. Without the change the first assertion fails and the program
dumps core partway through; with it the NSDockTile tests pass 3/3, and the
coverage tests in #585 pass either way.

### DIFF
--- a/Source/NSDockTile.m
+++ b/Source/NSDockTile.m
@@ -69,13 +69,13 @@
   return self;
 }
 
-- (oneway void) release
+- (void) dealloc
 {
   RELEASE(_contentView);
   RELEASE(_badgeLabel);
   RELEASE(_appIconImage);
   RELEASE(_dockTileImage);
-  [super release];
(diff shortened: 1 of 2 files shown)


fredkiefer (approved):
Thank you for fixing this

PR #590, opened 2026-07-16
Fix: NSTokenFieldCell archives its token style, delay and character set

DTW-Thalion:
Both branches of -encodeWithCoder: and -initWithCoder: were empty, so a token
field cell wrote none of its three ivars and came back from an archive with the
token style, the completion delay and the tokenizing character set all lost.

Archiving a cell on a macOS runner shows the keys AppKit uses, which are the
ones used here:

    NSTokenStyle = 2;
    NSCompletionDelay = "2.5";
    NSTokenizingCharacterSet = <the character set>;

The keyed branch guards each key with -containsValueForKey:, so an archive
written before this change decodes as it did.

The old style branch has no keys to guard with, and the class version was
already 1 when nothing was being written, so there is no telling an old empty
archive from a new one at that version. The version therefore goes to 2 and the
decoder only reads the three values at version 2 or later, the way
NSTextFieldCell and NSButtonCell handle their own format changes.

Without the change 6 of the test's assertions fail; with it the
NSTokenFieldCell tests pass 8/8.

### DIFF
--- a/Source/NSTokenFieldCell.m
+++ b/Source/NSTokenFieldCell.m
@@ -39,7 +39,7 @@
 {
   if (self == [NSTokenFieldCell class])
     {
-      [self setVersion: 1];
+      [self setVersion: 2];
     }
 }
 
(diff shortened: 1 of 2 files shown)


fredkiefer (approved):
You might want to have a look at the file ObjectTesting.h in GNUstep make. This has a few macros for simple test. These may save you some time and effort. But be warned, these only work properly when there is a suitable -isEqual: method on the class.

PR #599, opened 2026-07-16
tests: add NSNibConnector tests

DTW-Thalion:
NSNibConnector has no tests. This adds coverage for the source, destination and
label it holds, replacing an object it holds, how it compares itself to another
connector, establishing a connection, and archiving.

Unlike the other tests I have been sending, these are not checked against
AppKit. AppKit still has an NSNibConnector in its binary, but it is not
declared in any header any more, so there is nothing to write portable code
against and no published behaviour to match. What it does have differs anyway:
its -isEqual: is plain identity, where this one compares the objects joined,
the label and the class. That comparison looks deliberate here, and the nib,
gorm and xib loaders and the two subclasses are what use it, so these tests
pin what this class does rather than what AppKit's unpublished one does.

22 tests, all passing:

- a new connector has no source, destination or label
- all three read back, and the source and label can be cleared
- replaceObject:withObject: replaces the source and the destination when it
  holds the object, and changes nothing when it does not
- a connector equals itself; two connectors joining the same objects with the
  same label are equal; a different label, a different class, or something that
  is not a connector at all are not
- establishing a connection with nothing to join does not raise
- a connector keeps its source, destination and label through a keyed archive

### DIFF
--- a/Tests/gui/NSNibConnector/basic.m
+++ b/Tests/gui/NSNibConnector/basic.m
@@ -1,0 +1,138 @@
+/* Coverage for NSNibConnector: the source, destination and label it holds,
+ * replacing an object it holds, how it compares itself to another connector,
+ * and archiving.  These are plain model objects and need no backend.
+ */
+#include "Testing.h"
+#include <Foundation/Foundation.h>
+#include <AppKit/NSNibConnector.h>
+#include <AppKit/NSNibControlConnector.h>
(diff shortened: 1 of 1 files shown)


fredkiefer (approved):
As far as I am aware the compare behaviour is not used in the GNUstep code at the moment. So if you prefer the MacOS behaviour I am open to change that.

PR #601, opened 2026-07-19
tests: add NSView geometry and rendering tests

DTW-Thalion:
Adds tests for NSView.

geometry.m checks -convertPoint:toView: between a view and a subview: a point in
the subview's coordinates is offset by the subview's origin, and the conversion
round-trips. It also checks that a plain NSView is not flipped. The views are
built without a window, so the test needs no display.

rendering.m checks that -drawRect: reaches the drawing surface: it fills a view
with red, captures the focused view rect into an NSBitmapImageRep, and reads the
centre pixel back as red. It skips when no display is available.

Both use the existing backend skip guard, so they skip when no backend is
installed and run otherwise.

### DIFF
--- a/Tests/gui/NSView/geometry.m
+++ b/Tests/gui/NSView/geometry.m
@@ -1,0 +1,40 @@
+#import "Testing.h"
+#import <Foundation/NSGeometry.h>
+#import <AppKit/NSApplication.h>
+#import <AppKit/NSView.h>
+
+int main(int argc, const char **argv)
+{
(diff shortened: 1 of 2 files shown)


DTW-Thalion:
@fredkiefer -this PR #601 and Issue #602 are needed before diving into the heavy GUI classes (NSView, NSWindow, etc.) for testing and bug fixes. There are 130+ classes that will have new testing harnesses and these two are the start of layingf that groundwork.

fredkiefer (approved):

PR #603, opened 2026-07-19
tests: rework NSView geometry tests

DTW-Thalion:
The NSView geometry tests were scattered across a dozen small files with overlapping and inconsistent coverage. This consolidates them into five topic suites and expands each one, checking new cases against AppKit. rendering.m is unchanged.

File mapping:
- convert.m <- NSView_convertRect.m + geometry.m
- frameBounds.m <- NSView_frame_bounds.m + NSView_bounds_scale.m + NSView_setFrameSize_zero.m
- rotation.m <- NSView_frame_rotation.m
- autoresize.m <- NSView_autoresize_and_rounding.m
- visibleClip.m <- NSView_visibleRect.m + scrollRectToVisible.m

Every assertion from the removed files is carried over. The suites keep the existing sharedApplication/backend skip guard, so they skip cleanly where no backend is installed and run normally against one.


fredkiefer (approved):
I am really looking forward to see the cases where the behaviour didn't match. We had this tests working on a Mac once, but that was years ago.

PR #607, opened 2026-07-19
ci: run the gui tests headless against a cairo backend

### LINKED ISSUE #602: Run the gui tests against a backend in CI
The gui CI installs no backend. main.yml has the libs-back build step commented
out ("disabled due to various test failures"), so on every leg
[NSApplication sharedApplication] raises and the tests guard themselves out with
SKIP. The tests that need a backend never run in CI: NSView geometry, window
behaviour, toolbar, popup, save panel, and the rest.

I have a branch that adds one job which installs a cairo backend and runs the suite. 
On the GitHub Linux runner it reports 2481 passed and 0 skipped; before the change 
the same tests skip.

The job mirrors the existing clang gnustep-2.0 leg. It builds and installs
gnustep/libs-back with --enable-graphics=cairo. libs-back builds an x11 server
layer even for the cairo graphics backend, so the X development libraries (libxt,
libxmu, libxft, libxrandr, libxfixes, libxcursor) are needed at build time. It
then runs make check under xvfb-run: cairo draws off screen, but the tests that
create a window and ask the window server for the screen list need a display, so
without one they abort with NSWindowServerCommunicationException. A virtual
framebuffer is enough; no real display is required.

Turning the tests on exposes four pre-existing failures that were hidden while
the suite skipped, unrelated to the CI change: NSDataLink/basic.m raises "nil
data passed to initForReadingWithData:", NSSavePanel/basic.m expects a non-nil
default allowedTypes, and NSFormCell/title.m and NSTextFieldCell/attributes.m
abort before completing. So the job is red today only on those four.

The CI job could go in on its own, with separate issues or fixes for the four,
or wait until they are addressed so it lands green. If you would rather it not
gate merges until then, it can start non-blocking. Is this something you want in
the gui CI, and which of those would you prefer?

cc @fredkiefer

DTW-Thalion:
Closes #602 

This adds a Linux CI job that installs a cairo backend (gnustep/libs-back) and
runs the gui tests under xvfb, so the tests that need a backend run instead of
skipping. It addresses #602. On the runner it reports 2481 passed and 0 skipped
where the same tests skip today.

The job exposes four pre-existing failures, fixed in #604, #605 and #606. Once
those merge the job runs green; it needs a rebase onto master after they land.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -141,9 +141,9 @@ jobs:
             ${{ env.SRC_PATH }}/config.log
             ${{ env.SRC_PATH }}/Tests/tests.log
 
-  ########### Linux (headless, cairo backend) ###########
+  ########### Linux (cairo backend, virtual display) ###########
   linux-headless:
-    name: Ubuntu x64 Clang gnustep-2.0 (headless, cairo backend)
+    name: Ubuntu x64 Clang gnustep-2.0 (cairo backend, Xvfb)
     runs-on: ubuntu-latest
     # don't run pull requests from local branches twice
     if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name != github.repository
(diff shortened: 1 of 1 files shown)


DTW-Thalion:
@fredkiefer #607 relies on #604, #605 and #606 to work - so in terms of priorities these would be helpful. :)

fredkiefer:
The others are merged feel free to rebase this.

fredkiefer (approved):

fredkiefer:
Oops, I merged the windows cascade patch and now these tests are failing. Most likely you already have a fix for that, but I still have to catch up with all the work you did during the night (in my timezone).

DTW-Thalion:
The cairo lane here is red only on Tests/gui/NSWindow/cascade.m. The cascade offset came from +frameRectForContentRect:, which is zero under server-side decorations, so the window did not cascade in this headless run. #636 fixes that by using the theme title bar height. Once it lands this lane should be green after a rebase.

cc @fredkiefer

fredkiefer (approved):

PR #608, opened 2026-07-19
Fix: NSView convertRect:toView: ignores the view hierarchy without a window

DTW-Thalion:
-[NSView convertRect:toView:] and convertRect:fromView: returned the rect
unchanged when the view had no window, while convertPoint: applies the hierarchy
transform in the same case. AppKit transforms the rect. This removes the
no-window short-circuit so convertRect matches convertPoint and AppKit.
gui/NSView/convertRectNoWindow.m fails before and passes after.


fredkiefer (approved):
Thank you, I must have forgotten to remove these checks, when we switched the implementation.

PR #609, opened 2026-07-19
Fix: NSView centerScanRect: round origin and size independently

DTW-Thalion:
-[NSView centerScanRect:] applied the width and height corrections to the wrong
components, so the returned rect had its width and height swapped relative to
AppKit. This swaps them back. gui/NSView/centerScanRectSize.m fails before and
passes after.

### DIFF
--- a/Source/NSView.m
+++ b/Source/NSView.m
@@ -1706,8 +1706,8 @@ static NSSize _computeScale(NSSize fs, NSSize bs)
   y_org = aRect.origin.y;
   aRect.origin.x = GSRoundTowardsInfinity(aRect.origin.x);
   aRect.origin.y = [self isFlipped] ? GSRoundTowardsNegativeInfinity(aRect.origin.y) : GSRoundTowardsInfinity(aRect.origin.y);
-  aRect.size.width = GSRoundTowardsInfinity(aRect.size.width + (x_org - aRect.origin.x) / 2.0);
-  aRect.size.height = GSRoundTowardsInfinity(aRect.size.height + (y_org - aRect.origin.y) / 2.0);
+  aRect.size.width = GSRoundTowardsInfinity(aRect.size.width);
+  aRect.size.height = GSRoundTowardsInfinity(aRect.size.height);
 
   matrix = [self _matrixFromWindow];
   aRect.origin = [matrix transformPoint: aRect.origin];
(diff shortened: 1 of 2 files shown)


Source/NSView.m
```diff
@@ -1706,8 +1706,8 @@ - (NSRect) centerScanRect: (NSRect)aRect
   y_org = aRect.origin.y;
   aRect.origin.x = GSRoundTowardsInfinity(aRect.origin.x);
   aRect.origin.y = [self isFlipped] ? GSRoundTowardsNegativeInfinity(aRect.origin.y) : GSRoundTowardsInfinity(aRect.origin.y);
-  aRect.size.width = GSRoundTowardsInfinity(aRect.size.width + (x_org - aRect.origin.x) / 2.0);
-  aRect.size.height = GSRoundTowardsInfinity(aRect.size.height + (y_org - aRect.origin.y) / 2.0);
+  aRect.size.width = GSRoundTowardsInfinity(aRect.size.width + (y_org - aRect.origin.y) / 2.0);
```
fredkiefer: Combining values for both coordinate dimensions in one equation looks still wrong to me. It may yield a plausible value for your test case, but I will need some more explanation here.
I looked at the commit where I introduced this computation and it still looks correct to me.  We first do some rounding for the origin and then when we do the rounding for the width/height we take half the rounding error of the same dimension into account. If Apple is doing it differently they will need good reasons for that.

DTW-Thalion: You are right - need to fix this, though Apple is doing it differently - Apple AppKit rounds origin and size independently.

DTW-Thalion: You were right that the cross-axis form was wrong, so I have replaced it. Against AppKit, -centerScanRect: rounds the origin and the size independently: (0.4,0.9,5.3,5.3) gives {{0,1},{5,5}} and (10.3,10.7,20.4,20.6) gives {{10,11},{20,21}}. The current code couples the size rounding to the origin error to keep the far edge aligned, which is a deliberate difference from that. The new diff drops the coupling so the size rounds on its own. If you would rather keep the edge alignment, I will close this instead.

fredkiefer (approved):
This is sad, our old code was slightly better as it tried to preserve more of the original frame. But we really have to follow Apple here.

PR #611, opened 2026-07-19
Fix: NSView autoresize rounds on the device pixel grid and drops the flipped Y-margin swap

DTW-Thalion:
Two autoresize fixes in -[NSView resizeWithOldSuperviewSize:], both checked
against AppKit.

First, each resized subview frame was rounded with centerScanRect: (nearest
device pixel). AppKit floors the min and the max edge of the frame per axis and
takes the size as the difference, so a subview is never rounded up by a point.
This replaces the centerScanRect: call with that flooring.

Second, for a flipped superview the method swapped which of NSViewMinYMargin and
NSViewMaxYMargin was treated as the flexible margin. AppKit does not reinterpret
the mask by the superview's flip state, so the swap made a flipped superview
resize its subviews differently from an unflipped one. This removes the swap.

gui/NSView/autoresizeRounding.m checks the flooring across both axes on growth
and shrink, including a flipped superview; gui/NSView/flipMargin.m checks that
autoresize is unchanged by a flipped superview.

### DIFF
--- a/Source/NSView.m
+++ b/Source/NSView.m
@@ -2083,15 +2083,13 @@ static void autoresize(CGFloat oldContainerSize,
 	     (_autoresizingMask & NSViewMaxXMargin));
 
   {
-    const BOOL flipped = (_super_view && [_super_view isFlipped]);
-
     autoresize(oldSize.height,
 	       superViewFrameSize.height,
 	       &newFrame.origin.y,
 	       &newFrame.size.height,
-	       flipped ? (_autoresizingMask & NSViewMaxYMargin) : (_autoresizingMask & NSViewMinYMargin),
+	       (_autoresizingMask & NSViewMinYMargin),
 	       (_autoresizingMask & NSViewHeightSizable),
-	       flipped ? (_autoresizingMask & NSViewMinYMargin) : (_autoresizingMask & NSViewMaxYMargin));
+	       (_autoresizingMask & NSViewMaxYMargin));
   }
 
   newFrameRounded = newFrame;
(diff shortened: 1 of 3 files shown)


Source/NSView.m
```diff
@@ -2101,7 +2099,15 @@ - (void) resizeWithOldSuperviewSize: (NSSize)oldSize
    */
   if (![self isRotatedFromBase] && [self superview] != nil)
     {
-      newFrameRounded = [[self superview] centerScanRect: newFrameRounded];
+      CGFloat minX = floor(newFrameRounded.origin.x);
```
fredkiefer: i am not fully convinced that this part is correct. I would need to see a test case with real scaling, where the pixels on the screen don't correspond to integers in the user coordinate space.

DTW-Thalion: Added a scaling case: a subview autoresized inside a 2x-scaled parent. Against AppKit its frame comes back {{10,16.5},{20,33.5}}, i.e. the edges floor on the window pixel grid, not on the superview integer grid. The earlier version floored in the superview own space, which is only correct at 1x, so I reworked it to convert to the window base, floor there, and convert back. A small tolerance absorbs the transform rounding noise so an exact edge does not lose a pixel. The flipped-superview margin change is unchanged.

fredkiefer: The question now is, whether your fixed -centerScanRect: doesn't behave the same? In that case I would prefer to have just one place where we do this computation.

DTW-Thalion: They do not behave the same. -centerScanRect: rounds each edge to the nearest device pixel, but AppKit floors each edge when it integralizes an autoresized frame (for example a raw origin of 3.75 becomes 3, not 4, and the far edge is floored independently). Calling -centerScanRect: here would round instead of floor and give the wrong frame, so the rounding step has to differ. The only shared part is the transform to the window base and back. I can factor that into a small helper if you prefer, but the two callers still need different edge rounding.

fredkiefer (approved):

PR #616, opened 2026-07-19
Fix: NSWindow stores the aspect ratio set with setAspectRatio:

DTW-Thalion:
-[NSWindow aspectRatio] and -setAspectRatio: were stubs: the setter did nothing
and the getter always returned (1,1). This adds an _aspectRatio ivar and stores
the value, so the ratio set with setAspectRatio: is returned by aspectRatio.

This makes the accessor round-trip; constraining a live resize to the ratio is
not part of this change. The ivar is added at the end of NSWindow's existing
size ivars.

### DIFF
--- a/Source/NSWindow.m
+++ b/Source/NSWindow.m
@@ -2461,6 +2461,7 @@ titleWithRepresentedFilename(NSString *representedFilename)
 - (void) setResizeIncrements: (NSSize)aSize
 {
   _increments = aSize;
+  _aspectRatio = NSZeroSize;
   if (_windowNum > 0)
     {
       [GSServerForWindow(self) setresizeincrements: aSize : _windowNum];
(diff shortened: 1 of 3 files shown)


Source/NSWindow.m
```diff
@@ -2469,13 +2469,12 @@ - (void) setResizeIncrements: (NSSize)aSize
...
 {
-  // FIXME: This method is missing
-  return NSMakeSize(1, 1);
+  return _aspectRatio;
 }
 
 - (void) setAspectRatio: (NSSize)ratio
 {
-  // FIXME: This method is missing
+  _aspectRatio = ratio;
```
fredkiefer, line 2478: This method should unset the `resizeIncrements`  and vice versa.

DTW-Thalion: Yes, my mistake - operator error. :)

DTW-Thalion: Done. setAspectRatio: now clears the resize increments and setResizeIncrements: clears the aspect ratio; extended the test to cover both directions.

DTW-Thalion: Done. setAspectRatio: now clears the resize increments and setResizeIncrements: clears the aspect ratio; extended the test to cover both directions.

fredkiefer: Thank you!

fredkiefer (approved):

PR #626, opened 2026-07-20
Fix: NSStackView initialises all its state in the designated initializer

DTW-Thalion:
A programmatically created NSStackView left distribution, spacing, detachesHiddenViews and alignment at their zero values, but AppKit defaults them to gravity-areas distribution, spacing 8, detachment on, and centre-Y alignment. Set them in a shared helper called from initWithFrame: and initWithViews: (initWithCoder: already decodes them). Adds a test.

### DIFF
--- a/Source/NSStackView.m
+++ b/Source/NSStackView.m
@@ -443,16 +443,19 @@
 
 // Instance methods
 // Manage views...
-- (instancetype) initWithViews: (NSArray *)views
+- (instancetype) initWithFrame: (NSRect)frameRect
 {
-  self = [super init];
+  self = [super initWithFrame: frameRect];
   if (self != nil)
     {
-      NSUInteger c = [views count];
-      
-      _arrangedSubviews = [[NSMutableArray alloc] initWithArray: views];
-      _detachedViews = [[NSMutableArray alloc] initWithCapacity: c];
-      ASSIGNCOPY(_views, views);
+      _distribution = NSStackViewDistributionGravityAreas;
+      _spacing = 8.0;
+      _detachesHiddenViews = YES;
+      _alignment = NSLayoutAttributeCenterY;
+
+      _arrangedSubviews = [[NSMutableArray alloc] init];
+      _detachedViews = [[NSMutableArray alloc] init];
+      _views = [[NSMutableArray alloc] init];
       _customSpacingMap = RETAIN([NSMapTable weakToWeakObjectsMapTable]);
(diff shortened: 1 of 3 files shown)


Source/NSStackView.m
```diff
@@ -443,11 +443,30 @@ - (NSArray *) detachedViews
...
 // Manage views...
+- (void) _setStackViewDefaults
+{
+  _distribution = NSStackViewDistributionGravityAreas;
+  _spacing = 8.0;
+  _detachesHiddenViews = YES;
+  _alignment = NSLayoutAttributeCenterY;
+}
+
+- (instancetype) initWithFrame: (NSRect)frameRect
```
fredkiefer, line 446: As I already explained in other PRs, we need to have a designated initializer, so one of these has to call the other. And if we do so, your extra method won't be needed any more.
Best test on MacOS, how the -initWithFrame: behaves.

DTW-Thalion: Yes, I missed this one... fixing

DTW-Thalion: Done. initWithFrame: is now the designated initializer with the defaults inline, initWithViews: calls it, and the helper method is removed.

fredkiefer: No, it is not. Still half the instance variables are set in either method. Please move these to either of them.

DTW-Thalion: Wierd - something didn't push properly- I will fix.

DTW-Thalion: Done. All ivars are now initialised in the designated -initWithFrame:; -initWithViews: chains to it and only adds the given views. Rebased onto master.

DTW-Thalion: Confirmed on the current push: every default ivar (_distribution, _spacing, _detachesHiddenViews, _alignment, the collections and both map tables, the gravity containers) is set in the designated -initWithFrame:; -initWithViews: just chains to it and adds the given views. Helper removed, rebased onto master.

DTW-Thalion:
Pushed one more commit on top: -initWithCoder: was the remaining init path that didn't set up the arranged/detached/view arrays or the custom-spacing and visibility-priority map tables, so a stack view unarchived from a nib had them all nil and -setCustomSpacing:afterView: / -setVisibilityPriority:forView: silently did nothing. It now creates them the same way -initWithFrame: does. Added a coding test that archives a stack view and checks custom spacing round-trips on the decoded copy (verified it fails without the initWithCoder change and passes with it).

fredkiefer (approved):
Sorry, I didn't notice this was fixed.

PR #629, opened 2026-07-20
tests: add NSTableView tests

DTW-Thalion:
Adds tests for NSTableView covering the scalar state defaults and round-trips, column management (add, identifier lookup, back-pointer, move, remove), data-source-driven row selection (multiple selection, isRowSelected:, selectedRow, deselect), and a render regression lock. Machine dependent geometry (rowHeight, intercellSpacing values and the derived rects) is not asserted. The assertions were checked against AppKit.

### DIFF
--- a/Tests/gui/NSTableView/rendering.m
+++ b/Tests/gui/NSTableView/rendering.m
@@ -1,0 +1,72 @@
+#import "Testing.h"
+#import <Foundation/NSAutoreleasePool.h>
+#import <Foundation/NSGeometry.h>
+#import <Foundation/NSString.h>
+#import <AppKit/NSApplication.h>
+#import <AppKit/NSWindow.h>
+#import <AppKit/NSTableView.h>
+#import <AppKit/NSTableColumn.h>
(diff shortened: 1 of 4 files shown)


Tests/gui/NSTableView/rendering.m
```diff
@@ -0,0 +1,72 @@
...
+                  styleMask: NSWindowStyleMaskBorderless
+                    backing: NSBackingStoreBuffered
+                      defer: NO]);
+      NSTableView *tv = AUTORELEASE([[NSTableView alloc]
+        initWithFrame: NSMakeRect(0, 0, 120, 80)]);
+      NSTableColumn *col = AUTORELEASE([[NSTableColumn alloc]
+        initWithIdentifier: @"a"]);
+      [col setWidth: 100.0];
+      [tv addTableColumn: col];
+      [tv setDataSource: AUTORELEASE([RTVDS new])];
```
fredkiefer, line 47: In GNUstep we prefer separate alloc/int calls. But I accept this for now.

fredkiefer (approved):

PR #636, opened 2026-07-20
Fix: NSWindow cascade offset uses the theme title bar height

DTW-Thalion:
cascadeTopLeftFromPoint: derived its offset from +frameRectForContentRect:styleMask:, which adds a title bar only when the backend draws client-side decorations. With server-side decorations that delta is zero, so the window did not cascade at all. Use the theme's titlebarHeight directly, which is the same value with client-side decorations and non-zero regardless of the decoration mode. This is what the headless test run in the cairo CI lane needs.


DTW-Thalion:
cc @fredkiefer - this fixes the PR #607 CI failure. Merge and rebase needed.

fredkiefer (approved):

PR #637, opened 2026-07-20
tests: add NSMatrix state, radio and rendering tests

DTW-Thalion:
Adds tests filling gaps in the existing NSMatrix coverage: the intercell spacing, background and selection-by-rect scalar defaults and round-trips, radio-mode single selection with selectedRow/selectedColumn and cell state, cell geometry from an explicit cell size, and a render regression lock. The assertions were checked against AppKit.

### DIFF
--- a/Tests/gui/NSMatrix/radio.m
+++ b/Tests/gui/NSMatrix/radio.m
@@ -37,20 +37,20 @@ int main()
          numberOfRows: 2
       numberOfColumns: 2]);
 
-      pass([r numberOfRows] == 2, "the matrix has two rows");
-      pass([r numberOfColumns] == 2, "the matrix has two columns");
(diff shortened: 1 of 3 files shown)


rfm (approved):
This all seems fine (and no tests fail on my system), apart from consistently incorrectly calling 'pass' rather than 'PASS'.   While calling the low level function works in terms of reporting test success/failure, it means that the test output is missing diagnostic information, and if exceptions are raised during tests they will crash the test framework.

DTW-Thalion:
Good catch - will fix.

DTW-Thalion:
Done. Converted to PASS here, and swept the same change across my other open test PRs.

fredkiefer (approved):

PR #638, opened 2026-07-20
Fix: NSMatrix defaults autosizesCells and tabKeyTraversesCells to NO

DTW-Thalion:
A new NSMatrix set autosizesCells and tabKeyTraversesCells to YES in its private frame setup, but AppKit defaults both to NO. Set them to NO to match. Adds a test. These are behavioural defaults, so please say if GNUstep prefers to keep the current values.

### DIFF
--- a/Source/NSForm.m
+++ b/Source/NSForm.m
@@ -73,6 +73,7 @@ static Class defaultCellClass = nil;
     return nil;
 
   [self setIntercellSpacing: NSMakeSize (0, 4)];
+  [self setTabKeyTraversesCells: YES];
   return self;
 }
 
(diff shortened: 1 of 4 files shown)


fredkiefer:
This change by itself seems correct. But I need to test how it changes the behaviour of GNUstep panels and other places where NSMatrix or a subclass gets used.

DTW-Thalion:
I went through the other users of NSMatrix and its subclasses in the tree.For autosizesCells, the only callers that depend on a value set it explicitly: GSThemePanel sets NO, and the NSSavePanel form and the NSColorPanel picker matrix set YES. On macOS, both a plain NSMatrix and an NSForm default autosizesCells to NO, so nothing there regresses.For tabKeyTraversesCells, nothing in the tree sets it, so only code relying on the default is affected. MacOS shows AppKit defaults a plain NSMatrix to NO but an NSForm to YES. NSForm was getting YES only from the old NSMatrix default, so I added a commit that sets tabKeyTraversesCells to YES in the NSForm initialisers, with a test for the NSForm defaults. That keeps form-field tab traversal working and matches AppKit for both classes.

fredkiefer (commented):
Thank you, with NSForm fixed I hope this is good enough to merge

PR #640, opened 2026-07-20
Fix: NSMatrix stores a tool tip per cell

### LINKED ISSUE #639: NSMatrix: per-cell tool tip methods are stubs
-toolTipForCell: returns an empty string and -setToolTip:forCell: is a no-op, so per-cell tool tips are not stored. Implementing them needs a per-cell storage map (a new ivar), so this is an ABI change rather than a small fix.

DTW-Thalion:
-toolTipForCell: returned an empty string and -setToolTip:forCell: was a no-op. Store the tool tips in a map table held in the previously unused _reserved1 ivar (weak cell keys), so -setToolTip:forCell: records the string, -toolTipForCell: returns it or nil when none is set, and a nil tool tip clears it. No new ivar. Closes #639.

### DIFF
--- a/Source/NSMatrix.m
+++ b/Source/NSMatrix.m
@@ -62,6 +62,7 @@
 #import <Foundation/NSKeyValueCoding.h>
 #import <Foundation/NSKeyValueObserving.h>
 #import <Foundation/NSNotification.h>
+#import <Foundation/NSMapTable.h>
 #import <Foundation/NSFormatter.h>
 #import <Foundation/NSDebug.h>
 #import <Foundation/NSString.h>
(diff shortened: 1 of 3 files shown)


Source/NSMatrix.m
```diff
@@ -373,6 +374,7 @@ - (void) dealloc
   [_cellPrototype release];
   [_backgroundColor release];
   [_cellBackgroundColor release];
+  [_reserved1 release];
```
fredkiefer: We should rename the ivar to something like `_tooltipMap`.

DTW-Thalion: Done, renamed to _tooltipMap.

fredkiefer (approved):

PR #643, opened 2026-07-20
Feature: Implement the Auto Layout constraint solver

### LINKED ISSUE #540: NSLayoutConstraint activation does not register with the layout engine (setActive:/activateConstraints: are no-ops)
`-[NSLayoutConstraint setActive: YES]` and `+[NSLayoutConstraint activateConstraints:]` don't take effect: `-isActive` stays `NO`, and the constraint isn't applied to layout. On macOS, `setActive: YES` makes `-isActive` return `YES` and registers the constraint.

`setActive:`/`isActive` are backed by a static `activeConstraints` array that is disconnected from Auto Layout: real constraints reach the solver through `-[NSView addConstraint:]` → `GSAutoLayoutEngine`, but `setActive:` never calls that. It only calls `_activateConstraint:`, whose one line (`[activeConstraints addObject: constraint]`) is additionally commented out, so even the `isActive` flag never flips:

```objc
+ (void) _activateConstraint: (NSLayoutConstraint *)constraint
{
  // [activeConstraints addObject: constraint];
}
```

A correct fix would route `setActive:`/`activateConstraints:` into the constraint's view and its layout engine (and leave factory-created constraints inactive, matching AppKit), rather than tracking a separate array — so raising it as an issue rather than a patch.

DTW-Thalion:
GSCassowarySolver was a stub, so activating layout constraints produced no solution and constrained views laid out to a zero rect. This implements the solver and wires up activation so the anchor and NSLayoutConstraint APIs actually position views.

Changes:

- GSCassowarySolver: implement the simplex orchestration on top of GSCSTableau. Adding and removing required and non-required constraints, the artificial-variable path, objective optimisation, edit variables (suggestEditVariable:equals:), and reading the external variable values back out as a GSCSSolution.
- GSCSVariable: -isRestricted classified external variables as restricted and dummy variables as unrestricted, the reverse of the classification the tableau relies on for infeasible-row detection and exit-variable selection.
- GSCSTableau: -substituteOutTerm:withExpression:inExpression:subject: folded in the terms of the expression being modified instead of the terms of the substituting expression.
- GSCSEditInfo: retain the objects it stores, expose the error variables and previous constant, and track the constant as a CGFloat so fractional values are preserved.
- GSAutoLayoutEngine: -alignmentRectForView: and its helpers returned NSZeroRect; return the solved minX, minY, width and height for a view.
- NSLayoutConstraint: -setActive:, +activateConstraints: and the anchor methods now install a constraint against the layout engine of the view that owns it (the nearest ancestor shared by its two items) and remove it on deactivation. Constraints activated before their view has a window are installed once it moves into one. A constraint created with a factory method or initialiser is inactive until activated, matching AppKit.
- NSWindow: when the layout engine is created, the content view is pinned to its own bounds, so subview constraints resolve to absolute positions without the application constraining the root view.
- GSAutoLayoutEngine / GSAutoLayoutVFLParser: the engine multiplied a relational constraint's constant by -1 for the right, top and trailing attributes, which inverted the constant for anchor constraints and for those attributes in a visual format string. The constant is now applied as written, and the visual format parser negates the spacing for vertical arrangements.
- Resizing a window reflows its constrained subviews: the engine keeps the content view's size constraints and refreshes them from the current bounds, and a content view lays its constrained subtree out again when its frame changes.
- NSControl reports an intrinsic content size (its cell size), and NSView initialises the content hugging and compression resistance priorities, so a control can be laid out from its content without an explicit size constraint. This also fixed a freed-expression use in GSCSConstraint's inequality factory and a duplicate edit constraint in the solver, both first reached once the intrinsic size constraints run.

The GSCSVariable, GSCSTableau and GSCSEditInfo errors were never reached while the solver was stubbed.

The added gui tests drive subviews through the anchor and visual-format APIs and check the solved frames; they skip when no display is available.

Closes #540.

### DIFF
--- a/Source/GSCassowarySolver.m
+++ b/Source/GSCassowarySolver.m
@@ -21,6 +21,7 @@
 */
 
 #import "GSCassowarySolver.h"
+#import "GSCSEditInfo.h"
 #import "GSCSFloatComparator.h"
 #import "GSCSLinearExpression.h"
 #import "GSCSStrength.h"
@@ -53,6 +54,17 @@
 - (GSCSVariable *) exitVariableForMarker: (GSCSVariable *)marker;
 
 - (void) optimize: (GSCSVariable *)objective;
+
+- (GSCSEditInfo *) editInfoForVariable: (GSCSVariable *)variable;
+
+- (GSCSEditInfo *) addEditVariable: (GSCSVariable *)variable
+                          strength: (GSCSStrength *)strength;
+
+- (void) deltaEditConstant: (CGFloat)delta
+              plusVariable: (GSCSVariable *)plusVariable
+             minusVariable: (GSCSVariable *)minusVariable;
+
+- (void) dualOptimize;
 
 - (void) setExternalVariables;
 
@@ -210,7 +222,25 @@
 
 - (void) suggestEditVariable: (GSCSVariable *)variable equals: (CGFloat)value
 {
-  // FIXME Suggest edit variable
+  GSCSEditInfo *editInfo = [self editInfoForVariable: variable];
+  if (editInfo == nil)
+    {
+      editInfo = [self addEditVariable: variable
+                              strength: [GSCSStrength strengthStrong]];
+    }
+
+  CGFloat delta = value - [editInfo previousConstant];
+  [editInfo setPreviousConstant: value];
+
+  // The edit constraint expression is (targetValue - variable), so the sign of
+  // the change is the reverse of the classic (variable - targetValue) form.
+  [self deltaEditConstant: -delta
+             plusVariable: [editInfo plusVariable]
+            minusVariable: [editInfo minusVariable]];
+
+  [self dualOptimize];
+  [self setExternalVariables];
+  [[_tableau infeasibleRows] removeAllObjects];
 }
 
 // Build the augmented expression for a new constraint. Basic variables are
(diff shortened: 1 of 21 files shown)


DTW-Thalion:
@fredkiefer - you can let this PR marinade for a while. This is a feature, not a bug fix per se - I noted a lot of the plumbing existed, but wasn't wired up and I don't think even tested, so I wired it up and fixed the problems I could find against MacOS AppKit related behaviors. So this is a welcome diversion from test harnesses, but needs some solid testing by others in actual environments.

fredkiefer:
I hope to find time to look into this over the weekend. It is too much to work on in the evening.

DTW-Thalion:
Sounds good - tackle #692 if you could

Source/GSCassowarySolver.m
```diff
@@ -1,58 +1,831 @@
...
+  [self optimize: [_tableau objective]];
+  [self setExternalVariables];
 }
 
-- (void) removeConstraint: (GSCSConstraint*)constraint
+- (void) addConstraints: (NSArray *)constraints
 {
-  // FIXME Remove constraint from solver
+  FOR_IN(GSCSConstraint *, constraint, constraints)
+    [self addConstraint: constraint];
```
fredkiefer: This is fine for now, but later we should split up -addConstrain: so that we only solve the resulting constraints once here and not for each of the added ones separately. The same is true for removal.

Source/NSView.m
```diff
@@ -385,6 +389,10 @@ - (void) _viewDidMoveToWindow
     {
       [vc _viewDidMoveToWindow];
     }
+  if ([self window] != nil)
+    {
+      [NSLayoutConstraint _installPendingConstraintsForView: self];
```
fredkiefer, line 394: Should we always call this or only when the window has a layout engine?l

fredkiefer (commented):
Very impressive code. My main concern is, whether there is any performance impact on applications not using this feature?
And as this is a real mayor contribution, I would like to leave it to @gcasa whether we merge that now.

DTW-Thalion:
I don't see it being particularly performance degrading - the code is actually reasonably light for the feature, and if it was a performance issue, I would be more inclined to look at the backend.

I will make one change to -setFrame: and -setFrameSize: that will save about 5 ns per call, but it is on a hot path - other than that nothing else I can see that is costly.

Also will change the -addConstraint: -removeConstraint: per your suggestion.

DTW-Thalion:
Measured on a driver that creates no constraints at all. 200000
-setFrameSize: calls on ordinary subviews cost about 1.2 ms more than master,
near 6 ns a call, all of it -_autoLayoutContentViewResized sending -window and
-contentView before it can return. Nothing else moved: 5000 views entering a
window, 200000 view creations, and content view resizing all measured the same
either way. Minima of 5 interleaved runs, with a control row to catch drift.

The resize path now tests a flag set when the first layout engine is created,
which puts it back at master's cost. The install call at NSView.m:394 cannot
take that guard. A constraint activated before its view has a window is
installed there, and adding it is what creates the engine, so testing for an
existing engine would drop it. That call measured free, since it walks an
array that is empty until a constraint is activated.

Constraints are now added and removed in one solver pass. The engine handed
them to the solver one at a time, so the split had to reach it as well. A group
now collects its solver constraints and reads the solution back once.
Tests/gui/NSLayoutConstraint/batch.m covers both paths against the same
constraints applied singly, and fails without the change.

NSView.h now declares the two removal methods, which were implemented but
missing from the header.

fredkiefer:
Sounds great, thank you. I review in the morning!

DTW-Thalion:
One note, +activateConstraints: still loops singular, so the batch path is only reached via [view addConstraints:]. Making the modern path batch means grouping constraints by container view - I did not do that, but can if you prefer.

fredkiefer (approved):
Now the code looks even better. Thank you very much!

PR #648, opened 2026-07-20
Fix: NSCollectionView -setSelectionIndexes: tolerates indexes past the loaded items

DTW-Thalion:
Setting a selection index beyond the number of loaded items raised NSRangeException: the loop that marks items selected read -[NSArray objectAtIndex:] for every selected index without checking it against the item count. AppKit keeps the selection index set regardless of how many items exist.

The loop now skips indexes past the loaded items; the selection index set is still stored.

Added Tests/gui/NSCollectionView/selection.m, checked against AppKit on a macOS runner; backend-guarded.

### DIFF
--- a/Source/NSCollectionView.m
+++ b/Source/NSCollectionView.m
@@ -665,7 +665,7 @@ static NSString *_placeholderItem = nil;
     {
       if (index >= [_items count])
 	{
-	  continue;
+	  break;
 	}
 
       id item = [_items objectAtIndex: index];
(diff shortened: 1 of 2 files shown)


Source/NSCollectionView.m
```diff
@@ -658,6 +658,11 @@ - (void) setSelectionIndexes: (NSIndexSet *)indexes
   while ((index = [_selectionIndexes indexGreaterThanIndex: index]) !=
 	 NSNotFound)
     {
+      if (index >= [_items count])
+	{
+	  continue;
```
fredkiefer: I would rather break here, as no further index will be smaller.

DTW-Thalion: Done. The indexes are visited in ascending order, so once one is past the loaded items the rest are too; it now breaks. Rebased onto master.

fredkiefer (approved):

PR #674, opened 2026-07-20
Fix: GSHorizontalTypesetter recognises more East Asian scripts for word wrapping

### LINKED ISSUE #406: GSHorizontalTypesetter improvements: Fix the CJK Character Range Check

### Fix the CJK Character Range Check
The current CJK range check uses `0x2ff0` to `0x9fff`, but this is incorrect. The proper unified CJK range is `0x4E00` to `0x9FFF` (CJK Unified Ideographs). The current range includes Kangxi Radicals (0x2F00-0x2FDF) and Ideographic Description Characters (0x2FF0-0x2FFF), but misses many CJK blocks and includes non-CJK areas.

```objc
// CURRENT (incorrect):
else if ((character > 0x2ff0) && (character < 0x9fff))

// PROPOSED (comprehensive CJK support):
else if ((character >= 0x4E00 && character <= 0x9FFF) ||      // CJK Unified Ideographs
         (character >= 0x3400 && character <= 0x4DBF) ||      // CJK Extension A
         (character >= 0xAC00 && character <= 0xD7AF) ||      // Hangul Syllables
         (character >= 0x3040 && character <= 0x309F) ||      // Hiragana
         (character >= 0x30A0 && character <= 0x30FF) ||      // Katakana
         (character >= 0xFF65 && character <= 0xFF9F))       // Halfwidth Katakana
```

I cross checked the ranges proposed by Kimi with https://jrgraphix.net/research/unicode_blocks.php and all the ranges seem to be correct. However, I don't know if Kimi missed some ranges.

DTW-Thalion:
Fixes #406.

The word-wrapping code treats each East Asian character as its own word, since these scripts do not separate words with spaces. The range it used, `character > 0x2ff0 && character < 0x9fff`, missed several blocks: Hangul Syllables, CJK Compatibility Ideographs and Halfwidth Katakana were never recognised, so runs in those scripts were treated as a single unbreakable word.

The check now covers the ideographic, kana and hangul blocks through a small helper. The new set is a superset of the old range, so no character that previously wrapped per glyph loses that behaviour; the change only adds the missing blocks.

On testing: the effect is masked in simple cases because the typesetter falls back to character wrapping when a single word does not fit a line, so a run in an unrecognised script still wraps, just not at the natural per-character boundaries. There is no layout test harness in the suite to assert break positions directly, so this change rests on the range correctness above. I am happy to add a test if a suitable harness lands.

### DIFF
--- a/Source/GSHorizontalTypesetter.m
+++ b/Source/GSHorizontalTypesetter.m
@@ -272,6 +272,18 @@ typedef struct GSHorizontalTypesetterGlyphCacheStruct GlyphCacheEntry;
 }
 
 
+/* East Asian scripts (CJK ideographs, kana and hangul) do not separate words
+   with spaces, so each such character is treated as its own word when wrapping.
+   The ranges cover the ideographic, kana and hangul Unicode blocks. */
+static inline BOOL
+GSIsEastAsianWrapCharacter(unichar c)
+{
+  return (c >= 0x2E80 && c <= 0x9FFF)   /* CJK radicals through CJK Unified Ideographs */
+      || (c >= 0xAC00 && c <= 0xD7AF)   /* Hangul Syllables */
+      || (c >= 0xF900 && c <= 0xFAFF)   /* CJK Compatibility Ideographs */
+      || (c >= 0xFF65 && c <= 0xFF9F);  /* Halfwidth Katakana */
+}
+
 /*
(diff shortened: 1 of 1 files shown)


rmottola:
This PR addresses the same as #406 , but character ranges are different.

fredkiefer (approved):

PR #676, opened 2026-07-20
Fix: NSApplication keeps a drawable application icon image

### LINKED ISSUE #101: bug: [NSApp setApplicationIconImage:] doesn't set the image
Environment: freebsd 13, windowmaker, gnustep-gui from git master, clang 11.0.1

[Repro project available](https://github.com/iamleeg/icon-drawing-bug). This project draws a red square into its dock icon in two ways:

1. if you launch with no args, it calls `[NSApp setApplicationIconImage:dockImage]`. This doesn't do anything, and the app's dock tile remains empty.
2. if you launch as `IconDrawing.app/IconDrawing -UseIconWindow YES`, it sets the image in an `NSImageView`, and adds that as a subview to `[[NSApp iconWindow] contentView]`. This does work correctly, and draws the red square in the dock icon.

I believe that the first method would be expected by Cocoa developers as it's documented as supported (so is `NSDockTile`, which GNUstep doesn't yet have).

DTW-Thalion:
`-[NSApplication setApplicationIconImage:]` copies the image, since it changes its name and size, and `-[NSImage copyWithZone:]` does not copy cached representations. An image built by drawing into it, for example with `-lockFocus`, has only a cached representation, so the copy ends up with no representations and the icon draws nothing. This is the case reported in #101.

When the copy has no representations, the setter now takes an independent bitmap snapshot of the original image, via its TIFF representation, and adds it, so the icon still draws. Images loaded from a file already have a bitmap representation and are unaffected.

Added Tests/gui/NSApplication/applicationIconImage.m: a red image is drawn, set as the application icon, read back, and its centre pixel is checked to still be red. The test fails without this change, since the icon ends up with no representations.

Fixes #101.

### DIFF
--- a/Source/NSApplication.m
+++ b/Source/NSApplication.m
@@ -2424,6 +2424,26 @@ image.</p><p>See Also: -applicationIconImage</p>
   // Use a copy as we change the name and size
   ASSIGNCOPY(_app_icon, anImage);
 
+  /* -[NSImage copyWithZone:] does not copy cached representations, so an image
+     that was only drawn into (e.g. with -lockFocus, whose sole representation
+     is a cached one) copies to an image with no representations and would draw
+     nothing. Take an independent bitmap snapshot of the original in that case
+     so the icon still draws. */
+  if ([[_app_icon representations] count] == 0)
+    {
+      NSData *tiff = [anImage TIFFRepresentation];
+
+      if (tiff != nil)
+        {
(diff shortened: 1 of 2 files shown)


fredkiefer (approved):

fredkiefer:
Another idea that should be tested is to move the copy via tiff into the -copyWithZone: method of NSImage. We should test how this behaves on MacOS.

PR #679, opened 2026-07-21
Fix: NSMenuItem -setSubmenu: overwrites the submenu's title

### LINKED ISSUE #218: NSMenu ignores initWithTitle
This works:
    NSMenuItem *fileMenuItem = [NSMenuItem new];
    NSMenu *fileMenu = [[NSMenu alloc] initWithTitle:@""];
    [fileMenuItem setTitle:@"File"];
 
But this does not work:
    NSMenuItem *fileMenuItem = [NSMenuItem new];
    NSMenu *fileMenu = [[NSMenu alloc] initWithTitle:@"File"];

DTW-Thalion:
`[[NSMenu alloc] initWithTitle:@"File"]` keeps its title until the menu is attached to a menu item, at which point `-[NSMenuItem setSubmenu:]` copied the item's title onto the submenu and the menu's own title was lost. This is issue #218.

A menu's title and the title of the item it is attached to are independent. Verified against Apple AppKit (macOS 26): after `[item setSubmenu: menu]` the menu keeps its `-initWithTitle:` title and the item keeps its own title, for both empty and non-empty item titles, and `-setSubmenu:forItem:` behaves the same way.

The only place that relied on the submenu's title matching the owning item's title was `-[NSMenuView sizeToFit]`, which located the item in a horizontal supermenu with `indexOfItemWithTitle:`. That now uses `indexOfItemWithSubmenu:` (already used elsewhere in NSMenu for the same purpose), so the lookup no longer depends on the titles being equal.

`Tests/gui/NSMenu/submenuTitle.m` covers the behaviour; its three title-preservation assertions fail before this change and pass after, with no regression in the existing NSMenu/NSMenuItemCell/NSMenuToolbarItem tests.

Fixes #218

### DIFF
--- a/Source/NSMenuItem.m
+++ b/Source/NSMenuItem.m
@@ -238,7 +238,6 @@ static Class imageClass;
   if (submenu != nil)
     {
       [submenu setSupermenu: _menu];
-      [submenu setTitle: _title];
     }
   [self setTarget: _menu];
   [self setAction: @selector(submenuAction:)];
--- a/Source/NSMenuView.m
+++ b/Source/NSMenuView.m
@@ -831,7 +831,7 @@ static float menuBarHeight = 0.0;
               NSMenuItemCell *msr;
 
               msr = [r menuItemCellForItemAtIndex:
-                [m indexOfItemWithTitle: [_attachedMenu title]]];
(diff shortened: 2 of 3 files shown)


fredkiefer (approved):
Great catch!

PR #681, opened 2026-07-21
Fix: let themes inset a menu's item cells from its edges

### LINKED ISSUE #220: theming does not allow to add padding to the menu
looks like menu size is computed at NSMenuView > sizeToFit, it is not possible from a custom theme to add padding between 

see the attached image

Screenshot 2023-12-21 at 13 19 27

DTW-Thalion:
The size and layout of a vertical menu are computed in `-[NSMenuView sizeToFit]` before drawing, with no way for a theme to add space between the menu's edges and its item cells. This is issue #220 (@fredkiefer you confirmed it and asked whether influencing `_leftBorderOffset` would be enough or whether a per-border offset was needed; the OP asked for per-side values / `NSEdgeInsets`).

This adds a `-[GSTheme menuItemAreaInsets]` returning `NSEdgeInsets`. The default is zero on every side, so existing themes are unchanged. When a theme returns non-zero insets:

- `-[NSMenuView sizeToFit]` grows the menu frame by the left/right insets (width) and top/bottom insets (height), and positions the title view above the inset item area;
- `-[NSMenuView rectOfItemAtIndex:]` offsets each item cell by the left and bottom insets.

Because drawing (`drawMenuRect:`) and hit testing (`indexOfItemAtPoint:`) both go through `rectOfItemAtIndex:`, they follow the inset layout automatically; the item cells keep their size and the padding around them is filled by the menu background.

`Tests/gui/NSMenuView/menuItemAreaInsets.m` installs a theme returning `NSEdgeInsetsMake(7, 11, 3, 13)` and checks the frame grows by 24×10, an item moves right 11 / up 3 without resizing, and a point at the moved item's centre still hit-tests to that item. The default insets are zero and the existing menu tests still pass.

I went with `NSEdgeInsets` (per the OP's suggestion, covering your "all borders" option) rather than only `_leftBorderOffset`; happy to narrow it if you prefer the smaller change.

Fixes #220

### DIFF
--- a/Headers/Additions/GNUstepGUI/GSTheme.h
+++ b/Headers/Additions/GNUstepGUI/GSTheme.h
@@ -1204,6 +1204,13 @@ APPKIT_EXPORT_CLASS
 - (CGFloat) menuItemHeight;
 - (CGFloat) menuSeparatorHeight;
 
+/**
+ * Padding between the edges of a vertical menu and its item cells.  The
+ * default is zero on every side; a theme can return larger insets to inset the
+ * list of items within the menu.
+ */
+- (NSEdgeInsets) menuItemAreaInsets;
+
 // NSColorWell drawing method
 - (NSRect) drawColorWellBorder: (NSColorWell*)well
                     withBounds: (NSRect)bounds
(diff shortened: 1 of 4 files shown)


fredkiefer (approved):
Thank you

PR #684, opened 2026-07-21
Fix: let themes round the corners of a menu

### LINKED ISSUE #223: theming does not allow to define border radius of menu popover
theming does not allow to define border radius of menu popover

Screenshot 2023-12-21 at 13 19 27

DTW-Thalion:
Fixes #223.

The menu window (`NSMenuPanel`) is already created borderless with a
`[NSColor clearColor]` background, and `NSMenuView` is non-opaque, so the menu
relies entirely on the theme painting its content. It looks rectangular only
because `-[GSTheme drawBackgroundForMenuView:withFrame:dirtyRect:horizontal:]`
fills the whole bounds with `NSRectFill`, giving a theme no way to round the
corners.

This adds `-[GSTheme menuBorderRadius]`, defaulting to zero so the rectangle is
filled and the per-edge borders drawn exactly as before. A larger value fills
and strokes the background through a rounded rectangle, leaving the corners to
the clear menu window.

On @gcasa's point: the corners are transparent only where a compositing
backend composites the window's clear background, so rounded menu corners
appear on such a backend (for example Wayland, or X11 with a compositor and an
ARGB visual). On a non-compositing backend the corners fall back to the window
backing. This change only decides what the theme paints; because the default
radius is zero, menus are pixel-identical to before unless a theme opts in.

Tests in `Tests/gui/NSMenuView/menuBorderRadius.m`: the default radius is zero
and fills the corner; a non-zero radius still fills the centre but leaves the
corner unfilled. The rounded-corner case fails against the current code and
passes with this change.

### DIFF
--- a/Headers/Additions/GNUstepGUI/GSTheme.h
+++ b/Headers/Additions/GNUstepGUI/GSTheme.h
@@ -1078,8 +1078,23 @@ APPKIT_EXPORT_CLASS
  * should not.  Instead themes should provide default colors in the
  * GSThemeDomain (in their Info.plist).</p>
  */
-- (NSColor *) menuBorderColorForEdge: (NSRectEdge)edge 
+- (NSColor *) menuBorderColorForEdge: (NSRectEdge)edge
                         isHorizontal: (BOOL)horizontal;
+
+/**
+ * <p>Returns the corner radius used to draw a menu view background.</p>
+ *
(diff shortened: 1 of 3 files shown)


fredkiefer (approved):
Any theme could have overridden this code, to implement round borders. Your code makes it easier to achieve this even without further code.

PR #687, opened 2026-07-21
Fix: reload the save panel browser when the delegate filter changes

DTW-Thalion:
A save panel delegate that implements `-panel:shouldShowFilename:` filters the browser, but clearing or replacing that delegate leaves the previously hidden files hidden.

`-setDelegate:` ended with `-validateVisibleColumns`, which reloads a column only when `-browser:isColumnValid:` reports it invalid. That check walks the currently shown cells and returns invalid when one of them should now be hidden, so it catches files a new filter excludes but never files a removed filter should show again (there is a standing `FIXME` on it to that effect). Resetting the delegate therefore leaves the browser filtered.

Reload the browser when the show-filename filter is added or removed, matching what `-setShowsHiddenFiles:` and `-setDirectory:` already do for the changes they make. The reload only happens when that filter state actually changes, so the `-setDelegate: nil` call made during panel setup (before any filter) still takes the light path.

`Tests/gui/NSSavePanel/setDelegate_reload.m` already covered this as a hopeful case; it now passes, so the hopeful marker is removed.


Source/NSSavePanel.m
```diff
@@ -1496,7 +1498,13 @@ - (void) setDelegate: (id<NSOpenSavePanelDelegate>)aDelegate
     _delegateHasSelectionDidChange = NO;      
 
   [super setDelegate: aDelegate];
-  [self validateVisibleColumns];
+  /* -validateVisibleColumns can hide files that a new filter excludes, but it
+     cannot restore files that a removed filter should show again, so a change
+     to the show-filename filter needs a full reload. */
+  if (_delegateHasShowFilenameFilter != hadShowFilenameFilter)
```
fredkiefer: Not sure here. The two delegates might have completely different filters. Wouldn't it be better to reload when either has a filter? That way we get rid of the old filter and apply the new one, if any.

DTW-Thalion: Done. It now reloads whenever the old or the new delegate filters filenames, so replacing one filter with another drops the old and applies the new. Added a test case that replaces a filtering delegate with one that hides nothing and checks the previously hidden file comes back (it fails without the change, since -validateVisibleColumns can hide but not restore).

PR #688, opened 2026-07-21
Fix: prevent unbounded recursion when autohidesScrollers is set

### LINKED ISSUE #234: Possible Segmentation Fault with NSScrollView?
Maybe my source is totally wrong, but I have been trying to fix this behabiour for days and looking at the stack trace I amb begining to think this is a bug from the libraries.

Looks like there is a recursive cycle in the stack trace, and appears to be caused by an interaction between NSClipView and NSScrollView.

This is the GDB output causing Segmentation Fault:

```bash
#596 0x0000fffff7c79e34 in -[NSScrollView reflectScrolledClipView:] (self=0xaaaaab5c6068, _cmd=<optimized out>, 
    aClipView=<optimized out>) at NSScrollView.m:908
#597 0x0000fffff7b3a304 in -[NSClipView setFrame:] (self=0xaaaaab5cbee8, _cmd=<optimized out>, rect=...) at NSClipView.m:570
#598 0x0000fffff7c7b360 in -[NSScrollView tile] (self=<optimized out>, _cmd=<optimized out>) at NSScrollView.m:1303
#599 0x0000fffff7c79d10 in -[NSScrollView reflectScrolledClipView:] (self=0xaaaaab5c6068, _cmd=<optimized out>, 
    aClipView=<optimized out>) at NSScrollView.m:895
#600 0x0000fffff7b3a304 in -[NSClipView setFrame:] (self=0xaaaaab5cbee8, _cmd=<optimized out>, rect=...) at NSClipView.m:570
#601 0x0000fffff7c7b360 in -[NSScrollView tile] (self=<optimized out>, _cmd=<optimized out>) at NSScrollView.m:1303
```

Looking at NSClipView.m and NSScrollView.m they seem to trigger a series of calls that eventually loop back recursively, leading to an infinite loop?

The only way for me to reproduce this bug, is with a window containing scroll bar, and calling the 'setFrame' of the scroll bar on every window size change. 

The window resize must be done very fast until it chrashes.

I attach a video of the application crashing, and the source of the application itself.

https://github.com/gnustep/libs-gui/assets/7429620/6da5005c-3281-47ce-ba5f-49b0a0ec2745

[test-bug.zip](https://github.com/gnustep/libs-gui/files/13789788/test-bug.zip)

DTW-Thalion:
Fixes #234.

## Problem

With `autohidesScrollers = YES`, resizing a scroll view whose content is
close to the frame size can crash with a segmentation fault. The stack
trace is an unbounded `-reflectScrolledClipView:` / `-tile` recursion:

```
-[NSScrollView reflectScrolledClipView:]   (autohide toggles a scroller)
-[NSScrollView tile]
-[NSClipView setFrame:]
-[NSScrollView reflectScrolledClipView:]
...
```

`-reflectScrolledClipView:` decided each scroller independently by
comparing the document size to the current clip view bounds. Showing or
hiding a scroller calls `-tile`, which resizes the clip view and calls
`-reflectScrolledClipView:` again. A vertical scroller removes width and
a horizontal scroller removes height, so at sizes where the content is
within one scroller width of fitting, the two decisions keep flipping
each other and the method recurses until the stack is exhausted.

This also explains the question raised in the issue thread about how
disabling a scroller could create the need for that same scroller: the
coupling is across axes. Removing the vertical scroller gives back
width, which lets the horizontal scroller be removed, which gives back
height, which brings the vertical scroller back.

## Fix

Decide both scrollers together in a new private `-_autohideScrollers`.
It reconstructs the area the clip view would have with no scrollers (its
current bounds plus the footprint of whatever scrollers are currently
shown) and evaluates each axis against that area, allowing for the space
the other scroller occupies. The evaluation only ever turns scrollers
on, so it settles on one configuration. Because the reconstructed area
does not depend on which scrollers are currently shown, the `-tile`
triggered by applying the decision re-enters `-reflectScrolledClipView:`
with the same result, where `-setHasVerticalScroller:` /
`-setHasHorizontalScroller:` change nothing and the recursion stops. No
ivar or ABI change is needed.

The non-autohiding path is unchanged.

## Testing

Added `Tests/gui/NSScrollView/autohidesScrollers.m`. It lays out an
autohiding scroll view at the document size that used to recurse and
asserts the scroller visibility for fitting and overflowing documents.
Against the current library the test aborts (the crash); with this
change it passes. The existing NSScrollView, NSClipView, NSScroller,
NSTableView, NSTextView and NSBrowser test directories still pass.

### DIFF
--- a/Source/NSScrollView.m
+++ b/Source/NSScrollView.m
@@ -106,6 +106,7 @@ typedef struct _scrollViewFlags
 /* GNUstep private methods */
 - (void) _synchronizeHeaderAndCornerView;
 - (void) _themeDidActivate: (NSNotification*)notification;
+- (void) _autohideScrollers;
 @end
 
 @implementation NSScrollView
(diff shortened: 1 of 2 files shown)


fredkiefer (approved):
That is a very nice solution to this old issue. It took me some time to understand why we need the three tests and also why they are sufficient. You already wrote in the method comment that we are only interested in state changes to on. But I only understood this afterwards.

Issue #215, opened 2023-12-12
GSImageMagickImageRep problems


rmottola:
while investigating https://github.com/gnustep/apps-gworkspace/pull/20, I installed image magick and enabled support for it.

A small local patch is needed to support v7 (will try to insert version detection, since most OS has 6 or 7 available).

Several issues arise from GWorkspace now which needs to be analyzed, since they don't happen without IM support enabled.

When loading a PDF:
```
2023-12-12 09:19:08.875 GWorkspace[638:644] ImageMagick: zero-length blob not permitted `' @ error/blob.c/BlobToImage/450
2023-12-12 09:19:08.883 GWorkspace[638:638] ImageViewer: trying to display inconsistent image

2023-12-12 09:24:14.112 GWorkspace[638:644] ImageMagick: delegate library support not built-in `/tmp/magick-kjld8bf6U-7gHFW182cAJkuNv6amEkms' (XML) @ error/profile.c/ValidateXMPProfile/1943
```

But it also try to load an HTML
```
2023-12-12 09:26:02.549 GWorkspace[638:644] ImageMagick: zero-length blob not permitted `' @ error/blob.c/BlobToImage/450
2023-12-12 09:26:02.571 GWorkspace[638:638] ImageViewer: trying to display inconsistent image
```

DTW-Thalion:
@rmottola See PR #689 and some observations in the PR body I noted for you.

PR #689, opened 2026-07-21
Fix: harden GSImageMagickImageRep and NSPDFImageRep against non-image and short data

DTW-Thalion:
Addresses several of the problems in #215. Two commits, one per defect.

## 1. Out-of-range exception for short data

`+[GSImageMagickImageRep imageRepsWithData:allImages:]` probed the last
`SIGNATURE_LENGTH` (18) bytes for a Targa trailer with
`NSMakeRange([data length] - 18, 18)`. For data shorter than 18 bytes the
unsigned subtraction wraps and `-[NSData getBytes:range:]` raises
`NSRangeException`, so a truncated or very small image aborted the caller.
The probe now runs only when the data is at least `SIGNATURE_LENGTH` long.

## 2. Documents and undecodable formats treated as images

`+canInitWithData:` and `+imageUnfilteredFileTypes` accepted every format
ImageMagick knows, including document and vector formats decoded through an
external delegate (PDF, PostScript) and formats with no decoder (HTML).
NSImage could route those to the rep, where a PDF invoked the Ghostscript
delegate and an HTML file produced an empty representation — the "trying to
display inconsistent image" reports in the issue.

The rep now claims only formats ImageMagick can decode from a blob, using
the `decoder` and `blob_support` fields of the format's `MagickInfo`.
Raster formats (JPEG, PNG, GIF, TIFF, BMP, TGA, ...) keep `blob_support`
and a decoder; PDF/PS have no blob support, and HTML has no decoder.

`NSPDFImageRep` rasterises PDFs through `GSImageMagickImageRep` and then
read the first page unconditionally, raising an out-of-range exception
(shown as a blocking error panel in a running application) when the PDF
produced no page. That access is now guarded by the page count, so an
unrenderable PDF yields a zero-size representation instead of raising.
PDF rendering itself is unaffected: `NSPDFImageRep` still rasterises valid
PDFs through `GSImageMagickImageRep`.

## Testing

- `Tests/gui/GSImageMagickImageRep/shortData.m` — short and empty data yield
  no representations rather than raising.
- `Tests/gui/GSImageMagickImageRep/nonRasterFormats.m` — the rep accepts a
  raster format and rejects PDF and HTML, and does not advertise pdf/html.
- `Tests/gui/NSPDFImageRep/unrenderableData.m` — an unrenderable PDF and
  empty data produce a zero-size representation instead of raising.

The `GSImageMagickImageRep` tests look the class up with `NSClassFromString`
and skip when the GUI was built without ImageMagick. Verified against
ImageMagick 6.9.12: without the changes the new tests fail (exception, or
PDF/HTML accepted); with them they pass, real JPEG/PNG loading and valid
PDF rendering are unaffected, and the NSImage / NSBitmapImageRep test
directories still pass.

## Not fixed here

I could not reproduce the JPEG crash from the issue with ImageMagick
6.9.12 (plain, profiled and real-world sRGB JPEGs all load), so I have left
that alone rather than guess at it. A CMYK JPEG decodes but ends up as a
0x0 image through NSImage, which also seems worth a separate look.

### DIFF
--- a/Source/GSImageMagickImageRep.m
+++ b/Source/GSImageMagickImageRep.m
@@ -51,7 +51,21 @@
 #include <magick/MagickCore.h>
 #endif
 
-@implementation GSImageMagickImageRep 
+/* Whether we should try to decode a given format.  We only handle formats
+   ImageMagick can read directly from an in-memory blob: a format without a
+   decoder (such as HTML) produces nothing, and a document or vector format
+   without blob support (such as PDF or PostScript) is decoded through an
+   external delegate that can block on malformed input, neither of which
+   belongs in an image representation. */
+static BOOL
+GSImageMagickCanDecode(const MagickInfo *info)
+{
+  return (info != (const MagickInfo *)NULL
+    && info->decoder != (DecodeImageHandler *)NULL
+    && info->blob_support != MagickFalse);
+}
+
+@implementation GSImageMagickImageRep
 
 + (void) initialize
 {
@@ -136,7 +150,7 @@
   Image *images;
   Image *image;
   char signature[SIGNATURE_LENGTH];
-  
+
   // Set the background color to transparent
   // (otherwise SVG's are rendered against a white background by default)
 #if (MagickLibVersion >= 0x700)
@@ -195,16 +209,21 @@
 {
   char buf[32];
   ExceptionInfo *exception;
-  const MagicInfo *info;
+  const MagicInfo *magic;
+  const MagickInfo *magick = (const MagickInfo *)NULL;
 
   memset(buf, 0, 32);
   [data getBytes: buf length: 32];
 
   exception = AcquireExceptionInfo();
-  info = GetMagicInfo((const unsigned char *)buf, 32, exception);
+  magic = GetMagicInfo((const unsigned char *)buf, 32, exception);
+  if (magic != (const MagicInfo *)NULL)
+    {
+      magick = GetMagickInfo(GetMagicName(magic), exception);
+    }
   DestroyExceptionInfo(exception);
 
-  return (info != NULL);
+  return GSImageMagickCanDecode(magick);
 }
 
 + (NSArray *) imageUnfilteredFileTypes
@@ -221,7 +240,10 @@
 
       for (i=0; i<size; i++)
 	{
-	  [array addObject: [[NSString stringWithUTF8String: list[i]->name] lowercaseString]];
+	  if (GSImageMagickCanDecode(list[i]))
+	    {
+	      [array addObject: [[NSString stringWithUTF8String: list[i]->name] lowercaseString]];
+	    }
 	}
       RelinquishMagickMemory(list);
       DestroyExceptionInfo(exception);
--- a/Source/GSImageMagickImageRep.m
+++ b/Source/GSImageMagickImageRep.m
@@ -51,12 +51,12 @@
 #include <magick/MagickCore.h>
 #endif
 
-/* Whether we should try to decode a given format.  We only handle formats
-   ImageMagick can read directly from an in-memory blob: a format without a
-   decoder (such as HTML) produces nothing, and a document or vector format
-   without blob support (such as PDF or PostScript) is decoded through an
-   external delegate that can block on malformed input, neither of which
-   belongs in an image representation. */
+/* Whether GSImageMagickImageRep itself should claim a given format.  We only
+   claim formats ImageMagick can read directly from an in-memory blob: a format
+   without a decoder (such as HTML) produces nothing, and a document or vector
+   format without blob support (such as PDF or PostScript) has a dedicated
+   representation (NSPDFImageRep, NSEPSImageRep) that renders it through an
+   external delegate, bounded by a timeout (see imageRepsWithData:allImages:). */
 static BOOL
 GSImageMagickCanDecode(const MagickInfo *info)
 {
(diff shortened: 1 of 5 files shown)


rmottola:
At first, the patch idea looks good, but wanted to test it a little.

I tested this using ImageMagick 6.9. I removed PDFKit library and used GWorkspace content inspector as "test".
I too cannot reproduce issues with JPEG any more.

Once ImageMagick is installed, I can see webp files, so I can confirm "it works" and it does before and after your PR.


```
2026-07-22 12:20:36.946 GWorkspace[20233:20233] ImageViewer - _readImage: /home/multix/Documents/storia Unitas.pdf
2026-07-22 12:20:36.949 GWorkspace[20233:20272] ImageMagick: attempt to perform an operation not allowed by the security policy `../../magick/constitute.c' @ error/constitute.c/IsCoderAuthorized/424
2026-07-22 12:20:36.949 GWorkspace[20233:20272] (not resized) no valid image representation for /home/multix/Documents/storia Unitas.pdf

```

I thin there is some error or limitation in my imagemagick installation.

rmottola:
Need to enable

`  <policy domain="coder" rights="read" pattern="PDF" />`

in` /etc/ImageMagick-6/policy.xml `

However, I get no content displayed:

```
2026-07-22 12:29:12.659 GWorkspace[26781:26781] ImageViewer - displayPath - readImage
2026-07-22 12:29:12.659 GWorkspace[26781:26781] ImageViewer - _readImage: /home/multix/Documents/storia Unitas.pdf
2026-07-22 12:29:12.958 GWorkspace[26781:26840] (not resized) no valid image representation for /home/multix/Documents/storia Unitas.pdf

```

the security policy error is gone, but I don't get a valid image rep back.

PS: using ImageMagic viewer `display-im6`, I can open said PDFs, so it can open them as valid.

DTW-Thalion:
Thanks for testing. JPEG and webp are unaffected by this change, which matches what you saw: it only stops the crash on truncated data and stops GSImageMagickImageRep from accepting formats it cannot decode from an in-memory blob.

PDF is one of those. PDF and PostScript have an ImageMagick decoder but no blob support, so ImageMagick decodes them by running an external delegate (Ghostscript), and that delegate can block on malformed input. This PR declines them for that reason, so canInitWithData: returns NO and no representation is produced. display-im6 shows the file because it invokes that delegate directly.

That is why you get no image for the PDF even after allowing PDF in policy.xml: the representation is declined before ImageMagick is asked to decode. If it is worth loading PDFs through the delegate despite the hang risk, I can gate that path behind a timeout rather than declining outright. Let me know which you would prefer.

rmottola:
@DTW-Thalion well, but then if you disable it totally, I think NSPDFImageRep looses its use. Reading your original description, I understood a guard against 0 page and unrenderable PDFs, but still allow one page or first-page-of-many.

DTW-Thalion:
> @DTW-Thalion well, but then if you disable it totally, I think NSPDFImageRep looses its use. Reading your original description, I understood a guard against 0 page and unrenderable PDFs, but still allow one page or first-page-of-many.

I wasn't clear enough in my note, that's my fault. I will gate the path behind a timeout and see how well that works and hopefully it will be a rare instance.

DTW-Thalion:
The PDF path is not disabled. NSPDFImageRep is registered ahead of GSImageMagickImageRep in `+[NSImageRep initialize]`, so `imageRepClassForData:` always selects it for `%PDF` data; the `canInitWithData:` change only stops GSImageMagickImageRep from claiming formats it cannot read from an in-memory blob, and that class was never the one handling PDFs. A valid one-page or first-page-of-many PDF renders as before, through NSPDFImageRep.

For the hang risk that motivated declining those formats, I have added a timeout rather than relying on the format check. `imageRepsWithData:allImages:` now bounds the ImageMagick `TimeResource` to 30 seconds around `BlobToImage` when the content is decoded through an external delegate such as Ghostscript, and restores the previous limit afterwards. A valid document decodes well within that; malformed input that would otherwise block is aborted and returns no representation instead of hanging. The limit applies only to the delegate path, so ordinary in-memory decodes are unchanged.

That also covers what you saw: when the delegate yields no pages, NSPDFImageRep now returns an empty representation instead of raising on an out-of-range page, which is why you get "no valid image representation" rather than a crash.

The 30 second value is a single define (`GS_IMAGEMAGICK_DELEGATE_TIME_LIMIT`) if you want to tune it. I could not exercise the delegate timeout in CI, since it needs Ghostscript and a pathological input, so it is worth a look on your setup.

DTW-Thalion:
@rmottola can you try this again to make sure you can see the PDF per my comment above?

fredkiefer (approved):
As Riccardo had no further comments, let's merge this

rmottola:
> As Riccardo had no further comments, let's merge this

Just closing.... without testing isn't good. I didn't reply because I didn't have time for the last tests.. or they weren't good and had no time to report. Didn't give green light either though. Did you test it?

Current merged tree doesn't compile with ImageMagick though. Since it is now merged, I opened a new bug report, sorry.

DTW-Thalion:
> > As Riccardo had no further comments, let's merge this
> 
> Just closing.... without testing isn't good. I didn't reply because I didn't have time for the last tests.. or they weren't good and had no time to report. Didn't give green light either though. Did you test it?
> 
> Current merged tree doesn't compile with ImageMagick though. Since it is now merged, I opened a new bug report, sorry.

Will look at this this morning.

ivucica:
If it's not building, it might warrant submitting a revert commit (a rollback) followed by a roll-forward once a tested version is ready?

DTW-Thalion:
PR #888  - should solve this.

cc @rmottola @ivucica 

@fredkiefer might be worth adding an --enable-imagemagick flag to the CI workflow for this actually now that I am thinking about it.

rmottola:
> If it's not building, it might warrant submitting a revert commit (a rollback) followed by a roll-forward once a tested version is ready?

it is not building only if you explicitely enable imagemagic. I will exactly check what's wrong, I thought to have IM6 installed, but I may be confused, I have two systems where I am testing. Will update

rmottola:
Since IM6 and IM7 can coexist, to ease testing, I have IM6 on Devuan Laptop and IM7 on FreeBSD workstation. So things cannot get confused at configure - except for myself forgetting.

I confirm the code compiles fine with IM6 on linux, yet I get no valid image representations for PDFs.

DTW-Thalion:
> Since IM6 and IM7 can coexist, to ease testing, I have IM6 on Devuan Laptop and IM7 on FreeBSD workstation. So things cannot get confused at configure - except for myself forgetting.
> 
> I confirm the code compiles fine with IM6 on linux, yet I get no valid image representations for PDFs.

Interesting... let me see if I can figure out what's happening.

rmottola:
@DTW-Thalion I think I get a valid NSPDImageRep, but the GWorkspace inspector is not capable of properly resizing it, since it expects an NSBitmapImageRep and skips. I will look into it. Possibly it is the first time ever that this code path is fully implemented and reached. Usually it is hidden by the more powerful PDFKit. Will look into it and open a bug in GWorkspace. The NSPDFimageRep has sane looking size and pixel information, so although I cannot "see" it yet, I suppose it is fine.

PR #692, opened 2026-07-21
tests: local render/interaction harness with NSSavePanel, NSOpenPanel and NSWindowController

DTW-Thalion:
Some classes, particularly panels and controllers, can only be checked by
running them against a real window server: rendering their content and driving
them. This adds a small shared harness for that and uses it on three classes.

## The harness: `Tests/gui/GSRenderTest.h`

A test can render a view or window (whole subview hierarchy) to a bitmap, save
a PNG to look at, compare against a locally captured golden, sample regions of
the result structurally, deliver a synthesised mouse click, and run a panel's
modal session to completion without blocking on input.

Goldens and PNGs are written outside the source tree (`GS_TEST_ARTIFACTS`, else
`/tmp/gs-render-artifacts`) and are never committed: they are a local aid,
while the region assertions are what a headless run checks. Every helper needs
a backend, so a test keeps the usual `START_SET` / `SKIP` guard and skips
cleanly without a display. The header is pulled in through a one-line
`GNUmakefile.preamble`.

## Users

- NSSavePanel: `render.m` renders the panel and checks its browser and name
  field drew content; `interact.m` drives the modal session to OK and Cancel.
- NSOpenPanel: the same, for the open panel.
- NSWindowController: `lifecycle.m` loads a programmatic window lazily, checks
  windowDidLoad is sent once, shows and renders it, and closes it; `config.m`
  covers the window and cascade/autosave-name round-trips.

## One fix the exercise surfaced

`-[NSWindowController window]` checked `-isWindowLoaded` after `-loadWindow` to
decide whether to send `-windowDidLoad`, but that flag is only set on the nib
path. A controller whose `-loadWindow` builds the window and calls `-setWindow:`
was never marked loaded and never received `-windowDidLoad`. The completion is
now keyed off the window existing. `lifecycle.m` fails against the old code and
passes with the change.

## Verification

- Genuine display-less: the suites skip cleanly.
- Under `xvfb` with the X11 backend (the headless-CI mechanism): NSSavePanel,
  NSOpenPanel and NSWindowController all run and pass (12 + 8 + 16).
- With a live display: the same. NSWindow's tests are unaffected by the fix.


DTW-Thalion:
@fredkiefer - this PR needed for all the heavy display and rendering class testing to follow, so need this one prioritized before I can continue.

DTW-Thalion:
The only failing job here is the cairo/Xvfb backend, and the failure is not related to this PR. It aborts in gui/NSApplication/basic.m with `-[XGServer runLoopShouldBlock:]: message sent to deallocated instance`, the master-wide regression tracked in #700 (introduced by e82a3c5d). The other three jobs pass, and this branch's own NSOpenPanel, NSSavePanel and NSWindowController tests run without failure.

This PR was green until e82a3c5d landed on master on 2026-07-22. Since the PR check tests the branch merged with current master, it now inherits that regression. Nothing needs to change here until #700 is resolved.

Source/NSWindowController.m
```diff
@@ -312,7 +312,10 @@ - (NSWindow *) window
 	}
 
       [self loadWindow];
-      if ([self isWindowLoaded]) 
+      /* -loadWindow may supply the window through -setWindow: (for example a
+         subclass that builds it directly) rather than by loading a nib, so key
+         the completion off the window existing, not off -isWindowLoaded. */
+      if (_window != nil)
```
fredkiefer, line 318: We will need to investigate this work around later on. But as the next method call, -_windowDidLoad, sets this value to true anyway, we should be fine here.

DTW-Thalion: Sounds good to me. It seems to work fine, but I have another bunch of Classes to test so we'll see if it holds up.

rfm (approved):
I have tested this on a couple of systems, and I have done a visual check of the code to the extent of confirming that nothing appears wrong (or at least, harmful/dangerous), though I can't really be sure it's doing exactly what's intended in the GUI.
Given that it's blocking other stuff, I think we can go ahead.

Tests/gui/GSRenderTest.h
```diff
@@ -0,0 +1,211 @@
...
+   and lays out for real without blocking on the user.  Returns the modal
+   result. */
+/* Run a panel modally but end the session ourselves by performing `dismiss`
+   (e.g. @selector(ok:) or @selector(cancel:)), so the panel displays and lays
+   out without blocking on the user.  A manual modal session is pumped a few
+   cycles to let the panel appear, then dismissed; this avoids relying on a
+   timer firing inside -runModal (which will not wake without server events).
+   Returns the modal result code. */
+static inline NSInteger
+GSRunModalDismissing(id panel, SEL dismiss, NSTimeInterval delay)
```
fredkiefer, line 193: I don't see where `delay` gets used. Could you please either remove it or use it?

Tests/gui/NSOpenPanel/interact.m
```diff
@@ -0,0 +1,65 @@
+#import "Testing.h"
+#import "GSRenderTest.h"
```
fredkiefer, line 2: Why not import the header with a ".." prefix? That way you wouldn't need the GNUmakefile.preamble

Tests/gui/GSRenderTest.h
DTW-Thalion: Done in #722.

Tests/gui/NSOpenPanel/interact.m
DTW-Thalion: Done in #722.

PR #699, opened 2026-07-22
ci: start the distributed notification server before running tests

DTW-Thalion:
Tests that post a distributed notification rely on the gdnc server. On a headless runner gdnc is launched on demand by the first such test, and that startup intermittently times out with "unable to contact GDNC server", aborting the set. NSHelpManager basic.m (context help mode) and NSMenuItem encoding.m (key equivalent mask) fail this way.

Start gdnc before running the tests in both the matrix job and the cairo/Xvfb job so the daemon is already up. With this change all four jobs run the full suite green.


rfm (approved):
OK, looks fine.

PR #702, opened 2026-07-23
tests: add NSTabViewController tests

DTW-Thalion:
Adds a test suite for NSTabViewController:

- config.m: the tab style, transition options and title propagation flag round-trips.
- items.m: adding, inserting, removing and looking up tab items backed by view controllers, and the selected index.
- render.m: renders a tab view with two labelled tabs and checks the tab bar drew.
- delegate.m: the controller's tab-view delegate answer and its empty toolbar identifier lists.

The item and render sets use a window server and skip cleanly without one.


Tests/gui/NSTabViewController/render.m
```diff
@@ -0,0 +1,76 @@
+#import "Testing.h"
+#import "GSRenderTest.h"
```
fredkiefer, line 2: Again putting the ".." here would simplify things.

fredkiefer (approved):

Tests/gui/NSTabViewController/render.m
DTW-Thalion: Done in #764 - render.m now uses "../GSRenderTest.h" and the -I.. preamble is gone.

PR #703, opened 2026-07-23
Fix: NSTabViewController creates its tab view

DTW-Thalion:
A programmatic NSTabViewController (created without a nib) never built a tab view: -tabView returned -[NSViewController view], and that view stayed nil because -loadView does nothing without a nib name. As a result -addTabViewItem: and the other tab-item methods, which forward to the tab view, had no effect.

This adds a -loadView that creates an NSTabView when there is no nib, so a freshly created controller has a working tab view. A test drives item management on a controller whose tab view was never set by hand.


Tests/gui/NSTabViewController/GNUmakefile.preamble
```diff
@@ -0,0 +1 @@
+ADDITIONAL_INCLUDE_DIRS += -I..
```
fredkiefer, line 1: This file is not even used here. I will merge it anyway.

fredkiefer (approved):

PR #704, opened 2026-07-23
Fix: NSTabViewController reports -1 when no tab is selected

DTW-Thalion:
-selectedTabViewItemIndex asked the tab view for the index of its selected item, but with no selection that returned NSNotFound (a large positive value) rather than -1. Callers that compare the index against -1 to detect "no selection" were misled.

This returns -1 when there is no selected tab item.


Tests/gui/NSTabViewController/GNUmakefile.preamble
```diff
@@ -0,0 +1 @@
+ADDITIONAL_INCLUDE_DIRS += -I..
```
fredkiefer, line 1: Again this file is not used

fredkiefer (approved):

PR #705, opened 2026-07-23
Fix: NSTabViewController initialises its transition and title-propagation defaults

DTW-Thalion:
A freshly created NSTabViewController did not match the expected defaults:

- transitionOptions was zero (no options) rather than allowing user interaction during transitions.
- canPropagateSelectedChildViewControllerTitle was NO rather than YES.

Because -init did not funnel through the designated initialiser, neither default was applied. This sets them in -initWithNibName:bundle: and routes -init through it.

It also corrects the title that propagation sets: -setSelectedTabViewItemIndex: took the tab item itself as the title instead of the item's label, so with propagation enabled the controller title was set to a tab item rather than its text.


Source/NSTabViewController.m
```diff
@@ -31,6 +31,23 @@
...
+    {
+      _transitionOptions = NSViewControllerTransitionAllowUserInteraction;
+      _canPropagateSelectedChildViewControllerTitle = YES;
+    }
+  return self;
+}
+
+- (instancetype) init
+{
+  return [self initWithNibName: nil bundle: nil];
```
fredkiefer: I am a bit unsure whether this call could lead to a cycle in a subclass of NSTabViewController. I definitely need to fresh up my recollection of how Objective C resolves super calls. Maybe it would be better to move this implementation to NSViewController? Or should I just merge as there are currently no subclasses?

@rfm can you help out here?

fredkiefer: I just added this method to NSViewController to hopefully fix some test issues. I will remove the method here. If you still see any issues with the approach, please say so.

fredkiefer (approved):

Source/NSTabViewController.m
DTW-Thalion: The approach is correct. NSTabViewController's -initWithNibName:bundle: chains to super and your NSViewController -init funnels through that designated initializer, so there is no recursion or double init. The remaining -loadView override only builds the default tab view, which is class specific, so it stays. No issues.

PR #706, opened 2026-07-23
tests: add NSSplitViewController tests

DTW-Thalion:
Adds a test suite for NSSplitViewController:

- config.m: the minimum inline sidebar thickness round-trips.
- items.m: setting the split items, keeping their order, looking one up by its view controller, and the split view round-trip.
- render.m: renders a split view with two boxed panes and checks a pane drew.

The item and render sets use a window server and skip cleanly without one.


Tests/gui/NSSplitViewController/GNUmakefile.preamble
```diff
@@ -0,0 +1 @@
+ADDITIONAL_INCLUDE_DIRS += -I..
```
fredkiefer, line 1: Maybe we should clean these up later on in one go.

fredkiefer (approved):

Tests/gui/NSSplitViewController/GNUmakefile.preamble
DTW-Thalion: Cleaned up in one go in #764: the three remaining render-test dirs (NSPanel, NSSplitViewController, NSTabViewController) now import the harness as "../GSRenderTest.h" and their GNUmakefile.preamble files are removed.

PR #707, opened 2026-07-23
Fix: NSSplitViewController creates its split view

DTW-Thalion:
A programmatic NSSplitViewController (created without a nib) never built a split view: -splitView returned -[NSViewController view], which stayed nil because -loadView does nothing without a nib name.

This adds a -loadView that creates an NSSplitView when there is no nib, so a freshly created controller has a working split view.


Tests/gui/NSSplitViewController/GNUmakefile.preamble
```diff
@@ -0,0 +1 @@
+ADDITIONAL_INCLUDE_DIRS += -I..
```
fredkiefer, line 1: Once more

fredkiefer (approved):

PR #708, opened 2026-07-23
Fix: NSSplitViewController creates its items array

DTW-Thalion:
The split-item list was never created: _splitViewItems stayed nil, so -addSplitViewItem: and -insertSplitViewItem:atIndex: sent -insertObject:atIndex: to nil and did nothing, and a keyed-decoded controller dropped its decoded items for the same reason.

This creates the mutable list in the designated initialiser (with -init routed through it) and in -initWithCoder:, so items can be added to a freshly created or decoded controller.


Source/NSSplitViewController.m
```diff
@@ -34,6 +34,22 @@
...
+{
+  self = [super initWithNibName: nibNameOrNil bundle: nibBundleOrNil];
+  if (self != nil)
+    {
+      _splitViewItems = [[NSMutableArray alloc] init];
+    }
+  return self;
+}
+
+- (instancetype) init
```
fredkiefer: Again, as noted for the NSTabViewController, this could be dangerous. Also have the same code her points to moving it to the super class.

fredkiefer (approved):

Source/NSSplitViewController.m
DTW-Thalion: Resolved by the same move: -init now lives in NSViewController and the redundant override here is gone, with -initWithNibName:bundle: chaining to super. The -loadView override stays because it builds the default split view. Looks good.

PR #722, opened 2026-07-23
tests: simplify render harness includes and drop unused modal delay

DTW-Thalion:
Follow-up to the local render/interaction harness (#692), addressing review comments.

- GSRunModalDismissing declared a delay argument that is never used; the modal session is pumped a fixed number of cycles rather than timed. Drop the argument and its call-site values, and remove the stale duplicate comment above the function.
- Import GSRenderTest.h with a "../" prefix from the NSSavePanel, NSOpenPanel and NSWindowController test cases, so the per-directory GNUmakefile.preamble that only added -I.. is no longer needed. Those three preambles are removed.

No behaviour change. NSSavePanel, NSOpenPanel and NSWindowController build and their tests pass or skip as before.


fredkiefer (approved):
That was fast

DTW-Thalion:
> That was fast

We aim to please. 😁

PR #725, opened 2026-07-23
Fix: retain the view controller an NSSplitViewItem holds

DTW-Thalion:
initWithViewController: retained the view controller it was given, but setViewController: and the keyed decoder assigned it without retaining, and there was no dealloc to release it. So an item leaked the view controller it was created with, and after a later setViewController: it held a dangling pointer to a released controller.

setViewController: and initWithCoder: now retain, and a new dealloc releases, so the item consistently owns its view controller.

Added Tests/gui/NSSplitViewItem/memory.m, which uses a view controller subclass that records its deallocation to check the setter keeps it alive and dealloc releases it.

### DIFF
--- a/Tests/gui/NSSplitViewItem/memory.m
+++ b/Tests/gui/NSSplitViewItem/memory.m
@@ -1,0 +1,52 @@
+/* NSSplitViewItem owns the view controller it holds: setViewController: keeps
+   it alive and a deallocated item releases it. */
+#include "Testing.h"
+
+#include <Foundation/NSAutoreleasePool.h>
+
+#include <AppKit/NSApplication.h>
+#include <AppKit/NSSplitViewItem.h>
+#include <AppKit/NSViewController.h>
+
+static BOOL vcDeallocated;
+
+@interface ProbeViewController : NSViewController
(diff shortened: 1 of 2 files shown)


Tests/gui/NSSplitViewItem/memory.m
```diff
@@ -0,0 +1,52 @@
...
+int main()
+{
+  CREATE_AUTORELEASE_POOL(arp);
+  NSAutoreleasePool *pool;
+  NSSplitViewItem *item;
+  ProbeViewController *vc;
+
+  /* Keep the only reference to the item so that releasing it below runs its
+     deallocation. */
+  pool = [NSAutoreleasePool new];
```
fredkiefer, line 33: Why not `CREATE_AUTORELEASE_POOL(pool)`?

fredkiefer (approved):

Tests/gui/NSSplitViewItem/memory.m
DTW-Thalion: Opened #798 for this, since #725 merged first.

PR #727, opened 2026-07-23
Fix: NSAnimation -setProgressMarks: leaves the mark array usable

DTW-Thalion:
setProgressMarks: cleared the current marks with GSIArrayEmpty and then added the new ones. GSIArrayEmpty frees the array's backing store (it is the teardown call used in dealloc), so the following addProgressMark: added to a freed array and tripped an assertion. Using GSIArrayRemoveAllItems clears the contents while keeping the array usable.

Added Tests/gui/NSAnimation/setmarks.m covering replacing the marks, clearing them with nil, and adding a mark afterwards.


DTW-Thalion:
Folding this into a single progress-marks fix: verifying setProgressMarks: requires calling -progressMarks, which needs the cache-ownership fix too, so the two are being submitted together.

PR #728, opened 2026-07-23
Fix: NSAnimation owns the progress-mark numbers it caches

DTW-Thalion:
-progressMarks builds a cache of NSNumber objects, releases that cache when it is refreshed (after the marks change) and releases it again in -dealloc. The numbers were created with +numberWithFloat: and stored without being retained, so each refresh released numbers the animation did not own. Under a run that calls -progressMarks repeatedly with a change between the calls this over-releases the numbers.

Retain the numbers when they are cached, balancing the releases on refresh and in -dealloc.

Added Tests/gui/NSAnimation/cache.m, which calls -progressMarks after each of an addition, a removal and another addition.


DTW-Thalion:
Superseded by a combined progress-marks fix that also covers setProgressMarks:.

PR #736, opened 2026-07-23
Fix: set NSArrayController initial defaults and selection index set

DTW-Thalion:
initWithContent: set only selectsInsertedObjects, leaving avoidsEmptySelection, preservesSelection and clearsFilterPredicateOnInsertion at NO. AppKit defaults all three to YES.

It also never created _selection_indexes. With that ivar nil, -selectionIndex returned 0 (from -[nil firstIndex]) instead of NSNotFound, and -selectedObjects (which does objectsAtIndexes: on the arranged objects) raised an out-of-range exception on an empty controller instead of returning an empty array. AppKit reports NSNotFound and an empty array.

Set the three flags and initialise the selection to an empty index set.

Added Tests/gui/NSArrayController/defaults.m covering the three flag defaults, the empty-controller selection index and the empty selected-objects array.

### DIFF
--- a/Source/NSArrayController.m
+++ b/Source/NSArrayController.m
@@ -157,7 +157,11 @@
 {
   if ((self = [super initWithContent: content]) != nil)
     {
+      _selection_indexes = [[NSIndexSet alloc] init];
       [self setSelectsInsertedObjects: YES];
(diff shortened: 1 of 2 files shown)


fredkiefer (approved):
Great finding

PR #748, opened 2026-07-24
Fix: encode instead of decode in NSSplitViewController -encodeWithCoder:

DTW-Thalion:
The keyed-coding branch of -encodeWithCoder: was a copy of -initWithCoder: and called decodeObjectForKey: / decodeFloatForKey: while encoding. So a keyed archive stored none of the split view, split view items or minimum inline sidebar thickness. Worse, -setSplitView: was called with the nil result of decoding, so archiving the controller replaced its split view.

Encode the split view, split view items and minimum thickness under the same keys -initWithCoder: reads.

Added Tests/gui/NSSplitViewController/coding.m checking that archiving does not replace the split view and that the minimum thickness survives a keyed archive round-trip.

### DIFF
--- a/Source/NSSplitViewController.m
+++ b/Source/NSSplitViewController.m
@@ -190,12 +190,10 @@
   [super encodeWithCoder: coder];
   if ([coder allowsKeyedCoding])
     {
-      NSSplitView *sv = [coder decodeObjectForKey: @"NSSplitView"];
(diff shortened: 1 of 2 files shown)


fredkiefer (approved):
Thank you for fixing this. I am terribly sorry that we have code like this in GNUstep gui.

PR #750, opened 2026-07-24
Fix: match NSTitlebarSeparatorStyle values to AppKit

DTW-Thalion:
The NSTitlebarSeparatorStyle enumeration ordered the constants Automatic, Line, None, Shadow (0, 1, 2, 3), but AppKit orders them Automatic, None, Line, Shadow, so NSTitlebarSeparatorStyleLine and NSTitlebarSeparatorStyleNone had swapped values. Swap them to agree with AppKit. The class only stores and returns the value, so nothing else needs to change.

Added Tests/gui/NSSplitViewItem/enumvalues.m checking the four constants.

### DIFF
--- a/Headers/AppKit/NSSplitViewItem.h
+++ b/Headers/AppKit/NSSplitViewItem.h
@@ -54,8 +54,8 @@ typedef NSInteger NSSplitViewItemCollapseBehavior;
 enum
 {
  NSTitlebarSeparatorStyleAutomatic,
- NSTitlebarSeparatorStyleLine,
  NSTitlebarSeparatorStyleNone,
+ NSTitlebarSeparatorStyleLine,
  NSTitlebarSeparatorStyleShadow
 };
 typedef NSInteger NSTitlebarSeparatorStyle;
--- a/Tests/gui/NSSplitViewItem/enumvalues.m
+++ b/Tests/gui/NSSplitViewItem/enumvalues.m
@@ -1,0 +1,25 @@
+/* The NSTitlebarSeparatorStyle values match AppKit: Automatic, None, Line,
+   Shadow are 0, 1, 2, 3. */
+#include "Testing.h"
+
+#include <Foundation/NSAutoreleasePool.h>
+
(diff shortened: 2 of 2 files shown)


Headers/AppKit/NSSplitViewItem.h
```diff
@@ -54,8 +54,8 @@ typedef NSInteger NSSplitViewItemCollapseBehavior;
 enum
 {
  NSTitlebarSeparatorStyleAutomatic,
- NSTitlebarSeparatorStyleLine,
  NSTitlebarSeparatorStyleNone,
+ NSTitlebarSeparatorStyleLine,
```
fredkiefer, line 58: These values get stored in Gorm files, so changing these values might change how existing applications display. I tried to check within Gorm, whether that value is actually settable by a user and didn't find anything. If @gcasa won't object, I will merge this later today, even though there is a slight risk to this change

fredkiefer (approved):

Headers/AppKit/NSSplitViewItem.h
gcasa: I did add the ability to add separators in Gorm.  I think as long as the user builds with the current AppKit, the gorm should be created properly.  I am wondering if we should bump the version of the NSSplitViewItem to compensate for this for older gorm files.

gcasa: The class does not implement non-keyed archiving, which, by definition, means that we don't have any gorm files that could contain this value.   I will implement that (non-keyed Archiving) once this is merged (or, actually, I can do it now and then rebase).

So, with that in mind, go ahead and merge this; it should NOT have any effect on existing .gorm files.

gcasa: @fredkiefer reviewed... sorry about the delay.  Let me know if I have missed any other PRs.

gcasa (approved):
This change looks good to me. Given @fredkiefer 's earlier comment I will take the opportunity to merge.

PR #763, opened 2026-07-24
Fix: NSMutableFontCollection can be mutated after creation

DTW-Thalion:
The mutating methods of NSMutableFontCollection raised and left the collection
unchanged. A mutable collection is built by taking a mutable copy of an
immutable one, but -_setFontCollectionDictionary: stored an immutable copy of
the dictionary, so every later -setObject:forKey: on it raised. In addition
-setQueryDescriptors: and -setExclusionDescriptors: passed the result of
-mutableCopy straight to the dictionary: that raised when given nil (as
-addQueryForDescriptors: does when there are no exclusions yet) and leaked the
copy.

This keeps the collection dictionary mutable, treats a nil descriptor list as an
empty one, and releases the working copies. mutable.m covers setting the query
and exclusion descriptors and adding and removing query descriptors.

### DIFF
--- a/Source/NSFontCollection.m
+++ b/Source/NSFontCollection.m
@@ -288,7 +288,10 @@ static NSLock *_fontCollectionLock = nil;
 
 - (void) _setFontCollectionDictionary: (NSMutableDictionary *)dict
 {
-  ASSIGNCOPY(_fontCollectionDictionary, dict);
+  NSMutableDictionary *d = [dict mutableCopy];
+
+  RELEASE(_fontCollectionDictionary);
+  _fontCollectionDictionary = d;
 }
 
 - (void) _setQueryDescriptors: (NSArray *)queryDescriptors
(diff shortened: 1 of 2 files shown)


Source/NSFontCollection.m
```diff
@@ -288,7 +288,10 @@ - (NSMutableDictionary *) _fontCollectionDictionary
 
 - (void) _setFontCollectionDictionary: (NSMutableDictionary *)dict
 {
-  ASSIGNCOPY(_fontCollectionDictionary, dict);
+  NSMutableDictionary *d = [dict mutableCopy];
```
fredkiefer, line 291: There is an ASSIGNMUTABLECOPY macro. Using that might make things simpler.

fredkiefer (approved):

Source/NSFontCollection.m
DTW-Thalion: Done.

DTW-Thalion: Opened #794 for this, since it merged first.

PR #766, opened 2026-07-24
tests: add NSButton interaction test

DTW-Thalion:
Exercises the click path with `-performClick:`: a momentary button sends its target/action, a toggle button advances its state on each click and keeps sending the action, and a disabled button swallows the click. The set keeps the usual backend skip guard because building the button touches the font/graphics backend.

Checked against AppKit (a click sends one action; a toggle cycles off/on).


fredkiefer (approved):
Another very nice test!

PR #771, opened 2026-07-24
Feature: add the NSSlider convenience constructors

DTW-Thalion:
Adds the AppKit 10.12+ NSSlider factory methods, which GNUstep was missing:

- `+sliderWithTarget:action:` — a slider over the default zero-to-one range
- `+sliderWithValue:minValue:maxValue:target:action:` — a slider with the given range and value

Each wires up the target and action and returns an autoreleased slider. The configuration was checked against AppKit on a macOS runner. Adds `Tests/gui/NSSlider/convenienceConstructors.m`.

### DIFF
--- a/Source/NSSlider.m
+++ b/Source/NSSlider.m
@@ -89,6 +89,39 @@ static Class cellClass;
   return cellClass;
 }
 
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_12, GS_API_LATEST)
++ (instancetype) sliderWithTarget: (id)target
+                           action: (SEL)action
+{
+  NSSlider *slider = AUTORELEASE([[self alloc]
+    initWithFrame: NSMakeRect(0, 0, 100, 20)]);
+
+  [slider setMinValue: 0.0];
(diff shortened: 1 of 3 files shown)


Source/NSSlider.m
```diff
@@ -89,6 +89,39 @@ + (Class) cellClass
   return cellClass;
 }
 
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_12, GS_API_LATEST)
++ (instancetype) sliderWithTarget: (id)target
+                           action: (SEL)action
+{
+  NSSlider *slider = AUTORELEASE([[self alloc]
```
fredkiefer, line 96: Normally we would just call the other convenience method here.

fredkiefer (approved):

Source/NSSlider.m
DTW-Thalion: Done.

DTW-Thalion: Opened #795 for this, since it merged first.

PR #777, opened 2026-07-24
Feature: add NSCell drawWithExpansionFrame and drawFocusRingMask

DTW-Thalion:
NSCell is missing drawWithExpansionFrame:inView: and drawFocusRingMaskWithFrame:inView:, so macOS code that calls or overrides them does not compile against GNUstep.

drawWithExpansionFrame:inView: draws the cell in the expansion frame by calling drawWithFrame:inView: with that frame (the forwarding was verified against AppKit on macOS). drawFocusRingMaskWithFrame:inView: fills the cell frame to form the focus ring mask.

Tests/gui/NSCell/drawingDefaults.m checks the expansion-frame forwarding and renders the focus ring mask to confirm the frame is filled.

### DIFF
--- a/Headers/AppKit/NSCell.h
+++ b/Headers/AppKit/NSCell.h
@@ -675,6 +675,12 @@ APPKIT_EXPORT_CLASS
 - (void) setUsesSingleLineMode: (BOOL)flag;
 - (BOOL) usesSingleLineMode;
 #endif
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_5, GS_API_LATEST)
+- (void) drawWithExpansionFrame: (NSRect)cellFrame inView: (NSView *)view;
+#endif
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_7, GS_API_LATEST)
+- (void) drawFocusRingMaskWithFrame: (NSRect)cellFrame inView: (NSView *)view;
+#endif
 
 //
 // Target and Action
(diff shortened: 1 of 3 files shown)


fredkiefer (approved):
I think that both methods should in some way be related to their corresponding frame methods you added in another PR. They aren't at the moment so this seems wrong. We will need to look into that later on.

DTW-Thalion:
You are right that they should line up. For the focus ring the default drawFocusRingMaskWithFrame: fills the whole cell frame, so focusRingMaskBoundsForFrame: should return that same cell frame rather than NSZeroRect; I can send that as a follow-up.

The expansion pair is different: a plain NSCell never truncates, so expansionFrameWithFrame: returning NSZeroRect is correct for it and only NSTextFieldCell returns a real expansion frame. With no expansion frame, drawWithExpansionFrame: drawing the cell normally matches. Shall I send the focusRingMaskBoundsForFrame: change?

DTW-Thalion:
Correction on the focus ring: I checked -[NSCell focusRingMaskBoundsForFrame:inView:] on macOS and it returns NSZeroRect for a plain NSCell as well as for an NSButton cell and an NSTextFieldCell, with and without a view. So the current NSZeroRect already matches AppKit and returning the cell frame would be a divergence. The empty bounds sitting next to drawFocusRingMaskWithFrame: filling the frame is how AppKit itself behaves for the base cell, so I will leave focusRingMaskBoundsForFrame: as it is.

PR #779, opened 2026-07-24
Fix: implement NSTextView toggleTraditionalCharacterShape:

DTW-Thalion:
toggleTraditionalCharacterShape: was a placeholder that only logged "not implemented".

It now sets NSCharacterShapeAttributeName over the selection, toggling between the traditional character shape (1) and the default shape (0) based on the current value, and updates the typing attributes, following the same pattern as the ligature actions (useAllLigatures: etc).

Verified against AppKit on macOS: from no attribute the toggles produce 1, then 0, then 1. Tests/gui/NSTextView/characterShape.m checks that sequence and that the attribute is applied across the whole selection. Verified passing under the xlib, art and cairo backends.

### DIFF
--- a/Source/NSTextView_actions.m
+++ b/Source/NSTextView_actions.m
@@ -12,7 +12,7 @@
    Author: Felipe A. Rodriguez <far@ix.netcom.com>
    Date: July 1998
 
-   Author: Daniel Böhringer <boehring@biomed.ruhr-uni-bochum.de>
+   Author: Daniel Bï¿½hringer <boehring@biomed.ruhr-uni-bochum.de>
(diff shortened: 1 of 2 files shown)


fredkiefer (approved):
Your change scrambled the name of Daniel Böhringer. He now is a professor at the local university clinic and we had a beer together just the other week. So I had to restore it. :-)

DTW-Thalion:
> Your change scrambled the name of Daniel Böhringer. He now is a professor at the local university clinic and we had a beer together just the other week. So I had to restore it. :-)

That would have been awkward !! :)

PR #780, opened 2026-07-24
Fix: implement NSWindow drawers

DTW-Thalion:
-[NSWindow drawers] was a placeholder that only logged "not implemented" and returned nil.

NSDrawer now records each live drawer in a table held without retaining them (so there is no retain cycle with the drawer, which retains its parent window, and no leak), removing itself on dealloc. -[NSWindow drawers] returns the drawers whose parent window is the receiver.

Verified against AppKit on macOS: the list is a non-nil empty array when there are none, follows setParentWindow: (adding and, with nil, removing), and is per window. Tests/gui/NSDrawer/windowDrawers.m covers those cases, verified passing under the xlib, art and cairo backends.

### DIFF
--- a/Source/NSDrawer.m
+++ b/Source/NSDrawer.m
@@ -33,6 +33,7 @@
 #import <Foundation/NSCoder.h>
 #import <Foundation/NSArchiver.h>
 #import <Foundation/NSKeyedArchiver.h>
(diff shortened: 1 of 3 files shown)


fredkiefer (approved):
This may not be the most efficient implementation but it works, that is a lot more than before.

PR #781, opened 2026-07-24
Fix: NSComboBox resets the selected index when a removal drops it out of range

### LINKED ISSUE #656: NSComboBox: indexOfSelectedItem is not reset when a removal leaves it out of range
With an internal item list (usesDataSource is NO), -indexOfSelectedItem keeps its value when items are added or removed, and it is not clamped to the new item count. If a removal leaves the selected index past the last item, -objectValueOfSelectedItem then does not return.

Steps to reproduce, with a backend present:

```objc
NSComboBox *cb = [[NSComboBox alloc] initWithFrame: NSMakeRect(0, 0, 120, 22)];
[cb addItemWithObjectValue: @"alpha"];
[cb addItemWithObjectValue: @"beta"];
[cb addItemWithObjectValue: @"gamma"];
[cb selectItemAtIndex: 2];       // gamma
[cb removeItemAtIndex: 0];       // list is now: beta, gamma
[cb indexOfSelectedItem];        // 2, but the only valid indexes are 0 and 1
[cb objectValueOfSelectedItem];  // does not return
```

I checked the same operations against AppKit on a macOS runner, starting each time from a three item list (alpha, beta, gamma):

```
select index 2, then insert an item at index 0
    AppKit   indexOfSelectedItem 2   (value is now beta)
    GNUstep  indexOfSelectedItem 2   (value is now beta)

select index 2, then remove the item at index 0
    AppKit   indexOfSelectedItem -1
    GNUstep  indexOfSelectedItem 2   (out of range; objectValueOfSelectedItem does not return)

select index 1, then remove the item at index 1
    AppKit   indexOfSelectedItem 1   (value is now gamma)
    GNUstep  indexOfSelectedItem 1   (value is now gamma)
```

AppKit appears to keep the raw selected index and reset it to -1 only when a removal leaves it outside the list. GNUstep keeps the index but never resets it, so it can end up out of range.

Before sending a patch I would like to confirm the intended behavior. Should -removeItemAtIndex: reset -indexOfSelectedItem to -1 when the selected index would fall outside the list, matching AppKit, or should the selection follow its item as the item moves?

Separately, -itemObjectValueAtIndex: (reached through -objectValueOfSelectedItem) does not return for an out of range index rather than raising NSRangeException. I can add a bounds check for that in the same patch.

I have coverage tests for the current list behavior and can follow up with a patch once the selection behavior is settled.

cc @fredkiefer

DTW-Thalion:
With an internal item list (usesDataSource NO), -removeItemAtIndex:,
-removeItemWithObjectValue: and -removeAllItems left -indexOfSelectedItem
unchanged. When a removal left the selected index past the last item,
-objectValueOfSelectedItem reached -[NSArray objectAtIndex:] out of range and
raised NSRangeException.

Reset the selected index to -1 when a removal leaves it at or beyond the item
count. An insertion, or a removal that leaves the index in range, keeps the raw
index as before. The behaviour was checked against AppKit on a macOS runner:

    select index 2, then remove the item at index 0  -> indexOfSelectedItem -1
    select index 2, then insert an item at index 0   -> indexOfSelectedItem 2
    select index 1, then remove the item at index 1  -> indexOfSelectedItem 1

Tests added in Tests/gui/NSComboBox/selection.m cover insertion, in-range and
out-of-range removal, removal by object value and removeAllItems.

Closes #656

### DIFF
--- a/Source/NSComboBoxCell.m
+++ b/Source/NSComboBoxCell.m
@@ -1033,8 +1033,12 @@ static GSComboWindow *gsWindow = nil;
   else
     {
       [_popUpList removeObject: object];
+      if (_selectedItem >= (NSInteger)[_popUpList count])
+        {
+          _selectedItem = -1;
+        }
     }
-    
+
   [self reloadData];
 }
 
@@ -1053,8 +1057,12 @@ static GSComboWindow *gsWindow = nil;
   else
     {
       [_popUpList removeObjectAtIndex: index];
+      if (_selectedItem >= (NSInteger)[_popUpList count])
+        {
+          _selectedItem = -1;
+        }
     }
-    
+
   [self reloadData];
 }
 
@@ -1073,8 +1081,9 @@ static GSComboWindow *gsWindow = nil;
   else
     {
       [_popUpList removeAllObjects];
+      _selectedItem = -1;
     }
-    
+
   [self reloadData];
 }
 
--- a/Tests/gui/NSComboBox/selection.m
+++ b/Tests/gui/NSComboBox/selection.m
@@ -1,0 +1,108 @@
+/* NSComboBox internal-list selection tracking across insertion and removal
+   (usesDataSource NO).  The raw selected index is kept when an item is inserted
+   or when a removal leaves the index inside the list, and it is reset to -1 when
+   a removal leaves it past the last item.  The relationships were checked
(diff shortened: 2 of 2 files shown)


fredkiefer (approved):
That is quite some surprising behaviour

Issue #700, opened 2026-07-22
Regression: cairo/Xvfb CI aborts in NSApplication teardown; RELEASE(self) in double -init destroys the shared display server (e82a3c5d)


DTW-Thalion:
The cairo/Xvfb CI job has been failing on master just recently. `gui/NSApplication/basic.m` passes every assertion and then aborts at teardown:

```
*** -[XGServer runLoopShouldBlock:]: message sent to deallocated instance 0x...
gnustep-tests: line 465: Aborted (core dumped) ./obj/basic
```

Bisect: the cairo job was green on 943ad622c and red on the next master run at e82a3c5d, with exactly one commit between them, e82a3c5db "Fix tiny leak". Master has been red on the cairo job on every run since.

Cause. That commit changes the double-init guard in `-[NSApplication init]`:

```objc
  if (NSApp != nil)
    {
      RELEASE(self);
      [NSException raise: NSInternalInconsistencyException
                  format: _(@"[NSApplication -init] called more than once")];
    }
```

`NSApplication/basic.m` exercises this path (the "Second -init correctly raises" test). The new `RELEASE(self)` deallocates the second, partially initialized NSApplication instance, and `-[NSApplication dealloc]` unconditionally tears down the shared display server:

```objc
  GSDisplayServer *srv = GSServerForWindow(_app_icon_window); // nil on the bailed instance
  if (srv == nil)
    srv = GSCurrentServer();                                  // -> the real NSApp's server
  ...
  [srv closeServer];
  DESTROY(srv);
```

Because the second instance never got its own server, `srv` resolves to `GSCurrentServer()`, i.e. the server the real `NSApp` is still using. Closing and destroying it leaves `NSApp` with a dangling XGServer, and the process aborts when the run loop next sends `-runLoopShouldBlock:`.

The regression is deterministic in the cairo/Xvfb job (which runs with a real backend); the non-backend jobs skip the window-server path and are unaffected.

Possible directions: guard `-[NSApplication dealloc]` so it only closes and destroys the server when `self == NSApp`, or free the surplus instance in the double-init path without running the full dealloc (e.g. NSDeallocateObject) so it cannot touch shared state.

cc @rfm

rfm:
The abort should already be fixed: because it was due to the XGServer failing to remove itself from the run loop on shutdown, and I fixed that.
[...]

PR #782, opened 2026-07-25
Fix: implement -[NSView backingAlignedRect:options:]

DTW-Thalion:
Implements -[NSView backingAlignedRect:options:]. It delegates the alignment to NSIntegralRectWithOptions: at a backing scale of 1 it calls it directly, and otherwise it scales the rectangle into backing coordinates, aligns there, and scales back.

Depends on libs-base #700, which completes NSIntegralRectWithOptions for all the min/max/size mode combinations. The test checks the windowless (scale 1) case for nearest/inward/outward, per-edge modes, the flipped tie-break, negative origins and an already-integral rectangle, against Apple.

This is the version discussed in #700: it drops the private alignment helper and reuses NSIntegralRectWithOptions.

### DIFF
--- a/Headers/AppKit/NSView.h
+++ b/Headers/AppKit/NSView.h
@@ -311,6 +311,10 @@ PACKAGE_SCOPE
 - (NSSize) convertSizeFromBase: (NSSize)aSize;
 - (NSSize) convertSizeToBase: (NSSize)aSize;
 #endif
+#if OS_API_VERSION(MAC_OS_X_VERSION_10_7, GS_API_LATEST)
+- (NSRect) backingAlignedRect: (NSRect)aRect
+                      options: (NSAlignmentOptions)options;
+#endif
 
 /*
  * Notifying Ancestor Views
--- a/Source/NSView.m
+++ b/Source/NSView.m
@@ -1744,6 +1744,34 @@ static NSSize _computeScale(NSSize fs, NSSize bs)
   return aRect;
 }
 
+- (NSRect) backingAlignedRect: (NSRect)aRect
+                      options: (NSAlignmentOptions)options
+{
+  CGFloat scale = 1.0;
+
+  if (_window != nil)
+    {
+      scale = [_window backingScaleFactor];
+    }
+  if (scale <= 0.0)
+    {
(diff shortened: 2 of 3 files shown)


fredkiefer:
Your code looks very nice, but I am not sure whether it is what MacOS does here. Your test view is not inside a window, so it is a special case and there are no scaling or rotation transformations in the hierarchy.

DTW-Thalion:
You are right that the test view has no window, so the scale is 1 and there is no offset into a backing store. What the code does is align each edge and size to the backing pixel grid using the window's backingScaleFactor, falling back to 1 with no window. For a view positioned on integer points in its window (the usual case) that matches what AppKit returns.

It differs only when the view sits at a fractional position in the window or under a scaled or rotated hierarchy: there the backing grid is offset relative to the view's coordinates, so the rect would first need converting to backing coordinates (as -convertRectToBacking: does) before aligning, then back. GNUstep does not have -convertRectToBacking:/-convertRectFromBacking: yet.

I can add those two methods and route backingAlignedRect: through them so the general case is correct, or keep this change to the common case and add a window-based test. Which would you prefer?

DTW-Thalion:
I rebuilt this to use NSIntegralRectWithOptions and drop the private helper, as we discussed on #700. 

It now depends on #700 being completed and merged in (cc @rfm). See if you like it.

fredkiefer:
As you correctly wrote, we will require -convertRectToBacking:/-convertRectFromBacking: to properly implement this method. So your current implementation will do until then. 
I am ready to merge this but the problem is that the test is currently failing. Not sure whether this is caused by the missing flipped code or by some change to the code you moved to base?

DTW-Thalion:
CI isn't picking up the #700 change for some reason, will fix.

DTW-Thalion:
@fredkiefer before merging in this fix, merge in[ libs-back/PR#175](https://github.com/gnustep/libs-back/pull/175) and [libs-back/PR#176](https://github.com/gnustep/libs-back/pull/176) - two bugs exposed by #782 in these two backends.

fredkiefer (approved):

PR #784, opened 2026-07-25
Fix: load NSDataAsset from the bundle and report a missing asset

### LINKED ISSUE #582: NSDataAsset never loads an asset, and never reports a missing one
NSDataAsset never looks for an asset, and never says that it failed to find one.

-initWithName:bundle: copies the name, keeps the bundle, and sets _data and
_typeIdentifier to nil. Nothing else in the class reads the bundle or fills
those in, so -data and -typeIdentifier are nil for every asset, and the class
cannot succeed at what it is for.

The initialiser also returns an object whatever it is handed. AppKit returns
nil when it cannot find the named asset, checked on a macOS runner:

    [[NSDataAsset alloc] initWithName: @"NoSuchAsset"]            // nil
    [[NSDataAsset alloc] initWithName: @"NoSuchAsset"
                               bundle: [NSBundle mainBundle]]     // nil

so a caller written against AppKit checks for nil, gets an object, reads -data,
and gets nil with no indication of why.

I have not sent a patch because the two halves want deciding together. Returning
nil from the initialiser would match AppKit and tell callers the truth, but
until the lookup exists it would mean the class always returns nil, and I did
not want to make that call from outside. Reading the asset needs somewhere to
read it from: assets live in a compiled asset catalog on macOS, and there is no
equivalent here yet, so a decision about the on-disk shape comes first.

Happy to send a patch for whichever of those you would like, @fredkiefer

DTW-Thalion:
NSDataAsset kept its name and bundle but never read anything, so -data and -typeIdentifier were nil for every asset, and -initWithName:(bundle:) returned an object even for an asset that does not exist (AppKit returns nil, checked on a macOS runner).

The initialiser now loads the named asset from the bundle:

- an Apple .dataset directory (<name>.dataset), reading its Contents.json to find the data file, or
- a plain named resource in the bundle.

The type identifier is derived from the data file's extension for the common uniform types. The initialiser returns nil when the asset cannot be found or no name is given; a nil bundle uses the main bundle.

Since there is no compiled asset catalog on GNUstep, the loading is exercised in Tests/gui/NSDataAsset/loading.m against a temporary bundle built with both a .dataset directory and a plain resource; five of its assertions fail against the previous implementation. The .dataset layout follows Apple's source format so an .xcassets shipped as bundle resources resolves; the plain-resource path covers the simpler GNUstep case.

Closes #582

### DIFF
--- a/Source/NSDataAsset.m
+++ b/Source/NSDataAsset.m
@@ -22,30 +22,192 @@
    Boston, MA 02110 USA.
 */
 
+#import <Foundation/NSArray.h>
 #import <Foundation/NSBundle.h>
-#import <Foundation/NSString.h>
 #import <Foundation/NSData.h>
+#import <Foundation/NSDictionary.h>
+#import <Foundation/NSEnumerator.h>
+#import <Foundation/NSJSONSerialization.h>
+#import <Foundation/NSPathUtilities.h>
+#import <Foundation/NSString.h>
 #import <AppKit/NSDataAsset.h>
+
+@interface NSDataAsset (Private)
+- (BOOL) _loadData;
+@end
+
+/* A best-effort type identifier from a file extension.  Only the common,
+   standard uniform type identifiers are mapped; an unknown extension gives
+   nil. */
+static NSString *
+typeIdentifierForExtension(NSString *ext)
+{
+  static NSDictionary *map = nil;
+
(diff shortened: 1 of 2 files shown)


fredkiefer (approved):
Looks very nice

Issue #346, opened 2025-04-23
exception due to bad PPD whilesetting up page


rmottola:
I noticed a strange behaviour with Ink (happens on Solaris/sparc64 but I think it is just a coincidence) where I get a "white menu" instead of opening a page layout or print. In Graphos it works. From debugging, an exception is happening, not caught and this is due to invalid PPD file:

```
#0  -[NSException raise] (self=0x2f0c98, _cmd=0xfe961758 <_OBJC_SELECTOR_TABLE+240>) at NSException.m:1602
#1  0xfe5e6f00 in +[NSException raise:format:arguments:] (self=0xfe9618c4 <_OBJC_Class_NSException>, 
    _cmd=<optimized out>, name=0xff1a6cf8 <_OBJC_INSTANCE_23>, format=0xff180438 <_OBJC_INSTANCE_29>, 
    argList=<optimized out>) at NSException.m:1494
#2  0xfe5e7c90 in +[NSException raise:format:] (self=0xfe9618c4 <_OBJC_Class_NSException>, 
    _cmd=0xff17fd44 <_OBJC_SELECTOR_TABLE+512>, name=0xff1a6cf8 <_OBJC_INSTANCE_23>, 
    format=0xff180438 <_OBJC_INSTANCE_29>) at NSException.m:1479
#3  0xfee9f3f8 in -[NSPrinter(PPDParsingPrivate) loadPPDAtPath:symbolValues:inclusionSet:] (self=self@entry=0x40f7c8, 
    _cmd=_cmd@entry=0xff17fd24 <_OBJC_SELECTOR_TABLE+480>, ppdPath=ppdPath@entry=0x180558, 
    ppdSymbolValues=ppdSymbolValues@entry=0x3fabf0, inclusionSet=0x4162d8) at NSPrinter.m:917
#4  0xfee9eab4 in -[NSPrinter(PPDParsing) parsePPDAtPath:] (self=self@entry=0x40f7c8, 
    _cmd=_cmd@entry=0xfad75424 <_OBJC_SELECTOR_TABLE+72>, ppdPath=0x180558) at NSPrinter.m:790
#5  0xfad636a4 in +[GSLPRPrinter printerWithName:] (self=<optimized out>, _cmd=<optimized out>, name=<optimized out>)
    at GSLPRPrinter.m:100
#6  0xfee9d738 in +[NSPrinter printerWithName:] (self=<optimized out>, _cmd=<optimized out>, name=0x17f270)
[...]

PR #788, opened 2026-07-25
Fix: do not abort when a stored printer has a missing PPD

DTW-Thalion:
A printer stored under the `GSLPRPrinters` default can reference a PPD file that has moved or been removed (for example an old install path). GSLPRPrinter loads it with `-parsePPDAtPath:`, which raises `NSPPDParseException` for an unreadable file. Nothing catches it, so the application aborts while only setting up printing, e.g. `+[NSPrintInfo sharedPrintInfo]` -> `+[NSPrintInfo defaultPrinter]` -> `+[GSLPRPrinter printerWithName:]` -> `-parsePPDAtPath:` (as reported in #346 with Ink).

+[GSLPRPrinter printerWithName:] now catches the exception, logs a warning, and returns the printer without the PPD. `-[NSPrintInfo initWithDictionary:]` already falls back to its defaults (A4) when the printer provides no PPD values, so setup completes instead of aborting.

Test: Tests/gui/NSPrinter/badPPD.m sets a `GSLPRPrinters` entry with a non-existent PPD path and loads it; two assertions fail against the previous code (it raises and returns nil). Verified with the GSLPR bundle rebuilt and loaded from the user domain.

Note: GSCUPSPrinter loads a PPD the same way and has the same exposure. I left it out of this change since I could not exercise the CUPS path here; happy to follow up if you would like it covered too.

Addresses the first item in #346 (missing PPD on open should not abort). The print/page-setup path and whether per-app PPD storage is needed are separate items in that issue.

### DIFF
--- a/Printing/GSLPR/GSLPRPrinter.m
+++ b/Printing/GSLPR/GSLPRPrinter.m
@@ -97,8 +97,20 @@
                         withHost: [printerEntry objectForKey: @"Host"]
                         withNote: [printerEntry objectForKey: @"Note"]];
 
-  [printer parsePPDAtPath: [printerEntry objectForKey: @"PPDPath"]];
-                         
+  /* A stored printer may reference a PPD that no longer exists; an unreadable
+     PPD is a warning here and the printer is returned without it, so the caller
+     uses its defaults. */
+  NS_DURING
+    {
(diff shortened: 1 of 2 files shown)


fredkiefer (approved):
Yes, a similar solution for CUPS would be great.

DTW-Thalion:
The CUPS version is #789, which is now merged.

PR #790, opened 2026-07-25
Fix: autohidesScrollers respects hasHorizontal/VerticalScroller

### LINKED ISSUE #236: hasHorizontalScroller = NO; ignored if autohidesScrollers = YES;
When 'autohidesScrollers = YES;' the horizontal scroll is shown even if 'hasHorizontalScroller = NO;'

DTW-Thalion:
With `autohidesScrollers` set to YES, an axis configured to have no scroller (`setHasHorizontalScroller: NO`) still got one when its content overflowed, because `-_autohideScrollers` decided visibility purely from the content size. NSScrollView tracked a single flag per axis (`_hasHorizScroller`/`_hasVertScroller`) that served as both the configured setting and the current visibility, so the autohide pass overwrote the client setting.

This tracks the configured setting separately (`_hasHorizScrollerRequested`/`_hasVertScrollerRequested`), set from the public setters and when decoding, and gates the autohide pass by it: a scroller is only shown on an axis the client asked to scroll. The public setters keep their behaviour through an internal visibility setter that the autohide pass now uses, so it no longer disturbs the configured setting.

**ABI note:** this appends two BOOL ivars to NSScrollView, changing its ivar layout, in the same way as other recent state fixes. The library is built as a whole and the ivars are at the end of the block. Say if you would rather handle the state another way.

Test: Tests/gui/NSScrollView/autohidesScrollers.m is extended (it previously asserted the buggy behaviour, that a disabled scroller was shown). Two new assertions fail against the previous code: a horizontal scroller set to NO stays hidden for a wide document, and symmetrically for a disabled vertical scroller. The recursion regression from #234 and the two-axis coupling cases still pass.

Behaviour was checked against AppKit on a macOS runner. Note one further AppKit difference I did **not** change here: `-hasHorizontalScroller` on macOS returns the configured value, and autohide instead hides the scroller (`-[NSScroller isHidden]`) while keeping the property YES; GNUstep's getter returns the current visibility. Matching that would also touch the theme drawing (GSThemeDrawing reads the getter to lay out the scroller area), so I left the getter as-is and kept this change to the reported bug. Happy to follow up on the getter semantics if you want it.

Closes #236

### DIFF
--- a/Headers/AppKit/NSScrollView.h
+++ b/Headers/AppKit/NSScrollView.h
@@ -79,6 +79,8 @@ APPKIT_EXPORT_CLASS
   NSScrollElasticity _vertScrollElasticity;
   NSScrollerStyle _scrollerStyle;
   NSScrollerKnobStyle _scrollerKnobStyle;
+  BOOL _hasHorizScrollerRequested;
+  BOOL _hasVertScrollerRequested;
 }
 
 /* Calculating layout */
--- a/Source/NSScrollView.m
+++ b/Source/NSScrollView.m
@@ -107,6 +107,8 @@ typedef struct _scrollViewFlags
 - (void) _synchronizeHeaderAndCornerView;
 - (void) _themeDidActivate: (NSNotification*)notification;
 - (void) _autohideScrollers;
+- (void) _setHasHorizScroller: (BOOL)flag;
+- (void) _setHasVertScroller: (BOOL)flag;
 @end
 
 @implementation NSScrollView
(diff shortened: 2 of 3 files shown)


fredkiefer (approved):

fredkiefer:
Most likely it is best to also change the getter to be similar to MacOs.

DTW-Thalion:
Opened #796 to make the getters report the configured value, since this merged first.

PR #794, opened 2026-07-25
Use ASSIGNMUTABLECOPY in -_setFontCollectionDictionary:

DTW-Thalion:
Follow-up to #763, which was merged before I could apply the review comment. -_setFontCollectionDictionary: did the mutable copy, release and assign by hand; ASSIGNMUTABLECOPY does the same.

### DIFF
--- a/Source/NSFontCollection.m
+++ b/Source/NSFontCollection.m
@@ -288,10 +288,7 @@ static NSLock *_fontCollectionLock = nil;
 
 - (void) _setFontCollectionDictionary: (NSMutableDictionary *)dict
 {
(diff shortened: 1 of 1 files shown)


fredkiefer (approved):
You don't have to change everything I comment on. I just wanted to share the more elegant solution.

DTW-Thalion:
> You don't have to change everything I comment on. I just wanted to share the more elegant solution.

I like elegant solutions ! :) Plus with all the tests and fixes, I am bound to make some mistakes or miss some of these things - so I rather address the feedback as quickly as I can.

PR #796, opened 2026-07-25
Return the configured value from -hasHorizontalScroller/-hasVerticalScroller

DTW-Thalion:
Follow-up to #790. As on macOS, -hasHorizontalScroller and -hasVerticalScroller now report the configured setting rather than the current autohidden visibility, returning the requested flags that #790 added. The autohidesScrollers test checked the getters for visibility, so it now checks whether each scroller is in the view hierarchy instead, and gains a case that the getter still reports the configured value while the scroller is autohidden.


fredkiefer (approved):
We only see if this causes issues in applications after we merge it. So let's try.

PR #820, opened 2026-07-26
Fix: ignore paragraph first line head indent in string drawing

### LINKED ISSUE #350: AttributedString boundingRectWithSize vs drawInRect
in HelpViewer NoteCells are drawn in attachments.

First, the cell is sized and using pseudo-infinity height, the best width is determined:

```
	size.width -= leadingMargin*2;
	size.height = 9999.0;
	if (size.width <= 0)
	  size.width = [textView bounds].size.width;
        noteSize = [_note boundingRectWithSize:size options:0].size;
```

then, later, it is drawn:
```
      noteRect.origin = noteOrigin;
      noteRect.size = noteSize;
      [_note drawInRect: noteRect];
```

amazingly, this fails for certain strings. The issue was found with single-line strings where the string gets just truncated.

At a first glance, boundingRectWithSize returns a decent-wide enough size, it looks that drawInRect is not following it.
Also, the error doesn't seem constant.

Attributed String:
```
<b>{NSFont = "<NSFont: 0x562750b36720> DejaVuSansMono 12.000 0.000 0.000 12.000 0.000 0.000 P 0"; }votre texte en gras{NSFont = "<NSFont: 0x562750a8a120> TrebuchetMS-Bold 12.000 0.000 0.000 12.000 0.000 0.000 P 0"; }</b>{NSFont = "<NSFont: 0x562750b36720> DejaVuSansMono 12.000 0.000 0.000 12.000 0.000 0.000 P 0"; }
```
Size:
```
resizeWithTextView size: {width = 336; height = 9999} noteSize: {width = 158; height = 15}
```

Debug shows us reasonable text with a width of 158, yet it clips:

Image

empirically, I determined the error to be more than 10, less or equal to 15. Adding 15 to the width, makes the string (and others) draw. I suppose this is due to even off-by-one character causing a newline which is then clipped, since the height remained one-line.

DTW-Thalion:
-boundingRectWithSize:options: returned the paragraph first line head indent in its origin while -drawInRect: drew the glyphs at the indent, so the reported width did not cover the drawn text and the tail was clipped. This is the HelpViewer NoteCell case in the issue.

On macOS these convenience methods ignore the head indent: the origin stays at zero and the width stays the text width, so the reported size matches what is drawn. This removes the first line head indent from the private working copy before layout, so -boundingRectWithSize:options: and -drawInRect: agree and match AppKit.

Adds Tests/gui/NSAttributedString/headIndentBounding.m.

Closes #350.

### DIFF
--- a/Source/NSStringDrawing.m
+++ b/Source/NSStringDrawing.m
@@ -36,7 +36,9 @@
 #import <Foundation/NSLock.h>
 
 #import "AppKit/NSAffineTransform.h"
+#import "AppKit/NSAttributedString.h"
 #import "AppKit/NSLayoutManager.h"
+#import "AppKit/NSParagraphStyle.h"
 #import "AppKit/NSStringDrawing.h"
 #import "AppKit/NSTextContainer.h"
 #import "AppKit/NSTextStorage.h"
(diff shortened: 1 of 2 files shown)


Source/NSStringDrawing.m
```diff
@@ -308,6 +348,7 @@ static inline void prepare_string(NSString *string, NSDictionary *attributes)
 			  range: NSMakeRange(0, [string length])];
     }
   [scratchTextStorage endEditing];
+  strip_first_line_indent(scratchTextStorage);
```
fredkiefer, line 351: in this specific case it would be faster to just drop the indent from `attributes`.

fredkiefer (approved):
Thank you

DTW-Thalion:
Took your suggestion to drop the indent from the attributes for the plain string path in #821.

rmottola:
HelpViewer notes look fine now and like Apple, No further testing done though.

PR #821, opened 2026-07-27
Fix: NSStringDrawing - drop the head indent from the attributes for plain strings

DTW-Thalion:
Follow-up to #820. In the review there, @fredkiefer noted that for the plain string path it would be faster to drop the head indent from the attributes rather than rewrite the laid out text.

`prepare_string` applies a single attributes dictionary to the whole string, so this drops the first line head indent from a copy of that dictionary before it is applied, instead of calling `strip_first_line_indent` afterwards. The attributed string path still uses `strip_first_line_indent`, where the paragraph runs can vary.

No behaviour change: the plain string `boundingRectWithSize:options:attributes:` still reports origin x 0 with a head indent, and the NSAttributedString tests (including `headIndentBounding`) pass.

### DIFF
--- a/Source/NSStringDrawing.m
+++ b/Source/NSStringDrawing.m
@@ -344,11 +344,33 @@ static inline void prepare_string(NSString *string, NSDictionary *attributes)
                                     withString: string];
   if ([string length])
     {
-      [scratchTextStorage setAttributes: attributes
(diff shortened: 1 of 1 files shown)


fredkiefer (commented):
Thank you

Issue #756, opened 2026-07-24
NSCollectionViewItem: -highlightState missing and -draggingImageComponents returns an empty array


DTW-Thalion:
Two AppKit features are absent from NSCollectionViewItem:

- No -highlightState / -setHighlightState: and no NSCollectionViewItemHighlightState enum (AppKit default NSCollectionViewItemHighlightNone = 0).
- -draggingImageComponents returns an empty array with a FIXME ("We don't have NSDraggingImageComponent"); AppKit returns one component (the item view). This needs the NSDraggingImageComponent class.

Everything else on NSCollectionViewItem matches AppKit. Filing these as feature gaps.

PR #823, opened 2026-07-27
Fix: store the NSCollectionViewItem highlight state

DTW-Thalion:
NSCollectionViewItem did not implement -highlightState, so an item could not report the highlight the collection view assigns it during selection tracking or a drag (the enum NSCollectionViewItemHighlightState already exists in NSCollectionView.h).

Add -highlightState/-setHighlightState: backed by an ivar. Checked on a macOS runner: the default is NSCollectionViewItemHighlightNone, any of the four states round-trips, and the highlight state is independent of the selection state (setting one does not change the other).

The -draggingImageComponents FIXME is left as is; it still returns an empty array because there is no NSDraggingImageComponent class yet.

The existing NSCollectionViewItem/config.m test gains the highlight-state coverage.

This is the -highlightState half of #756. The -draggingImageComponents half is left for when NSDraggingImageComponent exists, so I have not marked the issue closed here.

### DIFF
--- a/Headers/AppKit/NSCollectionViewItem.h
+++ b/Headers/AppKit/NSCollectionViewItem.h
@@ -32,6 +32,7 @@
 
 #import <Foundation/NSArray.h>
 #import <AppKit/NSViewController.h>
+#import <AppKit/NSCollectionView.h>
 
 #if OS_API_VERSION(MAC_OS_X_VERSION_10_5, GS_API_LATEST)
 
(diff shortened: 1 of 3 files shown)


DTW-Thalion:
The -draggingImageComponents half of #756 is in #824, which adds NSDraggingImageComponent.

fredkiefer (approved):

PR #826, opened 2026-07-27
Fix: report an unbounded visibleRect for a view with no window

### LINKED ISSUE #613: NSView -visibleRect returns the view bounds without a window
-[NSView visibleRect] on a view that has no window returns the view's own
bounds. AppKit returns an effectively unbounded rect for a windowless view.

GNUstep's offscreen drawing paths depend on the current bounds return, so this
is noted as a known difference rather than a proposed change.

DTW-Thalion:
-[NSView visibleRect] returned the view's own bounds for a view that is not in a window, where AppKit reports an unbounded rectangle. This returns that rectangle for a windowless, non-hidden view to match (checked on a macOS runner: origin {-CGFLOAT_MAX/2, -CGFLOAT_MAX/2}, size {CGFLOAT_MAX, CGFLOAT_MAX}, for the view and for a windowless view nested in a windowless superview). A hidden view still reports NSZeroRect.

The coordinate and drawing machinery needs a bounded rectangle even without a window - the superview intersection in -_rebuildCoordinates would otherwise convert an unbounded rectangle and overflow - so that code, along with the lock-focus and display entry points, now uses a new private -_geometricVisibleRect that keeps the previous clipped-bounds behaviour. Only the public accessor changes, so a view in a window is unaffected (the full NSView/NSScrollView/NSClipView/NSSplitView/NSTableView/NSTextView suites pass unchanged).

CGFLOAT_MAX is defined locally and gated with #ifndef so the library does not depend on CoreFoundation for it.

New Tests/gui/NSView/visibleRectNoWindow.m covers the windowless cases.

Closes #613

### DIFF
--- a/Source/NSView.m
+++ b/Source/NSView.m
@@ -111,12 +111,28 @@ NSView *viewIsPrinting = nil;
 const CGFloat NSViewNoInstrinsicMetric = -1;
 const CGFloat NSViewNoIntrinsicMetric = -1;
 
+/* Largest representable CGFloat.  Defined here so the library does not pull in
+   CoreFoundation for it; gated so it does not clash where CoreFoundation is
+   present. */
+#ifndef CGFLOAT_MAX
+#  include <float.h>
+#  if defined(__LP64__) && __LP64__
+#    define CGFLOAT_MAX DBL_MAX
+#  else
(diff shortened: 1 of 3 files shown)


fredkiefer (approved):
Interesting finding

Issue #431, opened 2026-05-19
Themes does not completely Propagate to runnings apps and does it a bit slowly


rmottola:
Using SystemPreferences Themes Module
_Select a Theme. Activate theme._

**Expected**: selected theme is immediately applied to SystemPreferences and other running apps.
**Actual**: does not happen. However running  `--GNU-Debug=GSTheme`

```

2026-05-19 22:26:17.071 SystemPreferences[185129:185129] File GSTheme.m: 789. In [GSTheme -deactivate] Tango 0x55a13b12d9c0
2026-05-19 22:26:17.071 SystemPreferences[185129:185129] File GSTheme.m: 554. In [GSTheme -activate] Neos 0x55a13b363f90
```

Shows that the theme receives the activate message.

rmottola:
I suppose this is the root issue of Thematic not updating images - GSTheme does not update on activation.

fredkiefer:
Sorry I am not an expert on GNUstep themes. I am not even sure what to expect when switching a theme. Does this trigger a redraw of the whole application? And if so, does this work for colours and other other drawing elements but won't work for images? Or are all elements not working? For your second comment and our discussion yesterday I think only images are affected. So this behaviour might be caused by a cache for images.

rmottola:
Changing a theme in SystemPreferences should propagate to all running applications. It used to work, now it doesn't. Not only images, but also colours nor the name update in running apps.

The Thematic and icons thing is possibly a consequence of this (just a guess) since it re-activates a theme.

rmottola:
Correction for SystemPreferences

1) Apply is just a "preview" inside the running application, system preferences. It does work
2) Save take time, within 30 seconds, to update and it works. It slow, but works.

However there are some bugs when  an application reloads, e.g. if a theme does not override a colour, the old value remains. If one sets GNUstep theme it works and then setting the other theme fixes it.

rmottola:
https://github.com/gnustep/libs-gui/commit/b6eed598ad849c2dce6824a9e9df686300e10100 possibily fixed this issue. Need to be tested

DTW-Thalion:
@rmottola @fredkiefer - think I may have solved part of this - PR #827

rmottola:
I think no further fix are needed.

SystemPreferences had an issue setting the theme: it should first synchronise defaults before sending notifications. That fixed for me the "slowness".
The issue with Thematic not reloading part of the theme was instead fixed  by Richard with invalidating the cache.

[...]

PR #827, opened 2026-07-27
Fix: reload the theme when the preference change notification arrives

DTW-Thalion:
This addresses the slow, incomplete propagation part of #431.

+preferenceDidChange: is the handler for the distributed GSThemePreferenceDidChangeNotification that SystemPreferences posts when the user changes the theme. It only synchronized the defaults and did nothing else, so a running application did not actually switch theme until the local NSUserDefaultsDidChangeNotification happened to fire. That is what made the change take up to half a minute to show up in other applications, and it never showed up at all for an application still on the default theme (whose local defaults observer is not registered, since +setTheme: only registers it when the theme actually changes).

The handler now calls +defaultsDidChange: after synchronizing, so the new theme is loaded and activated as soon as the notification arrives. +defaultsDidChange: already compares against the current theme name, so it is a no-op when nothing changed, and calling it directly does not depend on the local observer being present.

The two unconditional NSLog calls in the handler are moved to the GSTheme debug channel (visible with --GNU-Debug=GSTheme), which is how this was diagnosed in the issue.

The image caching part was already handled in b6eed598, and the colour path in -[NSColor themeDidActivate:] already fills any colour the new theme does not override from the default system colours, so a switch reverts non-overridden colours.

I have not added a test: exercising this needs a second theme bundle and a second process posting the distributed notification, and it mutates the global active theme and the on-disk GSTheme default, which is not suitable for the headless suite. cc @rmottola 

Refs #431

### DIFF
--- a/Source/GSTheme.m
+++ b/Source/GSTheme.m
@@ -518,11 +518,11 @@ typedef	struct {
 {
   NSUserDefaults	*defaults;
 
-  NSLog(@"System preference changing");
   defaults = [NSUserDefaults standardUserDefaults];
   [defaults synchronize];
-  NSLog(@"System preference changed; default theme is %@",
+  NSDebugMLLog(@"GSTheme", @"default theme is now %@",
     [defaults objectForKey: @"GSTheme"]);
+  [self defaultsDidChange: nil];
 }
 
 + (void) setTheme: (GSTheme*)theme


rfm:
The effect of -synchronise is to cause an NSUserDefaultsDidChangeNotification to be posted immediately after the defaults database was read, so I think that part of the change is redundant.  The logging change looks like an improvement though.

rmottola:
I fixed SystemPreferences to sync before sending the notification.

[https://github.com/gnustep/apps-systempreferences/commit/39956a22804c0e1c59b8ff15ac4a9661ad0867cf](https://github.com/gnustep/libs-gui/issues/url)

fredkiefer (approved):
I still think this is an improvement

rmottola:
we can agree to disagree. I think it is useless/redundant. I hope it doesn't do harm except some CPU and disk waste.

PR #828, opened 2026-07-27
Feature: grow the typesetter glyph cache geometrically rather than in fixed steps

### LINKED ISSUE #408: GSHorizontalTypesetter improvements: Optimize Cache Growth Strategy

### Optimize Cache Growth Strategy
Current `CACHE_STEP` is fixed at 192. Use exponential growth to reduce reallocations:

```objc
// CURRENT:
#define CACHE_STEP 192
cacheSize = newLength;
glyphCache = realloc(glyphCache, sizeof(GlyphCacheEntry) * cacheSize);

// PROPOSED:
if (cacheSize < newLength) {
    cacheSize = MAX(newLength, cacheSize * 2);  // Exponential growth
    cacheSize = MAX(cacheSize, 64);             // Minimum allocation
    glyphCache = realloc(glyphCache, sizeof(GlyphCacheEntry) * cacheSize);
}
```

DTW-Thalion:
-_cacheGlyphsUpToLength: grew the glyph cache to exactly the requested length, and the callers extend it in CACHE_STEP (192) glyph increments, so laying out a long run reallocated the cache roughly every 192 glyphs.

Following the suggestion on the issue to prefer a larger initial cache with a slower rate of growth (rather than the simple doubling first proposed), this starts from a larger initial block and grows the allocation by half again each time it is too small, so a long run reallocates only a handful of times. The cache still fills in CACHE_STEP increments and the resulting layout is unchanged.

New Tests/gui/GSHorizontalTypesetter/longRun.m lays out a run much longer than the initial cache so the growth path runs several times, and checks every glyph is still laid out and the run flows onto later lines.

Closes #408

### DIFF
--- a/Source/GSHorizontalTypesetter.m
+++ b/Source/GSHorizontalTypesetter.m
@@ -96,7 +96,7 @@ cache fairly aggressively without having to worry about memory consumption.
   return shared;
 }
 
-#define CACHE_INITIAL 192
+#define CACHE_INITIAL 512
 #define CACHE_STEP 192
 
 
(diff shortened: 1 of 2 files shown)


fredkiefer (approved):
I don't think this is a real improvement, but I don't see any harm in trying this as we only may overshot by half.

DTW-Thalion:
I tend to agree - but limited downside, so I gave it a shot.

PR #829, opened 2026-07-27
Feature: implement the truncating line break modes in the typesetter

### LINKED ISSUE #414: GSHorizontalTypesetter improvements: Implement Truncation Modes Properly

### Implement Truncation Modes Properly
The current code treats all truncation modes as clipping. Implement proper head/middle/tail truncation:

```objc
case NSLineBreakByTruncatingTail: {
    // Calculate if truncation is needed
    if (position.x + glyphEntry->size.width > lineFragment->rect.size.width) {
        // Find truncation glyph (ellipsis) and fit it
        NSGlyph ellipsisGlyph = [font glyphWithName: @"ellipsis"];
        CGFloat ellipsisWidth = [font advancementForGlyph: ellipsisGlyph].width;
        
        // Remove glyphs until ellipsis fits
        while (index > firstGlyphIndex && 
               glyphEntry[-1].position.x + ellipsisWidth > lineFragment->rect.size.width) {
            index--;
            glyphEntry--;
        }
        
        // Insert ellipsis
        glyphEntry->glyph = ellipsisGlyph;
        glyphEntry->position = position;
        glyphEntry->size.width = ellipsisWidth;
        glyphEntry->dontShow = NO;
        
        lineFragment->lastGlyphIndex = index + 1;
        *newParagraph = NO;
        return NO;  // Line is complete
    }
    break;
}
```

DTW-Thalion:
NSLineBreakByTruncatingTail, ...Head and ...Middle all fell through to clipping: the glyphs that did not fit were hidden and no ellipsis was drawn, and head and middle kept the wrong glyphs.

This adds a real truncation indicator. A line fragment can now carry a truncation glyph (an ellipsis) with a position and font:

- GSLayoutManager stores it as transient layout information on the line fragment (in linefrag_t), set by the typesetter through -setTruncationGlyph:atPoint:font:forGlyphRange: and read back with -truncationGlyphForGlyphAtIndex:. Because it is layout information and not written into the glyph store, a re-layout regenerates it correctly and the glyph/character mapping is untouched.
- NSLayoutManager draws the truncation glyph after the fragment's glyphs.
- GSHorizontalTypesetter handles the three modes: tail drops the trailing glyphs and places the ellipsis at the end; head shows a trailing suffix after a leading ellipsis; middle keeps a leading prefix and a trailing suffix with the ellipsis between them. Dropped glyphs are marked not-shown and the shown suffix is repositioned.

The ellipsis glyph comes from the font, and if the font has none the line is clipped as before.

Tests: Tests/gui/GSHorizontalTypesetter/truncation.m checks, for each mode, which glyphs are shown or hidden and that the ellipsis is set. Layout was verified against the full text-layout suites with no regressions, and drawing of all three modes was checked against the X11 backend under Xvfb.

Closes #414

### DIFF
--- a/Headers/Additions/GNUstepGUI/GSLayoutManager.h
+++ b/Headers/Additions/GNUstepGUI/GSLayoutManager.h
@@ -323,8 +323,20 @@ invalidation.)
 - (void) setLocation: (NSPoint)location 
 	forStartOfGlyphRange: (NSRange)glyphRange;
 
-- (void) setAttachmentSize: (NSSize)attachmentSize 
+- (void) setAttachmentSize: (NSSize)attachmentSize
 	forGlyphRange: (NSRange)glyphRange; /* not OPENSTEP */
+
+/* Set a truncation glyph (typically an ellipsis) to be drawn at the given
+   point in the line fragment that contains glyphRange, indicating that a line
+   break mode dropped glyphs to fit.  A glyph of 0 clears it.  not OPENSTEP */
+- (void) setTruncationGlyph: (NSGlyph)glyph
+		    atPoint: (NSPoint)location
+		       font: (NSFont *)font
+	      forGlyphRange: (NSRange)glyphRange;
+
+/* The truncation glyph of the line fragment that contains the given glyph, or
+   NSNullGlyph if that fragment is not truncated.  not OPENSTEP */
+- (NSGlyph) truncationGlyphForGlyphAtIndex: (NSUInteger)glyphIndex;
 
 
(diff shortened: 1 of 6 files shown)


rmottola:
I'm a little bit weary because we were working on this code to modularize it and fix the long-standing bug of linew-break hadnling (can't find it anymore right now).

However it looks good and instead of reworking it later, I'd accept this and work with it. We don't know yet the size of the fix.

Current branch is: https://github.com/gnustep/libs-gui/commits/refactor_GSHorizontalTypesetter/

PR #830, opened 2026-07-27
fix missing </example> end tag in doc

rmottola:


rfm (approved):
Thanks, simple change in comment.

PR #832, opened 2026-07-27
Fix: keep the paragraph head indent when drawing wrapped strings

### LINKED ISSUE #831: NSStringDrawing: wrapped continuation lines do not align with the first line when a paragraph has a head indent
Follow-up to #350, split off as suggested there.

After #820 a single indented line matches AppKit, but a paragraph that wraps does not. The first line is drawn flush to the left while the continuation lines keep the head indent, so the block is misaligned (see the GNUstep screenshot on #350: check the first line against the second in each box). On macOS every line of the paragraph is drawn at the head indent, so the paragraph reads as one block with a common left edge.

#820 dropped the first line head indent from the working copy of the text before layout, to keep the convenience bounding rect and the drawing in agreement. That removes the indent from the first line only; the continuation lines still use the head indent, which is where the misalignment comes from.

Checking on macOS with -drawWithRect:options: and -boundingRectWithSize:options: shows AppKit draws every line at the indent and keeps the reported rect origin at zero: the indent narrows the wrapped column and shifts the drawn glyphs, but it is not folded into the reported origin. Matching that means keeping the indent in the drawn text and clearing the indent offset from the reported origin, rather than stripping the indent before layout.

DTW-Thalion:
Closes #831. Follow-up to #350 and #820.

#820 made a single line with a paragraph head indent match AppKit, but a wrapped paragraph did not: the first line was drawn flush while the continuation lines kept the head indent, so the block was misaligned (visible in the GNUstep screenshot on #350).

This keeps the head indent in the laid out text, so every line is drawn at the indent, one aligned block as AppKit does. The indent offset is cleared from the origin reported by -boundingRectWithSize:options: instead of stripping the indent before layout. The reported origin stays at zero, so the single line result from #820 is unchanged.

Checked against AppKit with -drawWithRect:options: and -boundingRectWithSize:options:: every line is drawn at the indent, the reported origin stays at zero, and the indent narrows the wrapped column rather than being folded into the origin.

Tests. NSAttributedString/headIndentWrapped.m checks that an indented wrapped paragraph reports the same size as the same text in a container narrowed by the indent (the indent applies to every line) and keeps the bounding rect origin at zero. NSAttributedString/headIndentDrawing.m renders a wrapped indented paragraph and checks that the first and second lines start at the same column and that the block is drawn at the indent; it skips without a window server.

### DIFF
--- a/Source/NSStringDrawing.m
+++ b/Source/NSStringDrawing.m
@@ -296,44 +296,6 @@ static cache_t *cache_lookup(BOOL hasSize, NSSize size, BOOL useScreenFonts)
   return c;
 }
 
-/* AppKit ignores a paragraph first line head indent in the convenience string
-   drawing methods: the text is laid out, measured and drawn from the origin,
-   not from the indent. Drop the indent from our private working copy so that
-   -boundingRectWithSize:options: and -drawInRect: agree and match AppKit. */
-static void strip_first_line_indent(NSTextStorage *textStorage)
-{
-  NSUInteger length = [textStorage length];
-  NSUInteger index = 0;
-
-  if (length == 0)
-    {
-      return;
-    }
-
-  [textStorage beginEditing];
-  while (index < length)
-    {
-      NSRange range;
-      NSParagraphStyle *style;
-
-      style = [textStorage attribute: NSParagraphStyleAttributeName
-                             atIndex: index
-                      effectiveRange: &range];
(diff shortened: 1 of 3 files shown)


DTW-Thalion:
@rmottola - if you can give this a try and see if you get better results.

fredkiefer (approved):
I prefer simple solutions and this looks like one.

PR #833, opened 2026-07-27
Attach combo box popup and tooltip windows to their owner

### LINKED ISSUE #404: Combo box popups and tooltips reopen as background windows under GNOME Wayland/Xwayland
## Summary

Under GNOME on a Wayland session, GNUstep combo box dropdowns and native tooltips can appear once and then fail to reappear on subsequent attempts.

The root cause appears to be that these popup-style panels are shown as independent top-level windows rather than transient children of their owning window.

## Environment

- Debian 13
- GNOME session on Wayland
- GNUstep apps running under Xwayland
- `libs-gui` current `master`

### Combo boxes
- The first combo box dropdown often appears.
- On the second attempt, the dropdown does not appear.
- GNOME shows a notification like `"Window" is ready`, which suggests the popup is being treated as a separate top-level window.

### Tooltips
- Native GNUstep tooltips show the first time.
- Subsequent tooltip displays may fail in the same way.

## Expected behavior

Combo box dropdowns and tooltips should remain associated with their owning window and reliably reopen in the foreground instead of being managed as unrelated top-level windows.

## Notes from investigation

This appears to be a `libs-gui` issue.

The issue was resolved locally by changing `libs-gui` so the combo popup window and tooltip window are attached to their owning window as child/transient windows before being shown, and detached when dismissed.

The relevant areas are:

- `Source/NSComboBoxCell.m`
- `Source/GSToolTips.m`

The working patch makes these popup windows follow the same general ownership model already used for menu windows.

## Proposed fix

Before showing these popup panels:

- attach the combo popup window to the control's window
- attach the shared tooltip window to the view's window

When the popup is dismissed:

- detach it again

This allows the window manager to keep these popup windows associated with the owning application window.

## Patch

I am attaching the working patch in a follow-up comment.

DTW-Thalion:
Patch by @danjboyd, from #404.

Under GNOME on Wayland the combo box dropdown and the tooltip window were shown as independent top level windows, so they could appear once and then fail to reappear. This attaches the combo popup window and the shared tooltip window to their owning window as child windows before showing them, and detaches them on dismissal, following the same ownership model as menu windows.

Applied unchanged to current master, builds, and the NSComboBox tests pass. The GNOME/Wayland reopen behaviour was verified by @danjboyd on the environment described in #404; I was not able to reproduce that specific window manager under Weston/Xwayland here, so this carries their fix upstream.

Closes #404.

### DIFF
--- a/Source/GSToolTips.m
+++ b/Source/GSToolTips.m
@@ -226,6 +226,34 @@ static BOOL   isOpening = NO;
 static NSSize		offset;
 static BOOL		restoreMouseMoved;
 
+static void
+GSSyncToolTipParentWindow(NSWindow *toolTipWindow, NSWindow *ownerWindow)
+{
+  NSWindow *parentWindow = [toolTipWindow parentWindow];
+
+  if (parentWindow != nil && parentWindow != ownerWindow)
+    {
+      [parentWindow removeChildWindow: toolTipWindow];
+      parentWindow = nil;
+    }
+
+  if (ownerWindow != nil && parentWindow != ownerWindow)
+    {
+      [ownerWindow addChildWindow: toolTipWindow ordered: NSWindowAbove];
+    }
+}
+
+static void
+GSDetachToolTipParentWindow(NSWindow *toolTipWindow)
+{
+  NSWindow *parentWindow = [toolTipWindow parentWindow];
+
+  if (parentWindow != nil)
+    {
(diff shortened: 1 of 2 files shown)


fredkiefer (approved):
In other place we just use -setParentWindow:. I am unsure whether the extra effort of storing the child in the parent is needed.

PR #834, opened 2026-07-28
NSTextView: don't flicker the mouse cursor while typing (#353)

### LINKED ISSUE #353: in textviews the mouse cursor flickers while typing
While typing in a NSTextView the mouse continuously changes between arrow and hairline.

Tests:
Ink Vs. TextEdit: in ink while you type, the cursor flickers, in TextEdit on Mac, not. The mouse cursor while typing hides, then reappears on first mouse movement.

GNUMail:
Create a new message 
- On GNUstep while typing in a TextField (address field) the mouse cursor stays a hairline (it always is so while inside the field)
- On GNUstep while typing the text message, the mouse cursor changes back and forth.
- On MacOS: both fields behave the same: the cursor is a hairline when inserting, when typing starts cursor hides and reappears at first mouse movements.

This is a carry-over of bug https://savannah.gnu.org/bugs/?47523 which still applies and was erroneously closed. It may be low-impact but it is still a bug, as per examples above (and because it used to work).

Original comment by Fred of 7-1-2018

> It turns out that this was caused by a change Eric made in 2011. In the method [NSTextView.m -_updateState:] he added the update of the cursor rectangles, which is needed for internal cursor rectangles but would not be needed for the main one.
> I tried a few simple solutions to fix this, but all of these would result in additional issues.

DTW-Thalion:
## Summary

Fixes the mouse-cursor flicker rmottola reported in #353: while typing in an `NSTextView` the pointer rapidly alternated between the I-beam and the arrow.

## Root cause

`-[NSTextView _updateState:]` is scheduled (coalesced) from `-_layoutManagerDidInvalidateLayout` on every layout invalidation — i.e. effectively on every keystroke. It unconditionally called `-[NSWindow invalidateCursorRectsForView:]`, which discards and rebuilds the view's cursor rects via `-resetCursorRects`. For plain text the only cursor rect is the static I-beam covering the whole visible rect, so rebuilding it momentarily drops the I-beam and the pointer falls back to the arrow before it is re-established — the flicker.

## Fix

The invalidation is only needed when a *custom* cursor region may have moved as the text reflowed. The only attribute that currently produces a custom cursor rect is `NSLinkAttributeName` (the pointing-hand over links; `NSCursorAttributeName` is still a FIXME in `-resetCursorRects`). A new private `-_hasCustomCursorRegions` checks for that, and `-_updateState:` only invalidates the cursor rects when it returns YES.

- Plain text keeps its stable full-visible-rect I-beam and no longer flickers — a hairline inside the text rect, an arrow outside, independent of typing (the behaviour rmottola asked for).
- When links are present the cursor rects are still refreshed on reflow, so pointing-hand tracking is unchanged (this is a conservative superset — a document that contains a link anywhere still invalidates exactly as before, so there is no regression for that case).

This matches the original intent of the code — the comment already said *"in case any sections of text with custom cursors were moved."* The per-view I-beam resize on a live window resize is handled by the window's own cursor-rect reset, not by `-_updateState:`, so gating here is safe.

## Test

`Tests/gui/NSTextView/cursorRectInvalidation.m` spies on the window (an `NSWindow` subclass counting `-invalidateCursorRectsForView:`) and asserts a plain-text layout change triggers **no** cursor-rect invalidation while a layout change with a link still does. Full NSTextView set: 54 passed, All OK.

## Out of scope

rmottola also floats a *cursor auto-hide while typing* enhancement. That is a separate feature and is not included here.

Closes #353

### DIFF
--- a/Source/NSTextView.m
+++ b/Source/NSTextView.m
@@ -515,6 +515,36 @@ decoded.
 
 
 /*
+Returns YES if the text contains any attribute that -resetCursorRects turns
+into a custom (non-I-beam) cursor rect.  Currently that is only
+NSLinkAttributeName (the pointing-hand cursor over links); NSCursorAttributeName
+is still a FIXME in -resetCursorRects and so is intentionally not checked here.
+*/
+- (BOOL) _hasCustomCursorRegions
+{
+  NSTextStorage *storage = [self textStorage];
+  NSUInteger length = [storage length];
+  NSUInteger i = 0;
+
+  while (i < length)
+    {
+      NSRange effectiveRange;
+      id linkValue = [storage attribute: NSLinkAttributeName
+				 atIndex: i
+			   longestEffectiveRange: &effectiveRange
+				 inRange: NSMakeRange(i, length - i)];
+
+      if (linkValue != nil)
+	{
+	  return YES;
+	}
+      i = NSMaxRange(effectiveRange);
+    }
+
+  return NO;
+}
+
+/*
 Called when our state needs updating due to external changes. Currently,
 this happens when layout has been invalidated, and when we are resized.
 */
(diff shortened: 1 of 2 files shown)


Source/NSTextView.m
```diff
@@ -514,6 +514,36 @@ - (void) _recacheDelegateResponses
...
+  NSUInteger length = [storage length];
+  NSUInteger i = 0;
+
+  while (i < length)
+    {
+      NSRange effectiveRange;
+      id linkValue = [storage attribute: NSLinkAttributeName
+				 atIndex: i
+			   longestEffectiveRange: &effectiveRange
+				 inRange: NSMakeRange(i, length - i)];
```
fredkiefer: Maybe we could restrict the range to -visibleCharacterRanges here? It will only make a difference for a long text where only a small part is displayed. But it is that case where the new code might be slow.

DTW-Thalion: Done. It now asks for the same range -resetCursorRects builds the rects for, through a shared -_visibleCharacterRange, so the two cannot disagree. That uses glyphRangeForBoundingRectWithoutAdditionalLayout: rather than -visibleCharacterRanges, since the latter forces layout.

Besides the cost you flagged, this also stops a link anywhere in a long document making every keystroke rebuild rects that cannot have moved. Tests/gui/NSTextView/cursorRectInvalidation.m has a case for a link below the visible text.

fredkiefer (approved):

Issue #329, opened 2025-01-26
Streamline binding implementation....


gcasa:
For NSBrowser and NSTreeController/NSOutlineView, and other classes bindings have been implemented.  However, the implementation is not as clean as we want.   The functionality works and passes testing, but is not as elegant as it should be.   Testing has indicated that there is an "adapter" pattern that is being used.  This might clean up the implementation.

The acceptance criteria of this is:

1) Implement GSBindingsAdapter class
2) Clean up existing implementation of bindings in NSOutlineView, NSTreeController, and NSBrowser

GC

PR #839, opened 2026-07-28
tests: add NSTreeController content array binding tests

DTW-Thalion:
Groundwork for #329. The bindings on these classes have no test coverage, so a rework that is meant to preserve behaviour has nothing to check itself against.

This covers the content array binding on `NSTreeController`: the exposed bindings, what `-infoForBinding:` reports before and after binding, that a new value at the observed key path reaches `-content` and the arranged tree, and that unbinding clears the binding and rebinding works.

Every assertion was checked against AppKit on a macOS runner first, so the file pins behaviour that is correct rather than behaviour that merely happens to be current.

The value is replaced at the observed key path with a different array rather than mutated in place, so the propagation assertion cannot pass just because the controller and the test share one mutable array.

One divergence came out of that comparison and is deliberately not asserted here: GNUstep exposes a `content` binding on `NSTreeController` that AppKit does not expose. It is described in #842, which explains why removing it is not a drive-by change.

`NSTreeController` needs no display, so the file runs in every environment.

Refs #329.

### DIFF
--- a/Tests/gui/NSTreeController/binding.m
+++ b/Tests/gui/NSTreeController/binding.m
@@ -1,0 +1,142 @@
+/* Coverage for the NSTreeController content array binding: which bindings are
+   exposed, what -infoForBinding: reports before and after binding, that a new
+   value at the observed key path reaches -content and the arranged tree, and
+   that unbinding clears the binding and rebinding works.  Every assertion here
+   matches AppKit (verified on a macOS runner) and passes on unmodified GNUstep.
+
+   The controller needs no display, so everything here runs in any environment.
+*/
+#include "Testing.h"
+
+#include <Foundation/NSArray.h>
+#include <Foundation/NSDictionary.h>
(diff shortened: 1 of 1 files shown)


Tests/gui/NSTreeController/binding.m
```diff
@@ -0,0 +1,142 @@
...
+int main()
+{
+  NSTreeController	*tc;
+  NSObjectController	*oc;
+  NSMutableDictionary	*model;
+  NSMutableArray	*replacement;
+  NSDictionary		*info;
+  NSArray		*exposed;
+
+  START_SET("NSTreeController content array binding")
```
fredkiefer, line 75: You seem to have missed to create an autorelease pool here.

DTW-Thalion: Done.

fredkiefer (approved):

PR #840, opened 2026-07-28
tests: add NSOutlineView content binding tests

DTW-Thalion:
Groundwork for #329, alongside the tree controller and browser coverage.

This covers the content binding on `NSOutlineView`: the exposed bindings, what `-infoForBinding:` reports before and after binding, that the binding info names the observed controller, and that content bound to an `NSTreeController` reaches the rows the view reports.

Every assertion was checked against AppKit on a macOS runner first, so the file pins behaviour that is correct rather than behaviour that merely happens to be current.

Two divergences came out of that comparison and are deliberately not asserted here:

- `-[NSOutlineView exposedBindings]` reported `content`, `selectionIndexes` and `sortDescriptors` twice, since the outline view exposes them and `NSTableView` exposes them again. Fixed separately in #842. The assertions here test for membership rather than for the exact array, so they hold either way.
- GNUstep exposes `selectionIndexes` where AppKit exposes `selectionIndexPaths` and no `selectionIndexes` at all. That needs the index path binding implemented, so it is left alone and noted in #842.

Refs #329.

### DIFF
--- a/Tests/gui/NSOutlineView/binding.m
+++ b/Tests/gui/NSOutlineView/binding.m
@@ -7,6 +7,7 @@
 #include "Testing.h"
 
 #include <Foundation/NSArray.h>
+#include <Foundation/NSAutoreleasePool.h>
 #include <Foundation/NSDictionary.h>
 #include <Foundation/NSString.h>
 
@@ -61,6 +62,7 @@ node(NSString *aName)
 
 int main()
 {
+  CREATE_AUTORELEASE_POOL(arp);
   NSOutlineView		*ov;
   NSTableColumn		*col;
   NSTreeController	*tc;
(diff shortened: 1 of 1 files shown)


Tests/gui/NSOutlineView/binding.m
```diff
@@ -0,0 +1,135 @@
...
+int main()
+{
+  NSOutlineView		*ov;
+  NSTableColumn		*col;
+  NSTreeController	*tc;
+  NSObjectController	*oc;
+  NSMutableDictionary	*model;
+  NSArray		*exposed;
+
+  START_SET("NSOutlineView content binding")
```
fredkiefer, line 73: Again no autorelease pool otherwise this PR looks good

DTW-Thalion: Done.

fredkiefer (approved):

PR #842, opened 2026-07-28
Fix: report each exposed binding once

DTW-Thalion:
`-exposedBindings` walks the class and each of its superclasses and collects the bindings each one exposes, appending every list to the result. A binding exposed by both a class and one of its superclasses is therefore reported twice.

`NSOutlineView` exposes `content`, `selectionIndexes` and `sortDescriptors`, and `NSTableView` exposes the same three, so `-[NSOutlineView exposedBindings]` reported each of those names twice. AppKit reports every name once, checked on a macOS runner.

The fix skips a name already collected rather than appending it again, so the set of exposed bindings is unchanged and only the repetition goes away. `-bind:toObject:withKeyPath:options:` tests the list with `containsObject:`, so it behaves exactly as before.

`Tests/gui/NSOutlineView/exposedBindings.m` asserts that each of the three shared names appears once, that a binding exposed by only one class in the chain is still reported, and that the whole list has no repeats. Four of its five assertions fail before the change and all pass after. No regression across the outline view, table view, browser, tree controller, array and object controller, collection view, text field, button, slider and popup button tests.

Found while writing binding coverage for #329. Two further divergences from AppKit came out of the same comparison and are **not** touched here, since neither is a cosmetic change:

- `NSTreeController` exposes a `content` binding that AppKit does not expose. Removing it is not a drive-by: `-bind:toObject:withKeyPath:options:` only binds names that appear in `-exposedBindings`, so dropping it would silently turn `bind:` with `NSContentBinding` into a no-op.
- `NSOutlineView` exposes `selectionIndexes`, where AppKit exposes `selectionIndexPaths` and no `selectionIndexes` at all. Matching that means implementing the index path binding rather than renaming the exposed one.

Refs #329.

### DIFF
--- a/Source/NSKeyValueBinding.m
+++ b/Source/NSKeyValueBinding.m
@@ -64,7 +64,18 @@
       tmp = [GSKeyValueBinding exposedBindingsForClass: class];
       if (tmp != nil)
         {
-          [exposedBindings addObjectsFromArray: tmp];
+          NSUInteger i;
+
+          /* A class may expose a binding its superclass also exposes.  */
+          for (i = 0; i < [tmp count]; i++)
+            {
+              id name = [tmp objectAtIndex: i];
+
+              if ([exposedBindings containsObject: name] == NO)
+                {
+                  [exposedBindings addObject: name];
+                }
+            }
         }
   
       class = [class superclass];
--- a/Tests/gui/NSOutlineView/exposedBindings.m
+++ b/Tests/gui/NSOutlineView/exposedBindings.m
@@ -8,6 +8,7 @@
 #include "Testing.h"
 
 #include <Foundation/NSArray.h>
+#include <Foundation/NSAutoreleasePool.h>
 #include <Foundation/NSSet.h>
 #include <Foundation/NSString.h>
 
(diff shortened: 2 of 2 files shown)


Source/NSKeyValueBinding.m
```diff
@@ -64,7 +64,18 @@ - (NSArray *) exposedBindings
       tmp = [GSKeyValueBinding exposedBindingsForClass: class];
       if (tmp != nil)
         {
-          [exposedBindings addObjectsFromArray: tmp];
+          NSUInteger i;
```
fredkiefer, line 67: I like the test, but I am not sure about this fix. The behaviour is rather a bug I did introduce when adding bindings for each class. So the specific cases you found could be resolved by removing the additional bindings in these classes. On the other hand, we could test, what happens for a new class that declares an binding that already exists in a super class.

fredkiefer (approved):

Source/NSKeyValueBinding.m
DTW-Thalion: Let me think about it a bit and I'll maybe come back with something.

PR #847, opened 2026-07-28
NSBrowser: send the type select delegate methods

### LINKED ISSUE #279: NSBrowser missing 10.6+ delegate methods
Add 10.6+ delegate methods to interface as well as handle them in the class of NSBrowser.

DTW-Thalion:
Closes #279 in part. Stacked on #846, which declares these methods and is needed for this to compile, so the change to review here is the second commit.

Three of the type select methods the protocol declares were never sent:

    browser:shouldTypeSelectForEvent:withCurrentSearchString:
    browser:typeSelectStringForRow:inColumn:
    browser:nextTypeSelectMatchFromRow:toRow:inColumn:forString:

-keyDown: now asks the first of them before a keystroke starts or extends a search, and a delegate that answers NO stops the keystroke there, leaving the search string as it was. Rows are matched against the string the second returns when the delegate implements it, otherwise against the cell's string value as before. The third takes over the search: it gets the row to start from, the row to stop at, the column and the characters typed so far, and the row it returns is selected. A row outside the column, -1 included, leaves the selection where it was.

Behaviour was checked against AppKit first. The first keystroke of a search carries a nil search string; the search starts one row below the selected row and wraps back round to it; extending the string searches again from the selected row, so a longer string can keep the row it just picked; and the string the delegate supplies is what a row is matched on, not the cell's own value.

Tests/gui/NSBrowser/typeSelect.m covers those. Its 11 assertions all fail without this change. With it, NSBrowser and NSMatrix give 144 passing.

The default search is left as it was. It is case sensitive where AppKit's is not, and it keeps the selected row when that row still matches, where AppKit moves on to the next match. Both belong in a separate change.

### DIFF
--- a/Source/NSBrowser.m
+++ b/Source/NSBrowser.m
@@ -262,6 +262,8 @@ static BOOL browserUseBezels;
 - (void) _setColumnTitlesNeedDisplay;
 - (NSBorderType) _resolvedBorderType;
 - (void) _themeDidActivate: (NSNotification*)notification;
+- (NSString *) _typeSelectStringForRow: (NSInteger)row
+			      inColumn: (NSInteger)column;
 @end
 
 // Category to handle bindings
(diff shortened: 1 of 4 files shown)


Source/NSBrowser.m
```diff
@@ -2834,38 +2846,55 @@ - (void) keyDown: (NSEvent *)theEvent
...
 		  if (([sv length] > 0)
 		      && ([sv hasPrefix: _charBuffer]))
 		    {
 		      match = i;
 		      break;
 		    }
 		}
+	      if (i == n)
+		{
+		  for (i = 0; i < s; i++)
```
fredkiefer, line 2886: For these two loops I would like to see another helper method that has the old selector caching plus the check for the delegate only once. But we can add that later on.

fredkiefer (approved):

Issue #167, opened 2023-01-13
Finish NSDataLink/NSDataLinkManager implementation...


gcasa:
This is old functionality left over from OPENSTEP, but it might have some modern uses.   Need to finish this up using filesystem level notifications on appropriate platforms.

DTW-Thalion:
I went through both classes against the OpenStep API to see what is left. Summary: the API surface is complete (every documented method exists and is implemented), and the data model and persistence work (the init variants, keyed and non-keyed coding, and file and pasteboard round-trips all round-trip correctly in testing). What is missing is behaviour in a few methods, all around the parts that make a link live.

Incomplete or stubbed:

1. `-[NSDataLink updateDestination]` (NSDataLink.m) returns NO unconditionally. This is the core update operation and is already called from three places (the monitor loop, `-checkForLinkUpdates`, and NSDataLinkPanel), so wiring it up would activate the existing plumbing.

2. `-[NSDataLink openSource]` (NSDataLink.m) returns NO unconditionally. Reopening the source needs to activate the source application and locate the source selection.

3. The inotify filesystem monitoring in NSDataLinkManager is set up but never used. `-initWithDelegate:fromFile:` creates the inotify fd and starts the monitor thread, and `-monitorLoop` reads events keyed by watch descriptor from `_watchDescriptors`, but `inotify_add_watch` is never called and `_watchDescriptors` is never populated. `-addLink:at:` / `-addSourceLink:` would need to register a watch on the link's source file and map the descriptor back to the link. As it stands no source file is ever watched.

4. The monitoring is Linux only (guarded by `HAVE_SYS_INOTIFY_H`). Other platforms would need kqueue (BSD, macOS) and ReadDirectoryChangesW (Windows) behind the same start/stop interface.

5. `-[NSDataLinkManager addLinkPreviouslyAt:fromPasteboard:at:]` (NSDataLinkManager.m) has an empty match loop and always returns nil.

6. `-[NSDataLink saveLinkIn:]` runs a modal NSSavePanel; the documented behaviour is to write the `.dlf` into the given directory.

See #690 #691 as partial completion.

One design question for you before I go further on `openSource`: OpenStep assumed a system-wide inter-application link protocol for reopening a source and reselecting the linked region. I don't think there is a modern GNUstep equivalent, so this looks like it can only be a best-effort NSWorkspace open of the source file rather than a true reselect. Does that match your expectation, or is there existing machinery I should be using?

cc @gcasa

[...]

PR #853, opened 2026-07-29
Watch the source file of a linked document

DTW-Thalion:
NSDataLinkManager never watched anything. It set up an inotify descriptor, a
thread reading events, and a table mapping watch descriptors to links, but
`inotify_add_watch` was never called and the table was never filled, so an edit
to a link's source file was never noticed. This is the filesystem notification
part of #167.

A watch is now registered on the source file when a link is added and removed
when the link is removed. Each event is handed to the main thread before the
table is consulted and the delegate is asked, since the delegate is the document
and was otherwise going to be messaged from the monitoring thread. The thread is
not started when there is no descriptor, since reading a closed one returns at
once and the loop would spin.

The `-dealloc` order is corrected but unreachable:
`-[NSThread initWithTarget:selector:object:]` retains the manager, so one whose
monitor thread runs is never deallocated. Whether to watch the descriptor from
the run loop rather than a thread is a design question I have put on the issue.
Monitoring stays Linux only.

Tests/gui/NSDataLink/sourceMonitoring.m had 1 assertion failing before and none
after. The full Tests/gui suite runs 4329 assertions clean.

Refers to #167

### DIFF
--- a/Source/NSDataLinkManager.m
+++ b/Source/NSDataLinkManager.m
@@ -146,13 +146,14 @@
       _destinationLinks = [[NSMutableArray alloc] init];
       _watchDescriptors = [[NSMutableDictionary alloc] init];
       _nextLinkNumber = 1;
+      _inotifyFD = -1;
 #ifdef HAVE_SYS_INOTIFY_H
       _inotifyFD = inotify_init();
-#endif
       if (_inotifyFD < 0)
 	{
 	  NSLog(@"Failed to initialize inotify");
 	}
+#endif
       [self startMonitoring];
     }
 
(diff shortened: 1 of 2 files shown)


gcasa (approved):
I think this change looks good.  I had made a similar change in this class but never committed it because I couldn't properly test it.

Source/NSDataLinkManager.m
```diff
@@ -206,10 +209,83 @@ - (void)stopMonitoring
...
 - (void)startMonitoring
 {
+  /* With no descriptor to read there is nothing to wait on, and the loop
+     would spin. */
+  if (_inotifyFD < 0)
+    {
+      return;
+    }
+
   _monitorThread = [[NSThread alloc] initWithTarget:self selector:@selector(monitorLoop) object:nil];
```
fredkiefer, line 219: I don't like the fact that the manager itself is will never be freed because of the retain loop here. But this bug was present before your change, so it should not stop merging.

DTW-Thalion: Will make a note to come back and look at this again.

Issue #96, opened 2021-03-16
Feature Request: Implement NSPredicateEditor and NSPredicateEditorRowTemplate (SDK 10.5)


Letterus:
It would be great to have full implementations of NSPredicateEditor and NSPredicateEditorRowTemplate (first introduced with SDK 10.5). The current classes are stubs, kindly copied by Fred from mySTEP.

https://developer.apple.com/documentation/appkit/nspredicateeditor

App to test implementation: https://github.com/Letterus/app-time-tracker

PR #856, opened 2026-07-29
Feature: Implement NSRuleEditor

DTW-Thalion:
Every method of NSRuleEditor was a stub, so an editor had no rows, asked its delegate for nothing and produced no predicate. NSPredicateEditor is built on it and could not work either.

The editor now keeps its rows in a tree. A new row is filled by walking the delegate from the root of its criteria, taking the first child at each step and asking for a display value, until the delegate offers no more. The display values are laid out a row at a time, indented by how deep the row sits. The predicate of a row is assembled from the parts the delegate gives for each of its criteria, and a compound row joins the rows under it with the compound type its own criterion carries. A single editor holds one row, a list joins its rows with or, and the nesting modes that hold rows under a compound row make that row along with the first row it holds.

Choosing another item in a row does not yet rebuild the criteria under it, the formatting dictionary is kept but not applied, and rows are not animated. The class gains instance variables.

Tests/gui/NSRuleEditor/basic.m. 29 assertions fail before the change and pass after.

Refers to #96.

### DIFF
--- a/Source/NSRuleEditor.m
+++ b/Source/NSRuleEditor.m
@@ -4,8 +4,7 @@
 
    Copyright (C) 2020 Free Software Foundation, Inc.
 
-   Author: Fred Kiefer <fredkiefer@gmx.de>
-   Date:   January 2020
+   Created by Fabian Spillner on 03.12.07.
 
    This file is part of the GNUstep GUI Library.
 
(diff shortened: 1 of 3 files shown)


Source/NSRuleEditor.m
```diff
@@ -4,8 +4,7 @@
 
    Copyright (C) 2020 Free Software Foundation, Inc.
 
-   Author: Fred Kiefer <fredkiefer@gmx.de>
-   Date:   January 2020
+   Created by Fabian Spillner on 03.12.07.
```
fredkiefer, line 7: This attribution is a bit strange, where does it come from?

fredkiefer: I went back to mySTEP to check whether I copied this file over, as I did with the headers, but this file is still empty there, so this seems unlikely.

fredkiefer:
@gcasa actually I am waiting here and with most of the other PRs by Todd on your decision whether we want to include that code.

Source/NSRuleEditor.m
DTW-Thalion: It may have been a copy and paste issue on my part.

Issue #95, opened 2021-03-16
Feature Request: Implement NSDatePicker (SDK 10.4)


Letterus:
It would be great to have a full implementation NSDatePicker (first introduced with SDK 10.4). The current class is a stub, kindly copied by Fred from mySTEP.

https://developer.apple.com/documentation/appkit/nsdatepicker

App to test implementation: https://github.com/Letterus/app-time-tracker

DTW-Thalion:
NSDatePicker is no longer a stub. The class now has the full property set (datePickerStyle, datePickerMode, datePickerElements, dateValue, minDate/maxDate, calendar, locale, timeZone, colours, bezel and border) backed by NSDatePickerCell, and these work: setting and reading them behaves as documented, and both the text-field-and-stepper and clock-and-calendar styles draw. There is recent test coverage for the configuration and rendering under Tests/gui/NSDatePicker.

What is likely still incomplete is the interactive side, editing the date by typing or by clicking the calendar, which is the harder UI part. @fredkiefer do you consider the current implementation enough to close this, or is there specific behaviour you would like tracked separately?

PR #859, opened 2026-07-29
Fix: edit a date picker from the keyboard

DTW-Thalion:
A date picker took no keys at all, so the date it held could only be changed in code.

The pattern the cell shows is split into the parts the locale writes. The left and right arrow keys move between those parts, the up and down arrow keys step the part they are on, and a digit types into it. A step is a step of the calendar, so it carries into the parts above it, a day the next month is too short for moves back to that month's last day, and the result stays inside the minimum and maximum dates. The part being edited is drawn the way selected text is while the picker holds the keyboard. An edit sends the action of the control, and the delegate is asked to validate the proposed date. The values are those AppKit produces for the same keys.

NSDatePickerCell holds three more ivars. Clicking a part, the stepper, and the clock and calendar style are not part of this. The branch is cut off #858 and needs it.

Tests/gui/NSDatePicker/keyboard.m and selection.m: 24 assertions fail before the change and 0 after, 77 green across the two picker directories.

Refers to #95

### DIFF
--- a/Headers/AppKit/NSDatePickerCell.h
+++ b/Headers/AppKit/NSDatePickerCell.h
@@ -76,6 +76,9 @@ APPKIT_EXPORT_CLASS
   NSDatePickerMode _datePickerMode;
   NSDatePickerStyle _datePickerStyle;
   BOOL _drawsBackground;
+  NSInteger _selectedField;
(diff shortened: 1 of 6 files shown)


rfm (approved):
Seems reasonable

PR #861, opened 2026-07-29
Fix: draw a month and a clock in the calendar style

DTW-Thalion:
The clock and calendar style drew the same line of text as the other styles and asked for no more room than that line needs.

It now draws the month the date falls in: a row for the name of the month and the year, a row for the initials of the weekdays and six weeks of days, with the day the picker holds marked the way selected text is. Beside it, when the time elements are asked for as well, it draws a clock face with an hour and a minute hand, and the cell size covers what both need. A click picks the day under the pointer and keeps the time of day, and the arrow keys walk the grid in this style, a day across and a week up or down, which is what AppKit does with them.

Days of the months either side of this one are not drawn, and the clock is there to be read rather than set. The branch is cut off #860, which is cut off #859 and #858, and needs all three.

Tests/gui/NSDatePicker/calendar.m holds 9 assertions. It works out where a day sits from the size of the cell, so it holds whatever weekday the month starts on.

Refers to #95

### DIFF
--- a/Headers/AppKit/NSDatePickerCell.h
+++ b/Headers/AppKit/NSDatePickerCell.h
@@ -76,6 +76,9 @@ APPKIT_EXPORT_CLASS
   NSDatePickerMode _datePickerMode;
   NSDatePickerStyle _datePickerStyle;
   BOOL _drawsBackground;
+  NSInteger _selectedField;
+  NSInteger _typedValue;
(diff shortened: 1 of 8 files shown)


rfm (approved):
This seems reasonable, and is new code with no real risk of breaking things...

PR #862, opened 2026-07-29
Fix: carry a date picker cell state through a copy and an archive

DTW-Thalion:
NSCell copies its object ivars as bare pointers and then retains the ones it knows about, so a copied date picker cell shared the two colours and the two date limits with the original without holding them, and whichever of the two cells went away first gave them all back. The copy retains them now, which is what AppKit's own cell has a -copyWithZone: for.

-encodeWithCoder: never reached super. A keyed archive of the cell therefore carried four of its own settings and nothing NSCell holds, and an archive written without keys could not be read back at all, since -initWithCoder: does call super. The mode, the two date limits, the text colour and whether the background is drawn were not written either, and neither was the date: NSCell writes the text of its value rather than the value, and the text of a date does not read back as one. All of those are written now, and the new keys are read only when present, so a nib written before them still decodes.

Tests/gui/NSDatePickerCell/copying.m and coding.m: 9 assertions fail before the change and copying.m aborts on the last release; 0 fail after. Tests/gui runs 4416 assertions green.

Refers to #95

### DIFF
--- a/Headers/AppKit/NSDatePickerCell.h
+++ b/Headers/AppKit/NSDatePickerCell.h
@@ -76,6 +76,9 @@ APPKIT_EXPORT_CLASS
   NSDatePickerMode _datePickerMode;
   NSDatePickerStyle _datePickerStyle;
   BOOL _drawsBackground;
+  NSInteger _selectedField;
+  NSInteger _typedValue;
+  NSInteger _typedDigits;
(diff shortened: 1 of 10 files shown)


rfm (approved):
More improvement of new date picking code.  Looks good.

PR #867, opened 2026-07-29
Fix: clear the display mark on a view with no area

### LINKED ISSUE #80: Contents of NSAlert not visible until window is moved...
When I run the SCSI2SD app I am working on... when an NSAlert is shown, it's contents are not visible until I move the alert panel, I am not sure why this is happening.

DTW-Thalion:
Both of NSView's display walks reach a subview only when the rectangle being displayed and the subview's frame overlap. A subview with no width or height never overlaps anything, so it is never displayed, its mark is never cleared, and both walks copy that mark up to every view above it. An NSBox two pixels high is enough: it leaves its content view a height of zero, and from then on that view, the box, the window's content view and the frame view all answer YES to -needsDisplay for good.

When the overlap is empty because the subview has no area, it is now asked to display an empty rectangle, which clears its mark. A subview that has an area and merely falls outside the rectangle keeps its mark, and there is an assertion for that.

This turned up in the alert panel of #80, whose separator is such a box. The symptom reported there, contents invisible until the panel is moved, does not reproduce on Windows against current master, whether the alert is raised before the application loop starts or from inside a running one. This mark is the only defect found on that path.

Tests/gui/NSView/emptyViewNeedsDisplay.m: 5 of its 7 assertions fail before the change and none after. Tests/gui runs 4345 assertions green.

Closes #80

### DIFF
--- a/Source/NSView.m
+++ b/Source/NSView.m
@@ -2601,6 +2601,16 @@ static void autoresize(CGFloat oldContainerSize,
                       isect = [subview convertRect: isect fromView: self];
                       [subview displayIfNeededInRectIgnoringOpacity: isect];
                     }
+                  else if (NSIsEmptyRect(subviewFrame))
+                    {
+                      /* A subview with no area has nothing to draw, and is
+                         never reached by the line above, so it would keep
+                         itself and every view over it waiting for a display
+                         that cannot happen.  Let it settle its own flag.
+                      */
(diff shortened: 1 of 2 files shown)


rfm:
Cool thing to spot.

fredkiefer:
Looks great, but I will have to review the whole working here. I wrote most of that, but that was a long time ago and I have to understand all of it again. Sorry,we could also just merge and I go through that later on.

PR #869, opened 2026-07-30
Implement NSTextCheckingController

DTW-Thalion:
Every method of NSTextCheckingController was an empty stub and the spell document tag was left at zero, so a client wired to one got nothing back.

The controller now:

-  takes a unique spell document tag when it is created, 
- asks its client for the text of a range and honours the shorter range a client may return,
- marks what the spell checker finds with the spelling state attribute,
-  carries on from the end of the selection in checkSpelling: so that asking again moves to the next word, 
- applies a correction in changeSpelling: and reselects it at the length of the replacement, 
- ignores a word for its own document tag, 
- offers the guesses for a marked range as a menu, and stops touching the client once it is invalidated. 

Those five behaviours were measured on a macOS 26 runner first.

Left alone: the substitutions panel, which GNUstep has no equivalent for, the candidate list, and data and link detection. The NSTextEdited annotation AppKit sets for inserted text has no constant here.

Tests/gui/NSTextCheckingController/basic.m: 10 of its assertions fail before the change and none after. Seven more need a spell checking service and are skipped where there is none, as they are here. 

Refers to #76

### DIFF
--- a/Source/NSTextCheckingController.m
+++ b/Source/NSTextCheckingController.m
@@ -1,40 +1,55 @@
 /* Implementation of class NSTextCheckingController
    Copyright (C) 2020 Free Software Foundation, Inc.
-   
+
    By: Gregory John Casamento
    Date: 02-08-2020
 
    This file is part of the GNUstep Library.
-   
+
    This library is free software; you can redistribute it and/or
    modify it under the terms of the GNU Lesser General Public
    License as published by the Free Software Foundation; either
    version 2.1 of the License, or (at your option) any later version.
-   
+
    This library is distributed in the hope that it will be useful,
    but WITHOUT ANY WARRANTY; without even the implied warranty of
    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
    Lesser General Public License for more details.
-   
+
    You should have received a copy of the GNU Lesser General Public
    License along with this library; if not, write to the Free
    Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
    Boston, MA 02110 USA.
 */
 
-#import "AppKit/NSTextCheckingController.h"
-#import "AppKit/NSSpellChecker.h"
+#import <Foundation/NSArray.h>
+#import <Foundation/NSAttributedString.h>
+#import <Foundation/NSDictionary.h>
+#import <Foundation/NSEnumerator.h>
+#import <Foundation/NSString.h>
+#import <Foundation/NSValue.h>
 
-/*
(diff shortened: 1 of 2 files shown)


rfm (commented):
While this looks good,  I get
Skipped set:         basic.m 158 ... no spell checking service is available
even though my config.log reports that the spell checker was found.

I need to figure out whey there is no spell checking service (or why the code thinks there is no service) on my system.

rfm:
My NSLanguages array says (IrelandEnglish, English)

Running make_services --verbose tells me that there is a spell checking service for 'AmericanEnglish'

Stepping through under gdb, it seems that NSSpellChecker only looks at the first value in NSLanguages (IrelandEnglish) and tries to find a spell checking service for that language.  I'm not sure it's correct to only look at the first/preferred value, but even if it wasn't there would be no match.

I think the problem I'm having is with the GSspell.m tool, rather than anything to do with this PR, but I can't fully test at the moment.

rfm:
So,  by default the spelling service (GSspell) is installed as supporting only american english.  This is an issue: when we install we should automatically launch the service so that it reads the available languages and updates the database of what languages are available.

If/when I do that manually (by running libs-gui/Tools/GSspell.service/GSspell) and then run make_services --verbose, I can see all the supported languages listed (various forms of english).  However, because most of the world doesn't consider Ireland to be a significant country, this does not include IrelandEnglish (though it now includes plain English).

So things still don't work, because the NSSpellChecker code only looks for the preferred/first language listed in NSLanguages.
I can obviously manually override the NSLanguages array to fix that by putting 'English' first), but I think we should find out whether the correct behavior is to only try the preferred language, or to try each language in turn.

DTW-Thalion:
Let me come back with a fix for this, I will figure it out today.

rfm:
Probably having the tests standardise things would solve this.
eg. by setting the language to be the one provides by default ...

  [NSUserDefaults setUserLanguages:
    [NSArray arrayWithObject: @"AmericanEnglish"]];

Source/NSTextCheckingController.m
```diff
@@ -66,44 +81,202 @@ - (void) setSpellCheckerDocumentTag: (NSInteger)tag
...
+  if (correction == nil)
+    {
+      return;
+    }
+
+  selected = [_client selectedRange];
+  [self _substringInRange: selected actualRange: &actual];
+
+  [_client replaceCharactersInRange: actual
+               withAnnotatedString: [[[NSAttributedString alloc]
```
fredkiefer, line 154: Shouldn't we try to get some of the attributes from the replaced string?

PR #871, opened 2026-07-30
Fix: use the exception macros instead of hand written handlers in tests. (Part 2)

DTW-Thalion:
These tests wrap their body in an autorelease pool of their own although the body already sits inside a START_SET. START_SET creates a pool and END_SET releases it, so the outer one does nothing.

Removed from 376 files. The change is deletions only, 856 of them, and adds no line anywhere.

Two files are deliberately left out. TextSystem/deallocation.m calls RECREATE_AUTORELEASE_POOL inside its set and so needs the declaration. NSApplication/basic.m needs an unreachable early return taken out before its pool can go, which is a change of a different kind and is sent separately.

Run gnustep-tests from Tests/gui. 4353 assertions pass and one is a dashed hope, before and after, with all 448 test programs building.

It was also run with base and gui built GNUSTEP_WITH_ASAN=1 and leak checking on, where the totals are 4351 and two, again the same before and after. The set of files LSAN reports is unchanged, no file's leak total moves by a single byte, and nothing reports being autoreleased with no pool in place, so none of the pools removed here was doing any work.

### DIFF
--- a/Tests/gui/NSFontCollection/mutable.m
+++ b/Tests/gui/NSFontCollection/mutable.m
@@ -12,12 +12,12 @@
 int
 main(int argc, char **argv)
 {
-  NSFontDescriptor *fd = [NSFontDescriptor fontDescriptorWithFontAttributes:
-    [NSDictionary dictionaryWithObject: @"Helvetica" forKey: NSFontFamilyAttribute]];
-  NSFontDescriptor *fd2 = [NSFontDescriptor fontDescriptorWithFontAttributes:
-    [NSDictionary dictionaryWithObject: @"Courier" forKey: NSFontFamilyAttribute]];
-
   START_SET("set query and exclusion descriptors")
+    NSFontDescriptor *fd = [NSFontDescriptor fontDescriptorWithFontAttributes:
+      [NSDictionary dictionaryWithObject: @"Helvetica" forKey: NSFontFamilyAttribute]];
+    NSFontDescriptor *fd2 = [NSFontDescriptor fontDescriptorWithFontAttributes:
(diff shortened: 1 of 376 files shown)


Tests/gui/NSFontCollection/mutable.m
fredkiefer: You should start the set before creating any autoreleased objects. I'll merge this anyway as this PR is already way to big.

Tests/gui/NSFontDescriptor/basic.m
```diff
@@ -16,7 +16,6 @@
 int
 main(int argc, char **argv)
 {
-  CREATE_AUTORELEASE_POOL(arp);
   NSDictionary *attrs = [NSDictionary dictionaryWithObjectsAndKeys:
```
fredkiefer, line 19: Same here and probably in other files to.

Tests/gui/NSFontCollection/mutable.m
DTW-Thalion: I will check those after - there were a couple of special cases, but I agree they should be all within the set.

fredkiefer: Don't worry, I will change stuff on the fly during my review

fredkiefer (approved):

Tests/gui/NSFontCollection/mutable.m
fredkiefer: You were right to worry, I managed to break things. I am now cleaning up my changes again.

DTW-Thalion: This is a tricky beast to manage - bash scripts only get us so far, then we start to see where little inconsistencies in how stuff is written (like my mistakes in not using the macros) can really come back to bite us.

PR #875, opened 2026-07-30
tests: add the general drawing tests, run against any backend

DTW-Thalion:
Tests of general AppKit drawing live under one backend's directory in libs-back, guarded to that backend, so each runs against one graphics backend only. gnustep/libs-back#167 proposed moving those here so they run against whichever backend is installed. This is the drawing half of that.

Ten files, covering fills, alpha compositing, strokes with their caps, joins and dashes, curves, the winding rule, clipping, the graphics state stack, the compositing operators, image drawing and shadows. A new Tests/gui/GSDrawTest.h holds the offscreen draw and pixel read-back helpers, beside the existing GSRenderTest.h.

Each skips when no backend is installed, and when the installed one cannot draw offscreen, which is how the headless backend reports rather than failing every assertion.

Gradients are left out until gnustep/libs-back#190 merges, since before it they hang rather than fail on every backend but cairo and opal.

Run against eight configurations: green on x11 with cairo and with art, and on wayland with cairo; skipped on headless; 9 failed on xlib, 10 on opal, 6 on win32 with cairo, 17 on winlib. Those are backend gaps, two already gnustep/libs-back#188 and #189.

The libs-back copies are not removed here, and fonts are not moved yet.

### DIFF
--- a/Tests/gui/GSDrawTest.h
+++ b/Tests/gui/GSDrawTest.h
@@ -1,0 +1,143 @@
+/* GSDrawTest.h
+
+   Shared support for the drawing tests: draw into an offscreen image through
+   the AppKit path, read the pixels back, and check them.  This exercises
+   whichever backend is installed rather than any one of them, so the same
+   test covers cairo, art, xlib and winlib.
+
+   These helpers need a backend, so a test that uses them must keep the usual
+   START_SET / SKIP guard.  Colours are checked with a small tolerance,
+   because a backend may carry them through fixed-point arithmetic.
+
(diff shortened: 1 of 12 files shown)


fredkiefer (approved):
Looks good, but all these tests seem to be missing a surrounding autorelease pool. Maybe we added these to the gui tests later on, after @rfm started his memory testing.

rfm:
A set of testcases manages autorelease for the tests within it, including making sure that if you do put of the set the pool is still cleaned up.

rfm:
Autocorrect error: not put of, skip out of.

PR #880, opened 2026-08-01
Fix: NSBitmapImageRep colorAtX:y: divides by a zero alpha

DTW-Thalion:
The colour components of a premultiplied bitmap are divided by the alpha to undo the premultiplication, so a pixel whose alpha is zero divides by zero. `colorAtX:y:` then answers a colour whose components are infinity or a NaN rather than a colour, and a NaN fails every comparison, so a caller testing a component gets neither the answer it expects nor the opposite of it. A pixel like that is not hypothetical: an offscreen drawable can hold a colour with a zero alpha, and reading the application icon back on the xlib backend produces exactly this.

A zero alpha leaves nothing to divide by, so the components are taken as they are. That is what AppKit answers: a stored 255 with a zero alpha reads back as 1.0, and an entirely zero pixel reads back as black and clear. The white, black and CMYK branches divided in the same way and were changed with the RGB one.

The test is Tests/gui/NSBitmapImageRep/colorAtZeroAlpha.m, which needs no backend and runs anywhere.

Five of its ten assertions fail before the change and none after.

### DIFF
--- a/Source/NSBitmapImageRep.m
+++ b/Source/NSBitmapImageRep.m
@@ -960,9 +960,11 @@ _set_bit_value(unsigned char *base, long msb_off, int bit_width,
               ia = pixelData[3];
             }
 
-          // Scale to [0.0 ... 1.0] and undo premultiplication
+          /* Scale to [0.0 ... 1.0] and undo premultiplication.  A zero alpha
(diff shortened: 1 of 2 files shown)


fredkiefer (approved):
Thank you for fixing this

PR #888, opened 2026-08-03
Fix: read a coder's capabilities through the ImageMagick accessors

### LINKED ISSUE #887: ImageMagick support fails to compile
Probably a side-effect of #689 
```

GSImageMagickImageRep.m:65:14: error: no member named 'blob_support' in 'struct _MagickInfo'
   65 |     && info->blob_support != MagickFalse);
      |        ~~~~  ^
GSImageMagickImageRep.m:101:14: error: no member named 'blob_support' in 'struct _MagickInfo'
  101 |     && info->blob_support == MagickFalse);
      |        ~~~~  ^
2 errors generated.

```

DTW-Thalion:
GSImageMagickCanDecode and GSImageMagickUsesDelegate read the decoder and blob_support members of MagickInfo directly. ImageMagick 7 keeps neither, holding those capability bits in a flags member instead, so the file fails to compile against it with "no member named 'blob_support' in 'struct _MagickInfo'".

Both values have accessor functions, GetImageDecoder and GetMagickBlobSupport, declared with the same signatures in ImageMagick 6 and ImageMagick 7. The four member reads become calls to them. The meaning is unchanged and the file no longer depends on the shape of the structure.

Only ImageMagick 6 is installed in CI, so the version that fails was not built; the ImageMagick 7 declarations were read from its header rather than exercised. ImageMagick support is compiled only when configure is given --enable-imagemagick, so the configuration this broke is not the one CI builds, and neither the original change nor this one is covered there.

To build it: ./configure --enable-imagemagick, then make -C Source.

With ImageMagick 6 the file compiled before this change and compiles after it, and the object file now names GetImageDecoder and GetMagickBlobSupport where it previously named neither. Against ImageMagick 7 it did not compile at all.

Closes #887

### DIFF
--- a/Source/GSImageMagickImageRep.m
+++ b/Source/GSImageMagickImageRep.m
@@ -61,8 +61,8 @@ static BOOL
 GSImageMagickCanDecode(const MagickInfo *info)
 {
   return (info != (const MagickInfo *)NULL
-    && info->decoder != (DecodeImageHandler *)NULL
-    && info->blob_support != MagickFalse);
+    && GetImageDecoder(info) != (DecodeImageHandler *)NULL
+    && GetMagickBlobSupport(info) != MagickFalse);
 }
 
 /* Look up the ImageMagick coder that claims the leading bytes of DATA, or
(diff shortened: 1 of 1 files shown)


fredkiefer (approved):
Looks good and hopefully resolves the problem Riccardo is facing.

rmottola:
Tested yesterday with compilation against ImageMagick 6 and ImageMagick 7, builds on both. FTBFS fixed.

PR #893, opened 2026-08-05
Fix: number NSTextAlignment as AppKit does

### LINKED ISSUE #752: NSTextAlignment NSRightTextAlignment and NSCenterTextAlignment have different values from AppKit
The NSTextAlignment constants in AppKit/NSText.h are ordered:

    NSLeftTextAlignment      = 0
    NSRightTextAlignment     = 1
    NSCenterTextAlignment    = 2
    NSJustifiedTextAlignment = 3
    NSNaturalTextAlignment   = 4

AppKit uses NSTextAlignmentRight = 2 and NSTextAlignmentCenter = 1, i.e. Right and Center are swapped relative to GNUstep (Left, Justified and Natural agree).

Code that uses the symbolic names round-trips correctly, but the raw integer stored differs (for example -[NSText setAlignment: NSRightTextAlignment] yields 1 on GNUstep and 2 on AppKit). This affects keyed archives written by one and read by the other, and any code comparing raw alignment values.

Aligning the values with AppKit would be an ABI change (archives, and any code relying on the current numbers), so I have not sent a patch. Filing to check whether the project wants to align these (and, relatedly, NSFloatingWindowLevel which is 2 vs AppKit 3, and NSTabViewControllerTabStyleUnspecified which is 3 vs AppKit -1). @fredkiefer ?

Found while adding NSText/paragraph-alignment tests.

DTW-Thalion:
NSTextAlignment numbered right 1 and centre 2, where AppKit numbers centre 1 and right 2 (issue #752). Left, justified and natural already agreed. Code that uses the names round-tripped, but the raw value differs, so a keyed archive written by one library and read by the other swaps right and centre.

The constants are now left 0, centre 1, right 2, justified 3 and natural 4, read from a macOS runner. This breaks the values already stored in archives, which is what was agreed in the issue.

NSParagraphStyle keeps its own alignment in an archive that is not keyed, so its class version goes to 4 and a version below that has right and centre swapped back on decode. NSCell stores the alignment in the NSCellFlags2 bits of a keyed archive, where there is no version to test, so a document saved with a right aligned cell reads as centred until it is saved again. GSXib5KeyedUnarchiver reads the alignment from the strings in the xib and needs no change.

Tests/gui/NSText/alignmentValues.m pins the five values, the current names against the deprecated ones, the class version, and a paragraph style through an archive that is not keyed. It fails 4 and passes 29 before, and passes 11 more after; Tests/gui is otherwise unchanged.

Closes #752

### DIFF
--- a/Headers/AppKit/NSText.h
+++ b/Headers/AppKit/NSText.h
@@ -72,10 +72,10 @@
 typedef NSInteger NSTextAlignment;
 enum {
   NSLeftTextAlignment = 0,
-  NSRightTextAlignment,
-  NSCenterTextAlignment,
-  NSJustifiedTextAlignment,
-  NSNaturalTextAlignment
+  NSCenterTextAlignment = 1,
+  NSRightTextAlignment = 2,
+  NSJustifiedTextAlignment = 3,
+  NSNaturalTextAlignment = 4
 };
 
 #if OS_API_VERSION(MAC_OS_X_VERSION_10_11, GS_API_LATEST)
--- a/Source/NSParagraphStyle.m
+++ b/Source/NSParagraphStyle.m
@@ -216,9 +216,9 @@ static NSParagraphStyle	*defaultStyle = nil;
 {
   if (self == [NSParagraphStyle class])
     {
-      /* Set the class version to 2, as the writing direction is now 
-	 stored in the encoding */
-      [self setVersion: 3];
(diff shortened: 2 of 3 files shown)


fredkiefer (approved):
As @gcasa approved in the corresponding issue, lets go ahead and merge this.

PR #903, opened 2026-08-09
Add setTemplate:/template in NSImage (macOS10.5) stencil/template images

gcasa:
This PR adds setTemplate:/template to enable stencil images to be created.  This is to improve compatibility with currently macOS.

The test for this is here: https://github.com/gcasa/NSImage_template_test

fixed_example

### DIFF
--- a/Headers/AppKit/NSImage.h
+++ b/Headers/AppKit/NSImage.h
@@ -517,3 +517,4 @@ APPKIT_EXPORT_CLASS
 #endif
 
 #endif // _GNUstep_H_NSImage
+
--- a/Headers/AppKit/NSImage.h
+++ b/Headers/AppKit/NSImage.h
@@ -142,6 +142,7 @@ APPKIT_EXPORT_CLASS
     unsigned	cacheSeparately: 1;
     unsigned	unboundedCacheDepth: 1;
     unsigned	syncLoad: 1;
+    unsigned	template: 1;
   } _flags;
   NSMutableArray	*_reps;
   NSColor		*_color;
@@ -458,6 +459,9 @@ APPKIT_EXPORT_CLASS
 - (void) setFlipped: (BOOL)flag;
 - (BOOL) isFlipped;
 
+- (void) setTemplate: (BOOL)isTemplate;
+- (BOOL) isTemplate;
+
 //
 // Assigning a Delegate 
 //
@@ -513,4 +517,3 @@ APPKIT_EXPORT_CLASS
 #endif
 
 #endif // _GNUstep_H_NSImage
-
--- a/Source/NSCachedImageRep.m
+++ b/Source/NSCachedImageRep.m
@@ -50,7 +50,6 @@
 @interface GSCacheW : NSWindow
 @end
 
-
 @implementation GSCacheW
 
 - (void) _initDefaults
@@ -326,3 +325,4 @@
 }
 
 @end
+
--- a/Source/NSCachedImageRep.m
+++ b/Source/NSCachedImageRep.m
@@ -50,6 +50,7 @@
 @interface GSCacheW : NSWindow
 @end
 
+
 @implementation GSCacheW
 
 - (void) _initDefaults
@@ -122,6 +123,7 @@
     return nil;
   
   [self setAlpha: alpha];
+  [self setOpaque: !alpha];
   [self setBitsPerSample: NSBitsPerSampleFromDepth(aDepth)];
   [self setSize: aSize];
   [self setPixelsWide: pixelsWide];
@@ -324,4 +326,3 @@
 }
 
 @end
-
--- a/Source/NSImage.m
+++ b/Source/NSImage.m
@@ -2309,8 +2309,10 @@ iterate_reps_for_types(NSArray* imageReps, SEL method)
   CGFloat blue = 0.0;
   CGFloat alpha = 1.0;
   NSGraphicsContext *ctxt = GSCurrentContext();
+  NSImageInterpolation interpolation = [ctxt imageInterpolation];
 
-  if (dstRect.size.width <= 0 || dstRect.size.height <= 0)
+  if (dstRect.size.width <= 0 || dstRect.size.height <= 0
+      || repSrcRect.size.width <= 0 || repSrcRect.size.height <= 0)
     {
       return;
     }
@@ -2318,16 +2320,18 @@ iterate_reps_for_types(NSArray* imageReps, SEL method)
   DPScurrentrgbcolor(ctxt, &red, &green, &blue);
   DPScurrentalpha(ctxt, &alpha);
 
-  maskRect = NSMakeRect(0, 0, dstRect.size.width, dstRect.size.height);
+  maskRect = NSMakeRect(0, 0, repSrcRect.size.width, repSrcRect.size.height);
   mask = [[NSCachedImageRep alloc]
-	    initWithSize: dstRect.size
-	      pixelsWide: ceil(dstRect.size.width)
-	      pixelsHigh: ceil(dstRect.size.height)
+	    initWithSize: maskRect.size
+	      pixelsWide: ceil(maskRect.size.width)
+	      pixelsHigh: ceil(maskRect.size.height)
 		   depth: [[NSScreen mainScreen] depth]
 		separate: YES
 		   alpha: YES];
 
   [[[mask window] contentView] lockFocus];
+  [[NSGraphicsContext currentContext]
+    setImageInterpolation: interpolation];
   NSRectFillUsingOperation(maskRect, NSCompositeClear);
   DPSsetrgbcolor(GSCurrentContext(), red, green, blue);
   DPSsetalpha(GSCurrentContext(), alpha);
--- a/Source/NSImage.m
+++ b/Source/NSImage.m
@@ -491,7 +491,7 @@ repd_for_rep(NSArray *_reps, NSImageRep *rep)
       if ([path length] != 0) 
         {
 	  image = [[[[GSTheme theme] imageClass] alloc]
-	    initByReferencingFile: path];
+		    initByReferencingFile: path];
           if (image != nil)
             {
               [image setName: realName];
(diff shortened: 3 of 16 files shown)


gcasa:
Disclosure: AI was used to tighten the code and in testing.  I need to review for formatting, etc.

Source/NSImage.m
```diff
@@ -2259,6 +2294,56 @@ + (void) _reloadCachedImages
...
+  CGFloat blue = 0.0;
+  CGFloat alpha = 1.0;
+  NSGraphicsContext *ctxt = GSCurrentContext();
+
+  if (dstRect.size.width <= 0 || dstRect.size.height <= 0)
+    {
+      return;
+    }
+
+  DPScurrentrgbcolor(ctxt, &red, &green, &blue);
```
fredkiefer, line 2320: Do we really need to get and set the current colour and alpha? Should these be already active?

Source/NSImage.m
```diff
@@ -2259,6 +2294,56 @@ + (void) _reloadCachedImages
...
+  DPScurrentalpha(ctxt, &alpha);
+
+  maskRect = NSMakeRect(0, 0, dstRect.size.width, dstRect.size.height);
+  mask = [[NSImage alloc] initWithSize: dstRect.size];
+  [mask setCacheMode: NSImageCacheNever];
+
+  [mask lockFocus];
```
fredkiefer: I must admit I don't remember all the details, but as far as I remember we already store a background colour for the cached image reps. Would it be possible to reuse this mechanism here? That is, for a template image to use a cache rep that uses the current colour. Please point your AI to that code and discuss the possibility and the issues with that approach.

Source/NSImageCell.m
```diff
@@ -264,6 +265,18 @@ - (void) drawInteriorWithFrame: (NSRect)cellFrame inView: (NSView *)controlView
       fraction = 0.5;
     }
 
+  if ([_cell_image isTemplate])
```
fredkiefer, line 269: What surprises me here is that you always use the text colours. I would have expected the background colours here.

Tests/gui/NSImage/drawingPixels.m
```diff
@@ -103,6 +110,26 @@
...
     "a 24-bit rgb image without alpha decodes to its colour");
 
+  /* A template image uses source alpha as a stencil and the current drawing
+   * colour as the fill. */
+  rep = [[NSBitmapImageRep alloc]
+    initWithBitmapDataPlanes: NULL pixelsWide: 2 pixelsHigh: 1
+                bitsPerSample: 8 samplesPerPixel: 4 hasAlpha: YES isPlanar: NO
+               colorSpaceName: NSDeviceRGBColorSpace bytesPerRow: 8 bitsPerPixel: 32];
+  d = [rep bitmapData];
+  d[0] = 0;   d[1] = 255; d[2] = 0; d[3] = 255;
```
fredkiefer, line 126: This image does not look like the description of a template image in the documentation. I would have expected to see only black or transparent pixels.

Source/NSImage.m
fredkiefer: Looking closer at your implementation, I think the big difference is the drawing operation NSCompositeDestinationIn. So these two colours serve a different purpose.

Source/NSImage.m
gcasa: This call captures the current drawing color/alpha before creating the offscreen mask. The mask is filled with that color, then clipped by the template image’s alpha using NSCompositeDestinationIn, which produces the tinted template image. NSImageCacheNever avoids reusing a cached mask with the wrong color, alpha, or size.

In short: this builds a non-cached, offscreen tinted mask using the current drawing color, so template images render with the correct color/alpha for this draw.  It preserves the existing color to make sure it doesn't change.

fredkiefer: Sorry, this sounds like an AI generated answer. It is correct and sounds rather deep, but it does not address what I asked for.

In short: Would the code, when you remove the get and set of the current colour behave the same?

gcasa: You're frustrating me with the anti-AI attitude and the accusations.  Let's be clear about that.

gcasa: The issue is that this wouldn't work reliably without preserving the color as is.  The current color in the original context is not guaranteed to carry over.  The get and set preserve the caller's current color settings and applies them inside the temporary mask (for the template).   I thought the above explained that.

gcasa: It's my understanding that it might not reliably work.  That's why it copies the color/alpha settings before filling.  If the color is reliably carried over, then I'd be happy to change it.

Source/NSImage.m
gcasa: I think it is possible to do that.  I am looking into whether the NSImageCacheNever is even needed.  The test I have is rendering with a black background when it SHOULD be transparent.  I'm looking into why.  I think it's because lockFocus was being prevented from creating the backing cache.   I'm working on a fix for that.

Tests/gui/NSImage/drawingPixels.m
gcasa: I'm working on fixing that now.

Source/NSImageCell.m
gcasa: From what I gather from testing this on macOS, it's because template images are drawn as foreground content, like glyphs.

Tests/gui/NSImage/drawingPixels.m
gcasa: example_gs
example_macos

Getting closer, as you can see.  I don't know why the example is showing red on macOS and white on GNUstep.

gcasa: I just figured out... contentTintColor on GNUstep isn't functional.  I think that's outside of the scope of this PR.  So, I think this is as close as we are going to get here.  It is clipping properly now, though.

Source/NSImageCell.m
gcasa: Actually, I'm wrong on this... it should use contentTintColor.

Tests/gui/NSImage/drawingPixels.m
gcasa: I'm going to go ahead and do the tint change.

gcasa:
fixed_example

Implemented the contentTintColor.

Tests/gui/NSImage/drawingPixels.m
gcasa: It should now look like the description.  Let me know if it works differently for you.

Source/NSImageCell.m
gcasa: Implemented, so now it only uses the text color as a backup if contentTintColor doesn't exist on the cell being used.

Source/NSImage.m
gcasa: I'm no longer making this call.  It's working now without it.

Source/NSImage.m
gcasa: The current code uses this and works properly.  I am going to test without it and if that works I will remove it and resolve this.

gcasa: I verified the template code doesn't work predictably without that code in place.   Once I made the contentTintColor change it became obvious.  Since the code is using contentTintColor to fill the template, it might be different than the current color.

Headers/AppKit/NSImage.h
```diff
@@ -513,4 +517,3 @@ APPKIT_EXPORT_CLASS
 #endif
 
 #endif // _GNUstep_H_NSImage
-
```
fredkiefer, line 516: Looks like you removed the return at the end of the file. Better restore it.

Source/NSCachedImageRep.m
```diff
@@ -50,6 +50,7 @@
 @interface GSCacheW : NSWindow
 @end
 
+
```
fredkiefer: Why another newline here?

Source/NSCachedImageRep.m
```diff
@@ -324,4 +326,3 @@ - (id) copyWithZone: (NSZone *)zone
 }
 
 @end
-
```
fredkiefer, line 327: And one lost here

Source/NSImageView.m
fredkiefer, line 469: Why not decode/encode the tint colour in the none keyed case?

Source/NSImageView.m
```diff
@@ -181,6 +189,17 @@ - (void) setAllowsCutCopyPaste: (BOOL)flag
   _ivflags.allowsCutCopyPaste = flag;
 }
 
+- (NSColor *) contentTintColor
+{
+  return _contentTintColor;
```
fredkiefer, line 194: Did you test whether the colour gets stored here or is passed on to the cell, like NSButton does?

gcasa: No.  I will test that.  I tested NSImageCell.  I will test that as well.

Source/NSImage.m
fredkiefer: That is a shame. This means your code will only work for colours that can be converted to RGB. We will have to live with that restriction.

Source/NSImageView.m
gcasa: I added the code to do this... will commit ASAP

Source/NSCachedImageRep.m
gcasa: Fixed.

Source/NSImageView.m
gcasa: Tested; I added a test for that.

gcasa: Sorry!! It works. :)

Source/NSCachedImageRep.m
gcasa: Fixed

Headers/AppKit/NSImage.h
gcasa: Fixed.

fredkiefer (approved):

PR #909, opened 2026-08-11
Hide an application properly when its windows are miniaturized

### LINKED ISSUE #897: hide: leaves isHidden NO on Windows and unhide: then restores nothing
On Windows an application reports that it is not hidden after it has been
hidden, and unhiding restores nothing.

A probe:

```
isHidden before          0
isHidden after hide:     0
isHidden after unhide:   0
```

`-[NSApplication hide:]` at `Source/NSApplication.m:2543` takes a separate
path on Windows, guarded on the MINGW32 macro: it calls
`[self miniaturizeAll: sender]` and returns, where the other path sets
`_app_is_hidden`, posts `NSApplicationWillHideNotification` and orders the
windows out.

Everything else follows from that flag staying NO. `-isHidden`
(`NSApplication.m:2666`) answers it directly, so it contradicts the call that
just ran. `-unhide:` acts only when the flag is set (`NSApplication.m:2683`),
so it never deminiaturises what `-hide:` miniaturised. Neither hide
notification is posted. The Hide and Show menu item still swaps its title and
action, so the menu offers Show for an application that reports itself
visible.

`Tests/gui/NSApplication/basic.m:239` fails on Windows for this reason and
passes on Linux.

DTW-Thalion:
Fixes #897.

On Windows hide: called miniaturizeAll: and returned, leaving _app_is_hidden NO. isHidden then contradicted the call that had just run, unhide: did nothing because it acts only on a hidden application, and neither hide notification was posted.

The same omission was reached on any platform, not only Windows: the branch taken when GSSuppressAppIcon is set also miniaturized the windows and did no bookkeeping either.

Both now call one private method that posts NSApplicationWillHideNotification, miniaturizes, swaps the Hide menu item to Show and sets the flag before posting NSApplicationDidHideNotification.

unhideWithoutActivation deminiaturizes those windows again. It tells the two ways of hiding apart by whether any window was ordered out: the ordering path fills the hidden list, the miniaturizing path leaves it empty. A window the user had miniaturized before hiding comes back with them.

Tests/gui/NSApplication on MSYS2 UCRT64, the platform of the Windows MinGW job: basic.m:239 failed before and 41 assertions pass after. Line 242 passed before only because the flag never became YES. On Linux the set gives 41 passed, against the built library rather than the installed one.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,14 @@
 2026-08-11 Todd White <todd.white@thalion.global>
 
+	* Source/NSApplication.m: Set the hidden flag, post the hide
+	notifications and swap the Hide menu item where an application
+	is hidden by miniaturizing its windows, which is the path taken
+	on Windows and the path taken anywhere when GSSuppressAppIcon is
+	set.  Deminiaturize those windows again in
+	-unhideWithoutActivation.
+
+2026-08-11 Todd White <todd.white@thalion.global>
+
 	* Printing/GSWIN32/GSWIN32Printer.m: Give an enumerated printer
 	the generic PostScript PPD and parse it, so that
 	-pageSizeForPaper: and +[NSPrintInfo sizeForPaperName:] answer a
--- a/Source/NSApplication.m
+++ b/Source/NSApplication.m
@@ -2540,22 +2540,45 @@ image.</p><p>See Also: -applicationIconImage</p>
  */
 - (void) hide: (id)sender
 {
-#ifdef __MINGW32__
-  [self miniaturizeAll: sender];
-#else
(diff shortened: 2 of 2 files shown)


fredkiefer:
The code itself looks good, what I don't like is the duplication. The old code has almost the same statements. Wouldn't it be possible to move the #ifdef further down and share most of the surrounding code. Otherwise the problem will be that in the future somebody will forget to update one of these implementations.

DTW-Thalion:
Will do.

DTW-Thalion:
Done.

fredkiefer (approved):

PR #911, opened 2026-08-12
Add the feature that allows windows added via addChildWindow to follow the parent

gcasa:
Add the feature that allows windows added via addChildWindow to follow the parent.  This method should cause the child window to follow the parent's movements.

### DIFF
--- a/Source/NSWindow.m
+++ b/Source/NSWindow.m
@@ -171,6 +171,7 @@ static GSWindowAnimationDelegate *animationDelegate;
 - (NSView *) _windowView;
 - (NSView *) _borderView;
 - (NSScreen *) _screenForFrame: (NSRect)frame;
+- (void) _moveChildWindowsByOffset: (NSSize)offset;
 @end
 
 @interface NSDrawer (GNUstepPrivate)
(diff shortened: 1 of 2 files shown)


Source/NSWindow.m
```diff
@@ -201,6 +202,37 @@ - (void) _handleAutodisplay
...
+      return;
+    }
+
+  count = [_children count];
+  if (count == 0)
+    {
+      return;
+    }
+
+  children = [_children copy];
```
fredkiefer: Why the copy here? If this is to prevent mutation, why don't you use an interator instead?

Source/NSWindow.m
```diff
@@ -2400,6 +2434,12 @@ - (void) setFrame: (NSRect)frameRect display: (BOOL)flag
    * We will recieve an event to tell us when the resize is done.
    */
   [self _applyFrame: frameRect];
+  if (_windowNum == 0)
```
fredkiefer, line 2435: This condition looks wrong. Why would we only move the sub windows of a window that is not started or terminated? Or is this to handle windows directly while they are not displayed, while the rest of the code takes care of displayed ones?

Source/NSWindow.m
```diff
@@ -5512,18 +5556,39 @@ - (NSArray *) childWindows
...
+      return;
     }
-  [_children addObject: child];
+
   [child setParentWindow: self];
+  RETAIN(child);
+  [_children removeObjectIdenticalTo: child];
+  if (place == NSWindowBelow)
+    {
+      [_children insertObject: child atIndex: 0];
```
fredkiefer, line 5569: This only works if we initialize `_children` in the dedicated initializer, which we currently don't do.

OK, I found the code below , you now create the array in setParentWindow:, this should work, although it is a bit confusing.

Source/NSWindow.m
gcasa: Done.

Source/NSWindow.m
gcasa: I'm adding a clarifying comment to the code...  I'm trying to dumb it down so I don't sound too intelligent so that you don't assume it's AI. (that's a joke)

gcasa: in short ( ;) ): It's so that the frames of unrealized/deferred windows are updated as well.

fredkiefer (approved):

PR #912, opened 2026-08-14
Update NSDiffableDataSource to be more compliant with the macOS implementation

gcasa:
There were some issues with the initial implementation.  These changes attempt to address some of those issues.

AI Disclosure:  AI was used in this change to help create tests, documentation, and detect bugs.

### DIFF
--- a/Headers/AppKit/NSDiffableDataSource.h
+++ b/Headers/AppKit/NSDiffableDataSource.h
@@ -43,6 +43,7 @@
 @class NSMutableSet;
 @class NSIndexPath;
 @class NSView;
+@class NSTableRowView;
 
 @protocol NSCollectionViewDataSource;
 @protocol NSCollectionViewPrefetching;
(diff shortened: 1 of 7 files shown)


gcasa:
@rfm Can you review this one?

rfm (commented):
I think the title of this issue is problematic in itself since in this case 'more compliant' seems to largely be 'less portable', though there do seem to be some functional improvements internally too.

To make this class work properly and portably, we need to avoid dependency on blocks, and Apple baked blocks into the new API.  So a decent fix would implement the blocks support in terms of more portable code.

The overarching design of the GNUstep/OpenStep APIs uses delegates (and occasionally,  function callbacks) to do the jobs that Apple has started using blocks for.  So we really need to make sure that new classes include delegate based methods, and implement blocks support on top of that.

eg.
-initWithCollectionView:itemProvider:
might call an -initWithCollectionView:delegate: method to initialise the instance with itself as its delegate, and then call a method to set the block as the mechanism the delegate uses to provide items.

With the current state of the class it is not usable for GNUstep development generally, only as a tool for porting stuff from OSX.

gcasa:
> I think the title of this issue is problematic in itself since in this case 'more compliant' seems to largely be 'less portable', though there do seem to be some functional improvements internally too.
> 
> To make this class work properly and portably, we need to avoid dependency on blocks, and Apple baked blocks into the new API.  So a decent fix would implement the blocks support in terms of more portable code.
> 
> The overarching design of the GNUstep/OpenStep APIs uses delegates (and occasionally,  function callbacks) to do the jobs that Apple has started using blocks for.  So we really need to make sure that new classes include delegate based methods, and implement blocks support on top of that.
> 
> eg.
> -initWithCollectionView:itemProvider:
> might call an -initWithCollectionView:delegate: method to initialise the instance with itself as its delegate, and then call a method to set the block as the mechanism the delegate uses to provide items.
> 
> With the current state of the class it is not usable for GNUstep development generally, only as a tool for porting stuff from OSX.  

Thank you for the feedback.  I will make sure it is more portable.

gcasa:
@rfm I addressed the issue you mentioned.

Issue #917, opened 2026-08-24
Use after free of NSViewController's _topLevelObjects when -viewDidLoad raises


ashalkhakov:
`-[NSViewController loadView]` does:

```
  [nib instantiateNibWithOwner: self topLevelObjects: &_topLevelObjects];
  ...
  [self viewDidLoad];      // may raise
  ...
  RETAIN(_topLevelObjects);   // <-- only reached if viewDidLoad returns
```

NSNib hands back an AUTORELEASED array. If -viewDidLoad raises (an unresolved selector, a KVC error, anything), the exception unwinds
 loadView before the RETAIN, the autorelease pool drains and frees the array — but the _topLevelObjects ivar still points at it. When the view  controller is later deallocated, its -dealloc runs

```
[_topLevelObjects makeObjectsPerformSelector: @selector(release)];
```

on the freed array: heap corruption / segfault at some later pool drain. I triggered this in my app.

Here is a test case:

1. `ViewControllerUAF.xib`:

```
<?xml version="1.0" encoding="UTF-8"?>
<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="24127" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none">
    <dependencies>
        <deployment identifier="macosx"/>
        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="24127"/>
        <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
    </dependencies>
    <objects>
        <customObject id="-2" userLabel="File's Owner" customClass="CrashVC">
            <connections>
                <outlet property="view" destination="1" id="201"/>
            </connections>
        </customObject>
        <customObject id="-1" userLabel="First Responder" customClass="FirstResponder"/>
        <customObject id="-3" userLabel="Application" customClass="NSObject"/>
        <view id="1">
            <rect key="frame" x="0.0" y="0.0" width="100" height="100"/>
            <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
        </view>
    </objects>
</document>
```

2. `viewcontroller-uaf-test.m`:

```
/*
 * Build (needs ViewControllerUAF.xib next to the binary):
 *   clang viewcontroller-uaf-test.m -o viewcontroller-uaf-test \
 *     -fobjc-runtime=gnustep-2.0 -fexceptions -fblocks \
[...]

PR #918, opened 2026-08-24
Fix use-after-free in NSViewController when viewDidLoad raises an exception

ashalkhakov:
Fix for #917 

AI disclosure: AI was used to find & fix the bug.


fredkiefer (approved):
Looks reasonable. Thank you for the fix

PR #919, opened 2026-08-24
Fix <null> being shown in textfields that use Cocoa bindings.

ashalkhakov:
With this patch, an NSTextField with Cocoa binding will show a <null> if the bound value is nil. With this patch, it will correctly show nothing at all, as on Cocoa.


fredkiefer (approved):
Looks good, we may have to inspect the code to see, whether we should actually handle that case further down, while decoding.

PR #920, opened 2026-08-24
NSTextView: add support for value binding

ashalkhakov:
Cocoa's `NSTextView` exposes a value binding (the plain-text string), but gnustep-gui's `NSTextView` exposes only `editable` — so  the nib loader silently drops the value bindings, leaving the views empty on load and their edits never reaching the model.

AI disclosure: AI was used to make this PR.


fredkiefer (approved):
Thank you

PR #921, opened 2026-08-24
Sound: apply NSSound sink volume as software gain

### LINKED ISSUE #914: NSSound sinks: apply volume as software gain (sample scaling) instead of mixer/device volume
Currently, `NSSound` volume handling in the sound sink plug-ins is inconsistent and, in the default case, non-functional:

- `Tools/sound/AudioOutputSink.m` (libao): `-setVolume:` is a no-op - "Functionality not supported by libao" - so `[NSSound setVolume:]` does nothing at all on systems using this sink.
- `Tools/sound/GSOSSSoundSink.m` (OSS): `-setVolume:` issues a `SNDCTL_DSP_SETPLAYVOL` ioctl, i.e. it changes the device play volume. That is effectively a mixer change that affects everything on the machine, not just the sound being played.

I propose to implement sink volume as software gain: scale the PCM samples themselves before they are handed to the output device, instead of touching any mixer or device-level volume.

The branch [`sound-software-gain-volume`](https://github.com/probonopd/libs-gui/tree/sound-software-gain-volume) contains a working implementation:

- A small new header, `Tools/sound/PCMGain.h`, with endian-safe signed PCM scaling for 8/16/24/32-bit samples (with clamping).
- `AudioOutputSink` stores the requested volume and scales samples in `-playBytes:` before calling `ao_play()`.
- `GSOSSSoundSink` drops the `SNDCTL_DSP_SETPLAYVOL` ioctl and uses the same sample scaling.

Reasons:

- `[NSSound setVolume:]` actually works everywhere, matching Cocoa semantics where an application's sound volume is its own business.
- Volume changes and fades stay local to the app playing the sound; the system mixer is never touched. As a concrete use case: a media player can fade out its playback by lowering amplitude when quitting without affecting the user's system volume.
- No behavior is lost: the libao path previously had no volume support anyway, and per-application software gain is arguably what an app-level volume API should mean.

Before I open a pull request: would a change in this direction be acceptable? Happy to adjust details (e.g. whether the OSS sink should keep device volume as a fallback, or whether the gain helper should live elsewhere).

Reference branch: https://github.com/probonopd/libs-gui/tree/sound-software-gain-volume

I appreciate your thoughts on this.

probonopd:
Closes #914

Description at
* https://github.com/gnustep/libs-gui/issues/914

### DIFF
--- a/Tools/sound/AudioOutputSink.m
+++ b/Tools/sound/AudioOutputSink.m
@@ -32,11 +32,17 @@
 #include <GNUstepGUI/GSSoundSink.h>
 #include <ao/ao.h>
 
+#include "PCMGain.h"
+
 @interface AudioOutputSink : NSObject <GSSoundSink>
 {
   ao_device *_dev;
   int _driver;
   ao_sample_format _format;
+  /* Software gain applied to samples in playBytes; never a mixer */
+  float _volume;
+  void *_gainBuffer;
+  NSUInteger _gainBufferSize;
 }
 @end
 
@@ -58,6 +64,9 @@
 
 - (void)dealloc
 {
+  free(_gainBuffer);
+  _gainBuffer = NULL;
+  _gainBufferSize = 0;
   [super dealloc];
 }
 
@@ -74,6 +83,9 @@
   
   _format.channels = (int)channelCount;
   _format.rate = (int)sampleRate;
+  _volume = 1.0;
+  _gainBuffer = NULL;
+  _gainBufferSize = 0;
   
   switch (encoding)
     {
@@ -131,19 +143,68 @@
 
 - (BOOL)playBytes: (void *)bytes length: (NSUInteger)length
 {
-  int ret = ao_play(_dev, bytes, (uint_32)length);
+  void *out = bytes;
+  int ret;
+
+  /* Fade/volume is done by scaling the samples (amplitude), so it
+     stays independent of any system mixer settings. */
+  if (_volume < 1.0f)
+    {
+      BOOL be;
+      int bits = _format.bits;
+
+      switch (_format.byte_format)
+        {
+          case AO_FMT_BIG:
+            be = YES;
+            break;
+          case AO_FMT_LITTLE:
+            be = NO;
+            break;
+          default:
+            /* AO_FMT_NATIVE depends on the host byte order. */
+            be = (NSHostByteOrder() == NS_BigEndian);
+            break;
+        }
+
+      if (_gainBufferSize < length)
+        {
+          void *nb = realloc(_gainBuffer, length);
+          if (nb != NULL)
+            {
+              _gainBuffer = nb;
+              _gainBufferSize = length;
+            }
+        }
+      /* If we could not allocate a scratch buffer, play unscaled. */
+      if (_gainBuffer != NULL && _gainBufferSize >= length)
+        {
+          PCMGainApply(bytes, length, bits, be, _volume, _gainBuffer);
+          out = _gainBuffer;
+        }
+    }
+
+  ret = ao_play(_dev, out, (uint_32)length);
   return (ret == 0 ? NO : YES);
 }
 
-/* Functionality not supported by libao */
+/* Volume as software gain on the samples, not a device/mixer change */
 - (void)setVolume: (float)volume
 {
-  return;
+  if (volume < 0.0f)
+    {
+      volume = 0.0f;
+    }
+  else if (volume > 1.0f)
+    {
+      volume = 1.0f;
+    }
+  _volume = volume;
 }
 
 - (float)volume
 {
-  return 1.0;
+  return _volume;
 }
 
 - (void)setPlaybackDeviceIdentifier: (NSString *)playbackDeviceIdentifier
--- a/Tools/sound/GSOSSSoundSink.m
+++ b/Tools/sound/GSOSSSoundSink.m
@@ -16,9 +16,12 @@
 #import <GNUstepGUI/GSSoundSource.h>
 #import <GNUstepGUI/GSSoundSink.h>
 
+#include "PCMGain.h"
+
 #include <sys/ioctl.h>
 #include <unistd.h>
 #include <fcntl.h>
+#include <stdlib.h>
 #include <sys/soundcard.h>
 
 
@@ -34,6 +37,10 @@
 	int channels;
 	int format;
 	int rate;
+	/* Software gain applied to samples; never the device play volume */
+	float volume;
+	void *gainBuffer;
+	size_t gainBufferSize;
 }
 @end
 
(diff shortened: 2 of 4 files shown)


Tools/sound/AudioOutputSink.m
```diff
@@ -131,19 +143,46 @@ - (void)close
...
 {
-  int ret = ao_play(_dev, bytes, (uint_32)length);
+  void *out = bytes;
+  int ret;
+
+  /* Fade/volume is done by scaling the samples (amplitude), so it
+     stays independent of any system mixer settings. */
+  if (_volume < 1.0f)
+    {
+      BOOL be = (_format.byte_format == AO_FMT_BIG);
```
gcasa: Do we need to use NSHostByteOrder() somewhere here since we are testing for big/little endian-ness?  On a big-endian host this might treat AO_FMT_NATIVE as little-endian since it would not be equal to AO_FMT_BIG.

Tools/sound/AudioOutputSink.m
```diff
@@ -131,19 +143,46 @@ - (void)close
...
+      int bits = _format.bits;
+
+      if (_gainBufferSize < length)
+        {
+          _gainBuffer = realloc(_gainBuffer, length);
```
gcasa: You should be checking the failure state here.  I believe realloc will return 0 if it doesn't succeed.

Tools/sound/GSOSSSoundSink.m
```diff
@@ -191,44 +202,90 @@ - (void)close
...
+		switch (format)
+		{
+			case AFMT_S8:			bits = 8;  be = NO; break;
+			case AFMT_S16_LE:		bits = 16; be = NO; break;
+			case AFMT_S16_BE:		bits = 16; be = YES; break;
+			case AFMT_S24_LE:		bits = 24; be = NO; break;
+			case AFMT_S24_BE:		bits = 24; be = YES; break;
+			case AFMT_S32_LE:		bits = 32; be = NO; break;
+			case AFMT_S32_BE:		bits = 32; be = YES; break;
+#ifdef AFMT_S16_NE
```
gcasa: Do we need the NE stuff since they are usually aliases?   On some platforms these might have the same value, which in a switch-case statement could cause a compile error.

Tools/sound/GSOSSSoundSink.m
```diff
@@ -191,44 +202,90 @@ - (void)close
...
+			case AFMT_S32_NE:		bits = 32; be = (AFMT_S32_NE == AFMT_S32_BE); break;
+#endif
+			default:
+				bits = 16; be = NO; break;
+		}
+
+		if (gainBufferSize < length)
+		{
+			free(gainBuffer);
+			gainBuffer = malloc(length);
```
gcasa: Needs to be failure safe, like the allocation above.

Tools/sound/PCMGain.h
```diff
@@ -0,0 +1,107 @@
...
+      p[i] = (uint8_t)((u >> shift) & 0xff);
+    }
+}
+
+/*
+ * Scale signed PCM data (bits = 8/16/24/32, byteFormat big-endian flag)
+ * by vol into out. out may equal bytes for in-place scaling; otherwise
+ * it must have room for length bytes.
+ */
+static inline void PCMGainApply(void *bytes, NSUInteger length, int bits,
```
gcasa, line 168: Tests?

gcasa (commented):
Please make sure to add any new classes, etc to the GNUmakefile. :) Also, did you have a copyright assignment on file?  if not then write assign@gnu.org and request one for past and future changes, if so, then cool. :)

Tools/sound/AudioOutputSink.m
probonopd: Good catch. Fixed: `AO_FMT_NATIVE` now resolves through `NSHostByteOrder()` so big-endian hosts treat native samples correctly instead of assuming little-endian.

Tools/sound/AudioOutputSink.m
probonopd: Fixed. The `realloc` return value is now checked; on failure we keep the old buffer and fall back to playing the samples unscaled rather than using a NULL pointer.

Tools/sound/GSOSSSoundSink.m
probonopd: Removed. They alias to LE/BE on most platforms and produced duplicate case labels (compile error). The explicit LE/BE cases plus the default already cover the formats.

Tools/sound/GSOSSSoundSink.m
probonopd: Fixed. Switched to `realloc` with a NULL check and only update the tracked size on success, falling back to unscaled playback if allocation fails.

Tools/sound/PCMGain.h
probonopd: Added Tests/`gui/PCMGain/PCMGain.m` covering 8/16/24/32-bit, little- and big-endian, in-place scaling, `vol>=1.0` passthrough, and `vol==0` zeroing.

gcasa:
Hey @fredkiefer Let me know if there is anything I missed.  Please.

probonopd:
> Also, did you have a copyright assignment on file? if not then write assign@gnu.org and request one for past and future changes

Not yet... do you think I need one fur such a small change already?

gcasa:
> > Also, did you have a copyright assignment on file? if not then write [assign@gnu.org](mailto:assign@gnu.org) and request one for past and future changes
> 
> Not yet... do you think I need one fur such a small change already?

Well, given that it is not that much it shouldn't be needed yet, but it might be good just to do it since you're likely to make more changes in the future.  My policy has always been to allow people to contribute large changes as long as they are in the process.  The FSF lately has been taking a LONGGGGG time lately. (talk to @IOOI-SqAR Lars about this... they took about 10 years to get to him... yes... 10 years)

gcasa (approved):
I approve of these changes.  Let's give @fredkiefer some time to chime in, but as he assigned this to me I think it's good.  The only additional comment I have is with respect to style.  Some of your indentiation is not correct.  It should conform to the GNU style.  Two spaces indented under a conditional or loop instead of flush with it.

fredkiefer:
Looks good to me too. I would suggest to unroll the two read and write helper functions by having one for each byte count (1, 2, 3 and 4). That would make the code a lot faster and easier to understand. But this is just a suggestion.

probonopd:
Done. The read/write helpers are now split into one function per sample width (8/16/24/32), so `PCMGainApply` dispatches directly without any per-sample loops or byte-swap math. Thanks for the suggestion.

gcasa:
Given that @probonopd made the requested changes, @fredkiefer and I both approved, I am merging this.

PR #922, opened 2026-08-25
Fix selectionIndexes init of nib-decoded NSArrayControllers

ashalkhakov:
Without the fix, _selectionIndexes would remain nil and eventually freeze the app. With the fix, selection works.

AI disclosure: this fix made with AI support.


fredkiefer (approved):
Look good.
Could you please try to tell your AI to leave out these useless comments? They should end up in the commit message, but not in the real code.

PR #923, opened 2026-08-25
Fix NSTableView bound selection

ashalkhakov:
If NSTableView has a selectionIndexes binding, update it, otherwise update the value binding.

The order was reversed, so if you had value bindings in the columns, the selection index would never be changed.

AI disclosure: this PR is made with AI support.

### DIFF
--- a/Source/NSTableView.m
+++ b/Source/NSTableView.m
@@ -6804,23 +6804,41 @@ For a more detailed explanation, -setSortDescriptors:. */
 
 - (void) _postSelectionDidChangeNotification
 {
-  NSTableColumn *tb = [_tableColumns objectAtIndex: 0];
   GSKeyValueBinding *theBinding;
-  
-  theBinding = [GSKeyValueBinding getBinding: NSValueBinding
-				   forObject: tb];
 
-  // If there is a binding, send the indexes back
+  /* The table view's own selectionIndexes binding (established by
+     [tableView bind: NSSelectionIndexesBinding toObject: controller ...],
+     e.g. from a nib) is registered with the table view as the source
+     object - push the new selection back through it so the bound
+     controller's selection follows user clicks. */
+  theBinding = [GSKeyValueBinding getBinding: NSSelectionIndexesBinding
+				   forObject: self];
   if (theBinding != nil)
     {
-      id observedObject = [theBinding observedObject];
+      [theBinding reverseSetValue: _selectedRows];
+    }
+  else if ([_tableColumns count] > 0)
+    {
+      /* Fallback: a selectionIndexes binding established with the observed
+	 controller itself as source, found via the first column's value
+	 binding. */
+      NSTableColumn *tb = [_tableColumns objectAtIndex: 0];
 
-      // Set the selection indexes on the controller...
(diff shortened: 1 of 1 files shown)


Source/NSTableView.m
```diff
@@ -6804,23 +6804,41 @@ - (void) _postSelectionIsChangingNotification
...
+				       forObject: tb];
+
+      // If there is a binding, send the indexes back
       if (theBinding != nil)
 	{
-	  [theBinding reverseSetValue: _selectedRows];
+	  id observedObject = [theBinding observedObject];
+
+	  // Set the selection indexes on the controller...
+	  theBinding = [GSKeyValueBinding getBinding: NSSelectionIndexesBinding
```
fredkiefer, line 6836: This looks wrong. We are already in the else branch, so this binding should not exist. Does this mean the whole change is wrong?

ashalkhakov: Thanks for pointing this out. I'll check it with and without this fix again and let you know.

ashalkhakov: @fredkiefer Could you take a look at [tableview-selection-push-test.m](https://github.com/ashalkhakov/UDQuakeTools/tree/gnustep-fixes/Scripts/gnustep-patch-repros)? I just tried this manually on a fresh GNUstep installation and I can see that without the patch, the selectionIndexes indeed does NOT change when row 1 is selected, even though selectedRow changes to 1. With the patch, both selectionIndexes and selectedRow correctly update.
Screenshot: image

ashalkhakov: image
A better screenshot is this. Sorry, I can't figure out how to paste text from my VM.

fredkiefer: I did not say that this patch does not address a real issue. What I am saying is that the code I pointed to should never be executed. I suggest that you first discuss that with the AI you used to generate that code. Maybe it has an explanation or it will fix the code. I don't want to spend too much time on AI generated fixes. Especially not when they are of questionable quality.

rfm:
I don't know what's going on with bindings, but I don't see why the indicated bit of code should never be executed.
It appears to be conditional on theBinding being non-nil,  but theBinding is set to [GSKeyValueBinding getBinding: NSValueBinding forObject: tb] immediately before the test, and I don't know why that would necessarily be nil.

Source/NSTableView.m
fredkiefer: @rfm actually it is not. It is set in line 6836 to a different value `[GSKeyValueBinding getBinding: NSSelectionIndexesBinding forObject: observedObject]` and I mistook this to be the same as on line 6814 `[GSKeyValueBinding getBinding: NSSelectionIndexesBinding forObject: self]`. It uses the same key, but on a different object. So the code is not as wrong as I originally thought.
This is just the old code that was in place before. So the change is just that the other binding is taken care of first and if that is not present the old code gets used.

fredkiefer (commented):
Sorry for the confusion. Looks fine now on a more thorough inspection.

ashalkhakov:
Thank you!

PR #928, opened 2026-08-28
Add methods to handle sending the image and badge over a DO connection via GSDockManager protocol

gcasa:
This addresses issue#927.  This change adds methods to send the image data and badge information over the GSIconManager protocol.  These methods are optional so that this doesn't break the existing contract between GNUstep and older users of this protocol.  I am using NSData to keep the utilization overhead low.

I am using the app in https://github.com/gcasa/apps-dock to test this change.  It is currently working.  The main driver for this is to make integration between GNUstep and external dock applications easier.  This does not change how the external app works, only how the dock manages the app.  I've used my work on the apps-dock DockWM to shape the changes I've made to libs-gui.  I have made them as minimal as possible.

Disclosure: I am using AI to help debug and test this.  The apps-dock application uses AI pretty extensively; this change uses it in a limited way to suggest fixes when needed and to optimize the libs-gui changes.

### DIFF
--- a/Headers/AppKit/NSDockTile.h
+++ b/Headers/AppKit/NSDockTile.h
@@ -59,8 +59,7 @@
 extern "C" {
 #endif
 
-@class NSView, NSString;
-@class NSImageRep;
+@class NSView, NSString, NSImage;
 
 APPKIT_EXPORT_CLASS
 @interface NSDockTile : NSObject
@@ -71,7 +70,6 @@ APPKIT_EXPORT_CLASS
   BOOL      _showsApplicationBadge;
   NSString *_badgeLabel;
   NSImage  *_appIconImage;
-  NSImage  *_dockTileImage;
 }
 
 /**
@@ -143,4 +141,3 @@ APPKIT_EXPORT_CLASS
 #endif	/* GS_API_MACOSX */
 
 #endif	/* _NSDockTile_h_GNUSTEP_GUI_INCLUDE */
-
--- a/Source/GSIconManager.h
+++ b/Source/GSIconManager.h
@@ -39,3 +39,6 @@ GSGetIconFrame(NSWindow *window);
 
 void
 GSUpdateIconManager(NSImage *image, NSString *badgeLabel);
+
+void
+GSRequestUserAttention(NSUInteger requestType);
--- a/Source/GSIconManager.h
+++ b/Source/GSIconManager.h
@@ -24,6 +24,8 @@
 
 #import <Foundation/NSGeometry.h>
 
+@class NSImage;
+@class NSString;
 #import "AppKit/NSWindow.h"
 
 NSSize
@@ -34,3 +36,6 @@ GSRemoveIcon(NSWindow *window);
 
 NSRect
 GSGetIconFrame(NSWindow *window); 
+
+void
+GSUpdateIconManager(NSImage *image, NSString *badgeLabel);
--- a/Source/GSIconManager.h
+++ b/Source/GSIconManager.h
@@ -42,3 +42,6 @@ GSUpdateIconManager(NSImage *image, NSString *badgeLabel);
 
 void
 GSRequestUserAttention(NSUInteger requestType);
+
+void
+GSCancelUserAttentionRequest(NSInteger request);
--- a/Source/GSIconManager.m
+++ b/Source/GSIconManager.m
@@ -45,6 +45,8 @@
  - (void) setApplicationIconData: (NSData *)data
                        badgeText: (NSString *)badgeText
                     appProcessId: (int)aProcessId;
+ - (void) requestUserAttention: (NSInteger)requestType
+		   appProcessId: (int)aProcessId;
 @end
 
 static BOOL verify = NO;
@@ -296,6 +298,31 @@ GSSendApplicationIconData(NSData *data, NSString *badgeText)
   NS_ENDHANDLER
 
   return sent;
+}
+
+void
+GSRequestUserAttention(NSUInteger requestType)
+{
+  checkVerify();
+
+  if (gsim == nil)
+    {
+      return;
+    }
+
+  NS_DURING
+    {
+      if ([gsim respondsToSelector: @selector(requestUserAttention:appProcessId:)])
+	{
+	  [gsim requestUserAttention: requestType
+			appProcessId: appId];
+	}
+    }
+  NS_HANDLER
+    {
+      GSLostIconManager();
+    }
+  NS_ENDHANDLER
 }
 
 NSRect
--- a/Source/GSIconManager.m
+++ b/Source/GSIconManager.m
@@ -54,9 +54,12 @@ static int appId = 0;
 static NSMutableSet *registeredIcons = nil;
 static unsigned int iconManagerUpdateCount = 0;
 static unsigned int lastIconManagerAttemptUpdate = 0;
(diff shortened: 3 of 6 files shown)


gcasa:
I found an additional issue with AClock: it does not reflect it's changes because it grabs the view and updates it.  So there is another way this needs to be updated.  I also found an error where the proxy isn't released properly.

fredkiefer:
Is this still work in progress or ready for review? If the former is the case please convert it to draft.

gcasa:
> Is this still work in progress or ready for review? If the former is the case please convert it to draft.

My apologies, still a WIP.

gcasa:
@fredkiefer This PR is ready for review.

gcasa:
@fredkiefer  I have converted this back to a draft.  As I am testing both with the new dock app and legacy apps I am discovering issues as I go.  Give me a couple of days to finish this up and I will convert it back.  I just didn't want to have you start reviewing and then change things out from under you.  When I do convert it back it will be ready.

gcasa:
@fredkiefer This should be safe to review now in its current form.

rmottola:
I ponder the idea of using distributed notifications. A long while ago I discussed an identical problem with GWorkspace's dock with Fred and Eric and that was the proposal - not implemented though.
Less flexible, but much easier for the receiving application to set up. We had performance doubts. Certain applications change their icon very often, like timemon. Should it be more "push" at each change or "pull" from the dock? Or push a notification that the icon changed, but let the receiver actually decide when to "pull".

I guess DO should be more efficient compared to a DN for sending NSData the size of an icon repeatedly, @rfm ?

rfm:
I haven't looked at the protocol  so I can't comment on whether a notification model would make more sense.

Technically though, distributed notifications run on top of DO: when a notification is posted it goes to gdnc and from there it is sent to all observers.

So it is intrinsically slower as an extra trip to gdnc is required, but processing/complexity is handed off to the gdnc process (which is doing nothing else), so that can be good for performance on any multi-core/mult-processor system.

In any case, icons are small amounts of data on current hardware.  For large amounts of data we can theoretically pass references to shared memory rather than copying the contents of the memory, but that's not needed here.

gcasa:
> I haven't looked at the protocol so I can't comment on whether a notification model would make more sense.
>
> Technically though, distributed notifications run on top of DO: when a notification is posted it goes to gdnc and from there it is sent to all observers.

Yes, this is how I thought it worked.
 
> So it is intrinsically slower as an extra trip to gdnc is required, but processing/complexity is handed off to the gdnc process (which is doing nothing else), so that can be good for performance on any multi-core/mult-processor system.

Not only is it slower, with notifications you can't have return values and you have to mashal things differently.  The proxy does this for us.  In my opinion there is no reason to switch to an NSDistrubutedNotificationCenter.

> In any case, icons are small amounts of data on current hardware. For large amounts of data we can theoretically pass references to shared memory rather than copying the contents of the memory, but that's not needed here.

Very true.

pcardona34:
While switching to the relevant branch, I needed too to rebuild several apps and I noted, by example, Grr stopped to load its main  window. Below a report of the error:

> patrick@pi500:~/SOURCES/agnostep-desktop/install/RESOURCES/FORKS/Grr $ 
> openapp Grr
> 2026-09-12 12:01:44.469 Grr[93207:93207] NSDocumentClass NSDocumentClass not 
> found
> 2026-09-12 12:01:44.469 Grr[93207:93207] NSDocumentClass NSDocumentClass not 
> found
> 2026-09-12 12:01:44.487 Grr[93207:93207] Loading bundle at 
> /Local/Applications/Grr.app/Resources/TreeDatabase.grrdb
> 2026-09-12 12:01:44.488 Grr[93207:93207] Tree Database Component starting 
> up...
> 2026-09-12 12:01:44.524 Grr[93207:93207] Loading bundle at 
> /Local/Applications/Grr.app/Resources/ArticleTable.grrc
> 2026-09-12 12:01:44.536 Grr[93207:93207] Loading bundle at 
> /Local/Applications/Grr.app/Resources/ArticleView.grrc
> 2026-09-12 12:01:44.540 Grr[93207:93207] Exception occurred while loading 
> model: expected unsigned int and got long long
> 2026-09-12 12:01:44.540 Grr[93207:93207] Failed to load Gorm
> 2026-09-12 12:01:44.540 Grr[93207:93207] ViewProvidingComponent: Failed to 
> load nib ArticleTextViewPlugin.
> 2026-09-12 12:01:44.540 Grr[93207:93207] Loading bundle at 
> /Local/Applications/Grr.app/Resources/DatabaseTreeView.grrc
> 2026-09-12 12:01:44.545 Grr[93207:93207] Loading bundle at 
> /Local/Applications/Grr.app/Resources/ArticleOperations.grrc
> 2026-09-12 12:01:44.546 Grr[93207:93207] Loading bundle at 
> /Local/Applications/Grr.app/Resources/DatabaseOperations.grrc
> 2026-09-12 12:01:44.549 Grr[93207:93207] Loading bundle at 
> /Local/Applications/Grr.app/Resources/Searching.grrc
> 2026-09-12 12:01:44.553 Grr[93207:93207] AppController.m:215  Assertion 
> failed in AppController(instance), method applicationDidFinishLaunching:. 
> Article View plugin could not be loaded.
> 2026-09-12 12:01:44.553 Grr[93207:93207] Problem posting <GSNotification: 
> 0x555594814040> Name: NSApplicationDidFinishLaunchingNotification Object: 
> <NSApplication: 0x55559426c980> Info: (null): <NSException:
> 0x5555951ddeb0> NAME:NSInternalInconsistencyException 
> REASON:AppController.m:215  Assertion failed in AppController(instance), 
> method applicationDidFinishLaunching:.  Article View plugin could not be 
> loaded.
> INFO:(null)

rfm (changes requested):
Mostly this looks pretty good/clean and at least the operation of the DO connection is clear to me.
What's less clear is the exact nature of the protocol between the app and the icon manager.  To be fair the original code here is completely undocumented so the problem is not new, but I feel it could do with a good overview of what messages are expected to be passed between the two ends to do what and when.
I note that the mechanism is enabled when a user default is set to yes or if the default is not set at all.  I would suggest starting by changing the logic so you have to set the user default to opt in to the system ... then this can be merged and tested safely by anyone interested, and we could change it to be turned on by default later.

gcasa:
> Mostly this looks pretty good/clean and at least the operation of the DO connection is clear to me. What's less clear is the exact nature of the protocol between the app and the icon manager. To be fair the original code here is completely undocumented so the problem is not new, but I feel it could do with a good overview of what messages are expected to be passed between the two ends to do what and when. 

Super short explanation: AppKit uses these functions to manage icons, Dock tiles, badges, and attention requests. They send requests to the **GSIconManager** service when available; otherwise, GNUstep handles them locally.  It's to make the transport of the images/badges easier so that they are not dependent on X11 internals.

I am going to write documentation for the protocol before merge.

> I note that the mechanism is enabled when a user default is set to yes or if the default is not set at all. I would suggest starting by changing the logic so you have to set the user default to opt in to the system ... then this can be merged and tested safely by anyone interested, and we could change it to be turned on by default later.

I turned it on by default in my branch only for now.  I think it is okay since the Dock can call the app with arguments to turn it on.   I went ahead and set it so that it now defaults to NO when the default is not set.

rfm:
I think this probably meets higher standards than I'd insist on for base now.

PR #929, opened 2026-08-30
Fix NSMovieView / GSMovieView skipping

gcasa:
This change fixes some annoying issues with NSMovieView/GSMovieView/GSAudioPlayer that were causing the user not to be able to seek to a given place in the video.  Also tried my best here to resolve the skipping issue at the beginning of the video playback.


gcasa:
@fredkiefer This PR is ready for review.

PR #930, opened 2026-08-31
libs-gui: only create icon/miniwindow when GSSuppressAppIcon is explicitly NO

probonopd:
As discussed in the last monthly meeting, creating miniwindows should only happen when requested (opt-in).

Previously, the app icon and miniwindow were created unconditionally and only the display was suppressed. This led to unnecessary allocations and potential issues when `GSSuppressAppIcon` was not set.

With this change, `_appIconInit` returns early if `GSSuppressAppIcon` is not set to `NO`, and miniwindow creation is skipped unless `GSSuppressAppIcon` is `NO`. Also adds a nil check before `GSRemoveIcon` to prevent potential crashes.

This change is based on https://github.com/gershwin-desktop/gershwin-developer/commit/e8f2110b5add3aee153d9f468536a15a38159774

cc @gcasa @pkgdemon

### DIFF
--- a/Source/NSApplication.m
+++ b/Source/NSApplication.m
@@ -3667,7 +3667,10 @@ struct _DelegateWrapper
       
       _app_is_running = NO;
 
-      GSRemoveIcon(_app_icon_window);
+      if (_app_icon_window != nil)
+        {
+          GSRemoveIcon(_app_icon_window);
+        }
       [[self windows] makeObjectsPerformSelector: @selector(close)];
       [NSCursor setHiddenUntilMouseMoves: NO];
       
@@ -4004,20 +4007,16 @@ struct _DelegateWrapper
 
 - (void) _appIconInit
 {
+  if ([[[NSUserDefaults standardUserDefaults]
+    objectForKey: @"GSSuppressAppIcon"] isEqual: [NSNumber numberWithBool:NO]] == NO)
+    {
+      return;
+    }
+
   NSAppIconView	*iv;
   NSUInteger	mask = NSIconWindowMask;
-  BOOL  	suppress;
-  
-  suppress = [[NSUserDefaults standardUserDefaults]
-    boolForKey: @"GSSuppressAppIcon"];
-#if	MINI_ICON
-  if (suppress)
-    {
-      mask = NSMiniaturizableWindowMask;
-    }
-#endif
-  
-  _app_icon_window = [[NSIconWindow alloc] initWithContentRect: NSZeroRect 
+
+  _app_icon_window = [[NSIconWindow alloc] initWithContentRect: NSZeroRect
 				styleMask: mask
 				  backing: NSBackingStoreRetained
 				    defer: NO
@@ -4043,12 +4042,9 @@ struct _DelegateWrapper
     RELEASE(iv);
   }
 
-  if (NO == suppress)
-    {
-      /* The icon window is not suppressed ... display it.
-       */
-      [_app_icon_window orderFrontRegardless];
-    }
+  /* The icon window is not suppressed ... display it.
+   */
+  [_app_icon_window orderFrontRegardless];
 }
 
 - (NSDictionary*) _notificationUserInfo
--- a/Source/NSWindow.m
+++ b/Source/NSWindow.m
@@ -3423,21 +3423,25 @@ checkCursorRectanglesExited(NSView *theView,  NSEvent *theEvent, NSPoint lastPoi
    */
   if (_counterpart == 0 && [srv appOwnsMiniwindow])
     {
-      NSWindow *mini;
-      NSMiniWindowView *v;
-      NSRect rect = NSMakeRect(0, 0, iconSize.height, iconSize.width);
+      if ([[[NSUserDefaults standardUserDefaults]
+        objectForKey: @"GSSuppressAppIcon"] isEqual: [NSNumber numberWithBool:NO]])
+        {
+          NSWindow *mini;
+          NSMiniWindowView *v;
+          NSRect rect = NSMakeRect(0, 0, iconSize.height, iconSize.width);
 
-      mini = [[NSMiniWindow alloc] initWithContentRect: rect
-                                             styleMask: NSMiniWindowMask
-                                               backing: NSBackingStoreBuffered
-                                                 defer: NO];
-      mini->_counterpart = [self windowNumber];
-      _counterpart = [mini windowNumber];
-      v = [[NSMiniWindowView alloc] initWithFrame: rect];
-      [v setImage: [self miniwindowImage]];
-      [v setTitle: [self miniwindowTitle]];
-      [mini setContentView: v];
-      RELEASE(v);
+          mini = [[NSMiniWindow alloc] initWithContentRect: rect
+                                                 styleMask: NSMiniWindowMask
+                                                   backing: NSBackingStoreBuffered
+                                                     defer: NO];
+          mini->_counterpart = [self windowNumber];
+          _counterpart = [mini windowNumber];
+          v = [[NSMiniWindowView alloc] initWithFrame: rect];
+          [v setImage: [self miniwindowImage]];
+          [v setTitle: [self miniwindowTitle]];
+          [mini setContentView: v];
+          RELEASE(v);
+        }
     }
   [self _lossOfKeyOrMainWindow];
   [srv miniwindow: _windowNum];
--- a/Tests/gui/NSFontManager/convert.m
+++ b/Tests/gui/NSFontManager/convert.m
@@ -17,8 +17,11 @@
 #define EQ(a, b) (fabs((double)(a) - (double)(b)) < 0.001)
 
 int
-main(int argc, char **argv)
+main(int argc, char **argv, char **env)
 {
+  GSInitializeProcess(argc, argv, env);
+  CREATE_AUTORELEASE_POOL(arp);
+
   START_SET("convertFont:toSize:")
 
   NS_DURING
@@ -32,7 +35,15 @@ main(int argc, char **argv)
   NS_DURING
     {
       NSFontManager *fm = [NSFontManager sharedFontManager];
-      NSFont *h24 = [NSFont fontWithName: @"Helvetica" size: 24.0];
+      NSArray *available = [fm availableFonts];
+      NSString *fontName = [available count] > 0
+        ? [available objectAtIndex: 0] : nil;
+      NSFont *h24 = fontName != nil
+        ? [NSFont fontWithName: fontName size: 24.0] : nil;
+
+      if (fontName == nil || h24 == nil)
+        SKIP("No installed font is available to test font conversion")
+
       NSFont *h12 = [fm convertFont: h24 toSize: 12.0];
       NSFont *same = [fm convertFont: h24 toSize: 24.0];
 
@@ -48,5 +59,6 @@ main(int argc, char **argv)
   NS_ENDHANDLER
 
   END_SET("convertFont:toSize:")
+  RELEASE(arp);
   return 0;
 }


probonopd:
I suspect the test failures are not caused by this PR (which doesn't touch fonts), what do you think?

```
gui/NSFont/named.m:
Failed file:  named.m aborted without running all tests!
Failed test:     (2026-08-31 14:26:57.787 UTC)   named.m:37 ... fontWithName:size: returns a font
Failed test:     (2026-08-31 14:26:57.787 UTC)   named.m:39 ... the font keeps the requested name
Failed test:     (2026-08-31 14:26:57.787 UTC)   named.m:40 ... the font has the requested point size
```

gcasa:
> I suspect the test failures are not caused by this PR (which doesn't touch fonts), what do you think?
> 
> ```
> gui/NSFont/named.m:
> Failed file:  named.m aborted without running all tests!
> Failed test:     (2026-08-31 14:26:57.787 UTC)   named.m:37 ... fontWithName:size: returns a font
> Failed test:     (2026-08-31 14:26:57.787 UTC)   named.m:39 ... the font keeps the requested name
> Failed test:     (2026-08-31 14:26:57.787 UTC)   named.m:40 ... the font has the requested point size
> ```

I just re-ran this.  It fails in the same way.  I am not sure what could have changed to cause this issue.

gcasa:
The failure in named.m is because there doesn't always have to be a "Helvetica" font.  In the art backend, this was a safe assumption since it copied in a "Helvetica.nfont" (nfonts are something created by Deek / Jeff Teuneissen) to fulfill this.  I am not sure why we didn't see this failure until now, but it's my guess that that is the cause here.

gcasa:
Please try this version of named.m... I have zipped it here.   It retrieves the names of all fonts and then grabs one by name.   This ensures were are using a font that is actually installed.

[named.m.zip](https://github.com/user-attachments/files/31660737/named.m.zip)

probonopd:
Thanks @gcasa.
Wouldn't you want to get that applied independently from my PR, and then rebase my PR on top of it?

gcasa:
> Thanks @gcasa. Wouldn't you want to get that applied independently from my PR, and then rebase my PR on top of it?

I had thought about that.  I will try it.

fredkiefer:
Not sure whether that change will help. What should happen with the name "Helvetica" is that it is hard coded into NSFont and if no font with that name can be found, we try the default font, whatever the backend uses for that. What could be happening there is that none of the well known fonts is present.
And as you correctly asked, why is this happening now and didn't happen for the last merged PR? One possible option is that a new base image is being used. 

I'll have a look at what the frame buffer version of cairo is using here, but I would expect it to have all the normal fontconfig fonts.

gcasa:
I have pushed this change on master.  Please pull that change.

gcasa:
FYI, I have been frustrated with the current behavior. In my view, GSSuppressAppIcon should act as you suggest.

gcasa:
> Not sure whether that change will help. What should happen with the name "Helvetica" is that it is hard coded into NSFont and if no font with that name can be found, we try the default font, whatever the backend uses for that. What could be happening there is that none of the well known fonts is present. And as you correctly asked, why is this happening now and didn't happen for the last merged PR? One possible option is that a new base image is being used.
> 
> I'll have a look at what the frame buffer version of cairo is using here, but I would expect it to have all the normal fontconfig fonts.

My approach here was to avoid the hardcoding and find all fonts and get the name of the first one.  This way it preserves the intention of the test.  I think you're saying we should hardcode "Helvetica" so that referencing that (if Helvetica itself isn't available) it will get the default font.  Is that correct?

gcasa:
So now the font tests are failing in a different place?

pkgdemon:
> So now the font tests are failing in a different place?

From the test log:
```
Testing convert.m...
Running gui/NSFontManager/convert.m...
Start set:       convert.m:22 ... convertFont:toSize:
2026-09-02 17:42:19.885 convert[28134:28134] Failed to get _NET_WORKAREA
2026-09-02 17:42:19.894 convert[28134:28134] Failed to get _NET_WORKAREA
Failed test:     (2026-09-02 17:42:19.904 UTC)   convert.m:40 ... the converted font has the requested size
Failed test:     (2026-09-02 17:42:19.904 UTC)   convert.m:42 ... the converted font keeps the original name
Passed test:     (2026-09-02 17:42:19.904 UTC)   convert.m:44 ... converting to the current size returns the original font
End set:         convert.m:50 ... convertFont:toSize:
```

probonopd:
Looks like the Cairo backend needs a window for the screen to be initialized, which is used for font size conversions. Since we no longer have the miniicon windows, we probably need to initialize the display connection manually in the test.

Does this sound correct @gcasa?

gcasa:
> Looks like the Cairo backend needs a window for the screen to be initialized, which is used for font size conversions. Since we no longer have the miniicon windows, we probably need to initialize the display connection manually in the test.
> 
> Does this sound correct @gcasa?

I don't think so.  From my experience... (granted I have limited experience with the backend, thankfully it is very stable)... initializeContext for CairoContext currently only registers the default graphics context and font classes.  It doesn't create or inspect a window.   It just creates the display server.  What leads you to the conclusion that a window is needed?

probonopd:
the test failures were caused by the default graphics context lacking proper screen/drawable initialization when the app icon window is no longer created.

Two options are proposed to fix this.

In [`NSApplication._init`](https://github.com/gnustep/libs-gui/blob/4b8a9233c8c81dad8bd716171399a5b729c3a18d/Source/NSApplication.m#L872-L937), the sequence is:

1. `initialize_gnustep_backend()` - registers context classes
2. `[GSDisplayServer serverWithAttributes:nil]` - connects to display  
3. `[[NSScreen mainScreen] deviceDescription]` at [line 899](https://github.com/gnustep/libs-gui/blob/4b8a9233c8c81dad8bd716171399a5b729c3a18d/Source/NSApplication.m#L899) - retrieves screen info
4. `[NSGraphicsContext graphicsContextWithAttributes:attributes]` - creates default context
5. `[self _appIconInit]` at [line 934](https://github.com/gnustep/libs-gui/blob/4b8a9233c8c81dad8bd716171399a5b729c3a18d/Source/NSApplication.m#L934) - previously created the icon window (which initialized screen-related state)

The problem: the default context created in step 4 is not associated with any window. In the xlib backend's [`GSContext.m`](https://github.com/gnustep/libs-back/blob/master/Source/gsc/GSContext.m#L225-L226), the association between context and window device only happens when a window is passed as the destination attribute (lines 225-226):

```objc
[server setWindowdevice: [(NSWindow*)dest windowNumber] 
                forContext: self];
```

With the old code, `_appIconInit` created an icon window, which triggered this `setWindowdevice:forContext:` call and initialized the gstate's `context` and `drawMechanism` fields. With the new code, this initialization no longer occurs.

Note: gcasa mentioned `initializeContext` for CairoContext, but this method doesn't exist in the codebase.

**Two options to fix this:**

**Option A:** Fix in the test's `main()` function in `Tests/gui/NSFontManager/convert.m`, before font operations:

```objc
// Force screen context initialization before font operations
(void)[GSCurrentServer() screenRContext];

// Now safe to do font operations
NSFontManager *fm = [NSFontManager sharedFontManager];
```

**Option B:** Fix in `NSApplication._init` in `Source/NSApplication.m`, after creating the default context:

```objc
_default_context = [NSGraphicsContext graphicsContextWithAttributes: attributes];
RETAIN(_default_context);
[NSGraphicsContext setCurrentContext: _default_context];

// Ensure screen context is initialized before any font operations
(void)[GSCurrentServer() screenRContext];
```

Option A is less risky because it only affects this specific test and doesn't change behavior for other apps. Option B is a more comprehensive fix but changes core initialization behavior that could have wider unintended consequences.

Which option do you prefer, or would you like to pursue both?

fredkiefer:
Sorry both options are wrong. Did you come up with this yourself or did you use AI for that? If the later is the case, please switch to a better version.
If this was your own idea, please have a look at where screenRContext is defined and used. It only exists in back and there in XGServer, which means, we can only use it in the x11 based backends and no others. So if you really want to use this method, do so in the backend.Perhaps call it at the end of _initXContext? But I am really not sure. 

As this PR already caused some unnecessary changes in gui, I am very reluctant here. Could we please postpone all this until after the release?

probonopd:
Agree that if this is seen as a risky change it should be done after the release.

Please note that already without further changes, the tests pass now, by the test creating a window. My theory is that the issue did not exist before this PR because a window was always created  (the miniwindow) that also initialized a context, and now that the miniwindow is not always created anymore, we need a window (or at least a context) to be created by some other means (e.g., the test). While I did use AI to help me research (after all, this codebase is still unfamiliar to me), i came to the conclusions myself, so I am to blame if the logic is wrong. Thanks for the hints, will investigate more.

pkgdemon:
If I am following correctly.  Does the test itself just need another app to open to create a window?  It sounds like just a broken test that could be fixed by simply opening an app as well?

probonopd:
Well, for now I have just changed the test to open a window, but this might be more than what is needed (establish a context).

pkgdemon:
Looking at the test app code itself a little bit and I will run this later and get a better understanding of what this test really does but for something that requires appkit an NSWindow has to at least exist somewhere right?  If it were a tool no window would be required right?  

https://github.com/gnustep/libs-gui/blob/master/Tests/gui/NSFontManager/convert.m#L13-L15

Then again I've seen appkit applications open where no NSWindow is created at the start until something is selected within a menu.  Document viewers come to mind.  I think TextEdit was like that until I forked and made it create a new document to have a window if not document was being opened.  I believe I may have seen image, or PDF viewers do the same where no NSWindow is created unless you are opening something inside it.

Source/NSApplication.m
```diff
@@ -4004,20 +4007,16 @@ - (void) _loadAppIconImage
...
-  NSAppIconView	*iv;
-  NSUInteger	mask = NSIconWindowMask;
-  BOOL  	suppress;
-  
-  suppress = [[NSUserDefaults standardUserDefaults]
-    boolForKey: @"GSSuppressAppIcon"];
-#if	MINI_ICON
-  if (suppress)
+  if ([[[NSUserDefaults standardUserDefaults]
+    objectForKey: @"GSSuppressAppIcon"] isEqual: [NSNumber numberWithBool:NO]] == NO)
```
rmottola, line 4011: I think this is overkill, boolForKey should do it. boolForKey returns NO if the key is not set, so no suppression, which is "default". If set, then suppression is active. By spec, if they key is not boolean or not parsed, it is set to NO, so it is safe.

gcasa: I agree with @rmottola.   I have been fighting to suppress miniwindows with apps-dock.  It's annoying that this default doesn't work properly.

pkgdemon: I also wanted to note I noticed that on stock gnustep without this fix if SupressAppIcon is enabled TextEdit always opens a new document window on bottom of screen, very tiny, that jumps to the top of the screen when moved anywhere because of this extra hidden window.   I can confirm this fix works.

rfm: The proposed code sort-of matches the issue description.  If the requirement is to check for GSSuppressIcon being explicitly off (rather than just not being set), the normal/correct code would be on the lines of

setting = [[NSUserDefaults standardUserDefaults] objectForKey: @"GSSuppressIcon"];
if (setting != nil && [setting boolValue] == NO)

Using the -boolValue method rather than an equality test with a boolean NSNumber means the setting works whether it is a boolean, a number with value zero, or a string "NO", which is much more reliable.

pkgdemon: I think the case where GNUstep is not the desktop option in as perhaps a different default like GSUseAppIcon = YES and only then create a mini window would make more sense?  Appimages, running on KDE, that would be seem a bit awkward by something that could be enabled by a theme preference.  To have to explicitly set NO makes distributing applications on those other desktops make things more hassle to figure out how to ship that setting?

rmottola: Well, then I would need to set it to YES when I needed it. The two settings both make sense in different environments. Personally I don't want to favour KDE over WindowMaker.
What makes sense is to provide a mean to allow distributing and setting a bunch of setting automatically to best "suit" the target.
Right now, the only way know is a script that runs and calls "default write" to the NSGlobal Domain (or if you have only one single app you can pass the value on the command line with a wrapper).

So if you know you target explicitly just one environment it works.

But more generally you don't know, you prepare a package, it could be installed into WindowMaker, KDE, XFCE, whatever! Even stand-alone on a server and then served remotely through X11.

An idea to improve this script would be anbe to pass defaults a full plist with a coherent set of settings at once. Then prepare different defaults plist.  E.g. maybe you want GSUseAppIcon to NO, but then want GSUseIconManager to YES to enable a custom dock. But on a different setup you may want "nothing" of all those icon stuff.

pkgdemon: If every single packaged app has to suppress icons to work in other environments would they still work right in an environment where appicons are desired?  I’m not 100 percent sure of the defaults order in that case, but could GWorkspace itself also opt into appicons?  I believe that it could?  I honestly believe a separate theme is the most ideal approach for this opt in setting but it seems like GWorksapce could be a middle ground?  Again it seems like opt in is the most POLA and more agnostic here unless GNUstep is not for making applications and its sole focus instead is being a desktop environment project.

pkgdemon: For what it's worth @rmottola I don't use KDE, or other desktops typically either, I don't have much interest in that use case.  But one of the things I like so much about GNUstep is how agnostic it already is, and putting that hat on I just think this is an area where the project could go a lot further by making AppIcons opt in.  I think it is something for the project to consider as best practices down the road if not here in this PR.

probonopd: Well I think this comes down to what is "POLA" (as they call it in FreeBSD; the "principle of least astonishment"). Inside a NeXT-like desktop environment, miniicons are expected by the user. Outside of this very special use case they are surprising; people don't expect them and have no use for them.

Since GNUstep as a GUI framework that can be used for all kinds of applications also outside of NeXT-like desktop environments, I propose that the miniicons should only be created if the user proactively signals that he wants them. Hence, the title of this PR, "only create icon/miniwindow when GSSuppressAppIcon is explicitly NO".

My personal view is that this kind of "strangeness" - strange square windows people don't understand, a default theme that draws menus in unexpected ways (vertical rather than horizontal), etc. has so far given the (wrong!) impression that GNUstep as a GUI framework is still stuck in the NeXT 90s and is not suitable for modern applications. As for myself, I was fooled by this back in the day when I tried to make LIVEstep a couple of years ago. Quite possibly I am not the only one. Hence, for the benefit of GNUstep, I would suggest to change the default. That does not take away the feature for those who need it.

So before I make any further code changes in this PR it would be good to know whether the **intent** of this PR is even welcome in the project (from @gcasa I got the impression that it _is_). Maybe this discussion would better be had in the Monthly Official Meetings or on the Mailing List?

pkgdemon: As @probonopd said if we should discuss somewhere else please let us know.  I just keep thinking there has to be a way without the app developer, or packager doing anything to make an app work as expected no matter whether running apps-gworkspace, WindowMaker, Gershwin, KDE, etc.  That is really what I was trying to advocate for here by chiming in.  For me specifically I am not trying to prefer solutions that favor one over the other if that was not clear.  The more I've had time to think about it NSInterfaceStyle not being adaptive is perhaps another hurdle to overcome for that I don't think we can solve in this PR anyway.

@rmottola Could it be considered for this PR to gate what @probonopd was trying to do here behind certain interface styles for the opt in for the short term?  For example NSMacintoshInterfaceStyle, or NSWindows95InterfaceStyle?  Would a revision to gate behavior changes for opt in behind those two make this PR more acceptable?  What I am particularly interested in for now is not creating AppIcons or hidden windows by suppressing for those speciifc layouts especially when a WindowManger is used unless a user or theme opts into use AppIcons for those layouts.  

I've been doing some work on a native WindowManager solution that uses XCBKit since 2024.  I was able to get AppKit decorating everything (even X11 apps) in 2025.  I've more recently gotten it to be able to change to any theme and have everything update on screen instantly.  You can see it here.  

 https://www.youtube.com/watch?v=wB0PVTdTZ-g

I did plan on AppIcon support.  That is one of the reasons I am even more particularly interested in this PR now to take this work I am doing further.  I would like to continue towards shaping it into something that any project can use, not just Gershwin.

rfm (commented):
It seems to me there are two issues here:
1. how we want to handle whether a miniwindow is used or not.
2. the quality of the code used

The first needs some thought (and people are not yet agreed about it).

The second is that the proposed code is not good (it doesn't do the job well and isn't as simple as it could be) ... if we are using a boolean default we want that default to be operated using -boolForKey: and to normally be no/off/false so if we want to have a the icon window *not* used by default, we might scrap the existing default and use a new HSEnableAppIcon default instead, making the situation a whole lot clearer and the code very simple/clear.

However, there is a legitimate question abut what we want the normal behavior to be and whether we can have gui determine that from the way it's set up more generally.

My own feeling is that having no icon window unless someone explicitly wants it is probably the most pragmatic option considering that we don't normally operate inside a gnustep desktop.

I also feel that if we do have everything operating in a native gnustep environment it should be easy for some app like system preferences (or perhaps the workspace app) to configure settings, so the exact form of the setting does not seem to be worth arguing about.

probonopd:
I have rewritten the PR following @rfm's review: `GSSuppressAppIcon` is replaced by a new boolean default `GSEnableAppIcon`, read with `-boolForKey:` and off by default. The icon window and miniwindows are only created when it is YES, so the code that kept a hidden icon window around (and the dead `MINI_ICON` branches) could go. Two small tests cover both settings.

This follows rfm's view that having no icon window unless someone explicitly asks for it is the most pragmatic option, which @gcasa and @pkgdemon also argued for. For @rmottola's WindowMaker case it is a single `defaults write NSGlobalDomain GSEnableAppIcon YES`, which a workspace app or preferences panel could also set, as rfm suggested.

Happy to leave this until after the release, as @fredkiefer asked.

Source/NSApplication.m
rmottola: > @rmottola Could it be considered for this PR to gate what @probonopd was trying to do here behind certain interface styles for the opt in for the short term? For example NSMacintoshInterfaceStyle, or NSWindows95InterfaceStyle? Would a revision to gate behavior changes for opt in behind those two make this PR more acceptable? What I am particularly interested in for now is not creating AppIcons or hidden windows by suppressing for those speciifc layouts especially when a WindowManger is used unless a user or theme opts into use AppIcons for those layouts.


I would not tie it to InferfaceStyle... I would prefer a set of independent settings. Let's interface be interface. You could have several windowmanagers emulating NeXT style, you could have different dock applications. Let's leave Interface to things like menu style and not much more, in my experience it is safer.

rmottola:
> The second is that the proposed code is not good (it doesn't do the job well and isn't as simple as it could be) ... if we are using a boolean default we want that default to be operated using -boolForKey: and to normally be no/off/false so if we want to have a the icon window _not_ used by default, we might scrap the existing default and use a new HSEnableAppIcon default instead, making the situation a whole lot clearer and the code very simple/clear.

Yes the name is clearer. I think options should enable things, not suppress, it always irritated me. However, the default in this case should be YES, that it was it has always been.

> My own feeling is that having no icon window unless someone explicitly wants it is probably the most pragmatic option considering that we don't normally operate inside a gnustep desktop.

Well.. define "desktop". no gui on a server, no problem. WindowMaker / AfterStep or other similar NeXT style? Not a full desktop. KDE? Gershwin? If good detection for certain default can be done it would be good. Else just leave it on, @gcasa 

> I also feel that if we do have everything operating in a native gnustep environment it should be easy for some app like system preferences (or perhaps the workspace app) to configure settings, so the exact form of the setting does not seem to be worth arguing about.

That is true, we should have defaults. Currently SystemPreferences supports the existing Suppress value.
So if we change this, pleas notify me that I update SP. If done with the proper default, it doesn't upset upsers. It could even go with release, but @fredkiefer.

PR #939, opened 2026-09-16
Update to fix badge render issue

gcasa:
Centralize dock tile rendering so that recent changes for image updates are consistent between docktiles and miniwindows.


rmottola:
Unfortunately it doesn't work for me. My test is running GNUMail and GWorkspace dock (icn manager support branch).
My old specific test app does work again though (was broken yesterday) but it displays the badge only in the WindowMaker mini-window, not the dock.
It could be an issue with GWorkspace's Dock.

rmottola:
in DockWM docks show, but because they are re-drawn and look differently.

rmottola:
Now badges display inside GWorkspace's Dock! that is cool.

I think now we need to be sure of the code and coordinate all settings/preferences.

rmottola:
I wonder if we really need the retry-timer, I'm always weary for such retries. Perhaps if it fails, just let the next update fix it?
Just thinking if the next update perhaps is only the next update the badge content changes or if there is an earlier event to piggy-back.

gcasa:
> I wonder if we really need the retry-timer, I'm always weary for such retries. Perhaps if it fails, just let the next update fix it? Just thinking if the next update perhaps is only the next update the badge content changes or if there is an earlier event to piggy-back.

It seemed the only way to solve the problem if the update is missed.  Leaving it to the next change is going to be interpreted as a bug because the user doesn't see the current status.

rmottola:
> > I wonder if we really need the retry-timer, I'm always weary for such retries. Perhaps if it fails, just let the next update fix it? Just thinking if the next update perhaps is only the next update the badge content changes or if there is an earlier event to piggy-back.
> 
> It seemed the only way to solve the problem if the update is missed. Leaving it to the next change is going to be interpreted as a bug because the user doesn't see the current status.

yes understandable, next change is not feasible, but perhaps there is another update event. It was a guess. Maybe @rfm has an idea.

As I read the code I see the Timer is always there, maybe it oculd be created only on failure.

gcasa:
I have added @rfm as a reviewer.  Please take a look and give some advice on this change.

PR #940, opened 2026-09-16
Stop views from redrawing over themselves with a user space scale factor

probonopd:
With GSScaleFactor other than 1.0, views displayed through fractional coordinates kept darkening: anti-aliased edges turned bold and translucent fills went opaque, again with every display pass. The rect a view is displayed in has been converted through the view hierarchy and can miss its invalid rect by floating point error (a rect at y = 2.84e-14 instead of 0), and -displayRectIgnoringOpacity:inContext: only cleared the invalid rect when the displayed rect contained it exactly. The view then stayed marked as needing display for good and every later pass drew it again on its own, over its previous drawing and without its background. The containment test now allows 0.01 units of slack, far below one device pixel at any usable scale factor, so a view that has been fully drawn is marked clean and is drawn once.

cc @pkgdemon


rmottola:
I never set explicitly the scale factor. I remember having a place where text darkens and darkens on redraw.. as soon as I find it again, I will test if this patch helps and report back.

PR #946, opened 2026-09-19
Fix access to deallocated expressions in GSCSTableau

### LINKED ISSUE #945: Illegal access to deallocated expressions in GSCSTableau
In both `-removeRowFromVariable:` and `-pivotWithEntryVariable:exitVariable:` the expression instance is accessed after its owner is deallocated. The access pattern is (in pseudocode):

```
GSCSLinearExpression *expression = [owner expression];
RELEASE(owner);
[expression message]; // Illegal access
```

Since the owner is deallocated, the expression instance is deallocated too, so accessing it is illegal.

https://github.com/gnustep/libs-gui/blob/890ff4ed47c48c34866de88f226d866d2076a7d7/Source/GSCSTableau.m#L222

https://github.com/gnustep/libs-gui/blob/890ff4ed47c48c34866de88f226d866d2076a7d7/Source/GSCSTableau.m#L466

doganov:
If we're gonna continue using the expression instance, take ownership of it before releasing its current owner.  Fixes #945.


rfm (approved):
Yes  a clear memory management fix. Thanks

PR #948, opened 2026-09-19
NSMenuView: carry out no item when the mouse is released away from the menu

probonopd:
Clicking away from an open menu can carry out the item that is still highlighted. An item whose submenu has just been attached keeps the highlight while the pointer is off the menu, so hovering such an item and then clicking somewhere else performs it. Seen with a menu of folders that open when clicked and list their contents when hovered: the click meant to put the menu away opened a folder.

`Tests/gui/NSMenuView/clickOutside.m` presses the mouse on an item that has a submenu, moves the pointer off the menu with `-[GSDisplayServer setMouseLocation:onScreen:]` and releases it there. `cd Tests && gnustep-tests gui/NSMenuView` against master:

```
Passed test:     clickOutside.m:187 ... the item is in the menu and the point clicked away from it is not
Passed test:     clickOutside.m:196 ... releasing the mouse on an item carries it out
Failed test:     clickOutside.m:199 ... releasing the mouse away from the menu carries out nothing
```

and with this change:

```
Passed test:     clickOutside.m:187 ... the item is in the menu and the point clicked away from it is not
Passed test:     clickOutside.m:196 ... releasing the mouse on an item carries it out
Passed test:     clickOutside.m:199 ... releasing the mouse away from the menu carries out nothing
```

The test sets `NSMacintoshInterfaceStyle` on the menu view, since in the other styles `-_executeItemAtIndex:removeSubmenu:` stops at an item whose submenu is open, so the highlight left behind cannot be observed through its action.

`-_trackWithEvent:startingMenuView:` now drops the highlight and carries out no item when the mouse is released over no item of the menu.

cc @pkgdemon


rfm (approved):
Looks good, thanks.  I like that the check is only executed in the case where it is actually necessary (there is currently a selected item), avoiding unnecessary calculation of the menu item under the mouse location.

PR #949, opened 2026-09-20
NSWorkspace: Do not wait for the tool that opens a file or URL

probonopd:
## Problem

`-[NSWorkspace openURL:]` (and `-openFile:`, through the same code) hands the file or URL to the desktop's handler and then waits for that handler to exit:

```objc
NSTask *task = [NSTask launchedTaskWithLaunchPath: launchPath arguments: args];
if (task != nil)
  {
    [task waitUntilExit];
    ...
  }
```

Handlers such as `xdg-open` keep running for as long as the browser they start, so the calling application is blocked the whole time. Clicking a link in a chat window froze that window for minutes, with its main thread here:

```
-[NSTask waitUntilExit]
-[NSWorkspace _openUnknown:]
-[NSWorkspace openURL:]
...
-[NSApplication run]
```

On macOS `openURL:` returns as soon as the URL has been handed over.

## Fix

Start the tool and return YES. The termination status is no longer inspected, since waiting for it is what causes the problem; a handler that fails leaves the same trace it does today.

## How to reproduce

1. A stand-in for a handler that stays alive with the browser, somewhere on `PATH` as `slowhandler`:

```sh
#!/bin/sh
sleep 20
```

2. Tell NSWorkspace to use it:

```sh
defaults write NSGlobalDomain GSUnknownFileTool slowhandler
```

3. Build and run this, with a display available:

```objc
#import <AppKit/AppKit.h>

int main(int argc, const char **argv)
{
  NSAutoreleasePool *pool = [NSAutoreleasePool new];
  NSDate *start;
  BOOL ok;

  [NSApplication sharedApplication];
  start = [NSDate date];
  ok = [[NSWorkspace sharedWorkspace]
    openURL: [NSURL URLWithString: @"https://example.invalid/demo"]];
  printf("openURL: returned %s after %.1f s\n",
    ok ? "YES" : "NO", -[start timeIntervalSinceNow]);
  [pool release];
  return 0;
}
```

```sh
clang $(gnustep-config --objc-flags) openwait.m -o openwait $(gnustep-config --gui-libs)
./openwait
```

Without this patch the call takes as long as the handler runs; with it, it returns at once:

```
openURL: returned YES after 20.1 s      # before
openURL: returned YES after 0.0 s       # after
```

The handler starts in both cases.

Note for anyone reproducing on master: when no service accepts the URL, `openURL:` first puts up a modal panel ("Unable to use Application or Service to open URL") and only continues to the handler once it is dismissed. That panel is a separate issue; this patch does not touch it.

Measured on Linux (X11, cairo backend) with GNUstep gui 0.32, the app's own libs-gui build swapped in through `LD_LIBRARY_PATH`. Upstream master does not build against the libs-base installed here (`NSContentSetBinding`), so the before/after numbers come from the same change applied to that build; `NSWorkspace.m` compiles clean (`-fsyntax-only -Wall`) against this branch's headers.


rfm:
I think, while the observation of behavior is entirely correct, this pull request misses the fact theat current behavior is by design and arguably the best possible behavior.
The mechanism intentionally waits for the 'handler' to complete so that it can report success/failure status.  It's the responsibility of the developer of the handler to do that correctly.   If you want to make use of something like xdg-open then the handler is supposed to launch that in the background and exit, returning a status to the workspace code.
The workspace is entitled to depend on the handler being well behaved.

rfm:
PS. Not saying that GUI actually does anything useful with task exit status yet, but the design is that it can, and in future should.

probonopd:
Thanks - that is a fair point about the design, and I have taken the second change that had crept into this branch back out: the alert panel is now #954, so this PR is only about the waiting again.

On the waiting itself, I think the disagreement is about who the handler is. The code path is `-_openUnknown:`, which is reached when no GNUstep application and no Open URL service accepts the file or URL. What it launches is whatever the desktop nominates, found through `xdg-open`, `open`, `kde-open` and so on. Those are not handlers written for GNUstep, and `xdg-open` in particular does not background the program it starts: it execs the browser and stays until the browser exits. So the workspace is entitled to depend on a well behaved handler, but the handler it gets here is the one the desktop already has, and that one does not behave that way.

The status is the other half of it. `-_openUnknown:` returns `[task terminationStatus] == 0`, so with `xdg-open` the return value of `openURL:` means "the browser has now been closed and exited zero", which is not what a caller reads it as. A caller that wanted to know whether the URL was opened at all gets that answer minutes later, and only if it is prepared to block its user interface until then.

If GNUstep wants the status, I think it has to come from somewhere that can be had promptly. Two ways I can see:

1. Keep the wait, but only for a handler GNUstep knows is well behaved - an entry in the defaults naming a wrapper the desktop provides, say - and start anything else without waiting.
2. Wait with a deadline: return YES once the handler has stayed alive for a moment without failing, and keep the non-zero exit only when it comes quickly. The common failures, no such program and a handler that exits at once with an error, arrive within milliseconds.

I am happy to implement either instead of this patch. What I would like to avoid is the current behaviour on an ordinary Linux or BSD desktop, where clicking a link freezes the application for as long as the browser stays open.

rfm:
I had been thinking about it, and come to the conclusion that your second option  (wait for a short time) is probably the best option to defend against the possibility that the handler hangs (or deliberately stays running).
I agree that seems a perfectly good solution.

PR #954, opened 2026-09-20
NSWorkspace: do not report a failure before trying the last handler

probonopd:
`-[NSWorkspace openURL:]` reports a failure before it has tried the step that succeeds.

When no application and no service accepts the URL, it runs

```objc
NSRunAlertPanel(nil, @"Unable to use Application or Service to open URL",
  @"Continue", nil, nil);

if ([self _openUnknown: [url absoluteString]])
```

`-_openUnknown:` is what opens the URL on a desktop whose handler is not a GNUstep application, so on an ordinary Linux or BSD desktop the user is shown an error about something that is about to work, and the application waits at the modal panel until someone dismisses it.

With no GNUstep handler and no Open URL service registered, a program calling `openURL:` does not come back at all, because nobody dismisses the panel:

```
$ ./uprobe                         # upstream master
calling openURL:
                                   # still waiting when killed after 12 s

$ ./uprobe                         # with this change
calling openURL:
openURL: returned 1
```

```objc
#include <Foundation/Foundation.h>
#include <AppKit/AppKit.h>
int main(void) {
  [NSAutoreleasePool new]; [NSApplication sharedApplication];
  fprintf(stderr, "calling openURL:\n"); fflush(stderr);
  BOOL ok = [[NSWorkspace sharedWorkspace]
    openURL: [NSURL URLWithString: @"https://example.invalid/"]];
  fprintf(stderr, "openURL: returned %d\n", (int)ok); fflush(stderr);
  return 0;
}
```

A URL that cannot be opened at all still returns NO, as before.

This was part of #949 for a while; it is separate from the question of whether `openURL:` should wait for the handler, so it is its own pull request.

cc @pkgdemon


rfm:
I completely agree that this report is premature, but I think that means we should move it to where the final attempt fails rather than removing it completely.

